PR #45, opened 2014-12-31
cli file loader for Insulet

jebeck:
This works for me. Probably doesn't need much review since it's just a dev tool, but I'm going to mark one place (by commenting the diff in the PR) where I had a question for @kentquirk.

@jh-bate you're welcome to take a look as well.

Note that I didn't change much in `insuletDriver.js` yet - in particular there are a lot of `console.log`s (some commented out, some not), some of which will need to be turned into `debug` calls instead, but in the beginning stages of the work I'll be doing with the Insulet data, the `console.log`s will be more convenient for me since I'll be using the CLI file loader. I *will* change them at a later date, just not as part of this PR.


jebeck:
Moved all the drivers into `lib/drivers/` as requested by @kentquirk 

I think I caught all the places where file paths had to change as a result (tests, etc.), but another set of eyes wouldn't go amiss. @jh-bate?

jh-bate:
will do

jebeck:
As discussed with @jh-bate yesterday, merging this as it's low risk now that we've already tagged and deployed uploader to prod.

PR #46, opened 2015-01-08
Support Precision Xtra BGM

kentquirk:
And a bunch of cleanup along the way:
* Move active drivers into the drivers subdirectory (still more to do)
* This also has the early stages of OneTouch UltraMini support, but that doesn't work yet, so the driver is in there but it's not enabled in the UI. (Commented out in appState.js.)
* A new helper function in struct
* A change to npm start so that it works on windows


kentquirk:
@jebeck, you should probably be the one to look this over, but @jh-bate would be fine too. I will still do some final polish and testing before promoting to the store.

lib/drivers/abbottFreeStyle.js
```diff
@@ -0,0 +1,562 @@
...
+  var ETX = 0x03;
+  var CR = 0x0D;
+  var LF = 0x0A;
+
+  var dateTimePattern = '([A-Za-z ]{4}) ([0-3][0-9]) ([0-9]{4}) ' +
+                        '([012][0-9]):([0-5][0-9])(?::([0-6][0-9]))*';
+  function parseDateTime(match, offset) {
+    var fmt = 'MMMM DD YYYY HH:mm:ss';
+    var utc = sundial.parseAndApplyTimezone(match[0], fmt, cfg.timezone).toISOString();
+    var dev = sundial.formatDeviceTime(utc);
```
jebeck: `deviceTime` should be our preserved record of the device's local time; that's not what you're generating here since the timezone has already been applied.

The way I'd do it with the functions we currently provide in sundial is the following (note that `parseAndApplyTimezone` does not require a timezone; if none is provided, it simply parses from the format...perhaps the method name is not the most intuitive here):
```
var parsed = sundial.parseAndApplyTimezone(match[0], fmt);
var dev = sundial.formatDeviceTime(parsed);
var utc = sundial.applyTimezone(parsed, cfg.timezone).toISOString();
```

lib/drivers/abbottFreeStyle.js
```diff
@@ -0,0 +1,562 @@
...
+        }
+        if (found) {
+          return ix;
+        }
+      }
+      return -1;
+    }
+
+    // bstring = stringFromBytes(bytes, 0, bytes.length);
+    // console.log(bstring);
```
jebeck: Do you anticipate needing this (and any other) logging statements in the near future or can they be deleted?

lib/drivers/abbottFreeStyle.js
```diff
@@ -0,0 +1,562 @@
...
+                              units: 'mg/dL'      // everything the Precision XL meter stores is in this unit
+                            });
+    var dataToPost = [];
+    for (var i = 0; i < data.logEntries.length; ++i) {
+      var datum = data.logEntries[i];
+      var offset = sundial.getOffsetFromZone(datum.datetime.dev, cfg.timezone);
+      if (datum.readingType === 'glucose') {
+        var smbg = cfg.builder.makeSMBG()
+          .with_value(datum.reading)
+          // .with_deviceTime(datum.devicetime)
```
jebeck: Why did you comment this out? `deviceTime` should be included.

lib/drivers/abbottFreeStyle.js
```diff
@@ -0,0 +1,562 @@
...
+        // for now we don't post these but we'll want to someday
+      }
+    }
+
+    console.log(dataToPost);
+    return dataToPost;
+  };
+
+  var probe = function (cb) {
+    console.log('attempting probe of oneTouch Mini');
```
jebeck: Copy/paste error referring to oneTouch Mini here?

jebeck:
A few actionable things here around `deviceTime` (see comments above), but otherwise looks good. Since I can't test it locally myself, could you upload a Precision Xtra locally @kentquirk then either use the `window.downloadInputData()` in blip or octopus export and send me the resulting file, so I can give it a final look-see re: data model?

jebeck:
Also confirmed we will get (at least) some minor UI cutoff if we don't do something about that:
screenshot 2015-01-09 13 01 12

kentquirk:
I scrunched things a bit so that I hope you won't get that cutoff. Also fixed up the time stuff but haven't yet tested on windows (needed to check in to do that). Testing now.

kentquirk:
Final changes to the time handling that seem to work correctly, plus merge from latest master.

I think this is now ready to merge to master.

jebeck:
OK, did a last look-through. The squishing appears to work to prevent cut-off in the UI, so that's ready.

Ditto the time changes.

Did you test yet on Windows @kentquirk? If you did (and it was successful), I consider this ready for merge.

Two questions re: data model, both things I've noticed elsewhere, and now is as good a time as any to raise them to wider awareness, but we may decide not to do anything about them now.

1. The `source` field (with a value of simply `device`) is not as detailed as the data model intends (see http://developer.tidepool.io/data-model/v1/, fourth down under 'common fields'). Perhaps it doesn't need to be as detailed as the data model suggests (in which case we should change the data model). The `version` field we're including in the upload metadata object is very close to equivalent, at least on my reading. Thoughts?

2. After a very healthy debate about units and BG readings, we had agreed to store the original values - whether mg/dL or mmol/L - as string values, very similar to how we're storing unconverted `deviceTime`: for auditing purposes only. @jh-bate and I both remember this conclusion, although it's not in the data model documentation and has not been implemented. Raising this issue now for visibility because I don't want it to be completely forgotten, and we're only in the initial stages of adding BG meters to the uploader, so it feels particularly appropriate to re-open the issue now.

jebeck:
Oh, one last thing: since my "fix" for the CareLink parsing error is now a non-starter given the issue with duplicate boluses on subsequent uploads, but it *is* (unfortunately) already on master, we need to be careful about merging this and tagging a new version. Depending on which potential solution we pick, it may be as simple as doing a `git revert` on my commit with the fix.

kentquirk:
Although we have an issue with Precision Xtra on Windows right now, these changes won't be affected by the resolution of that issue, and we need to move forward on other things, so I'm merging them.

PR #47, opened 2015-01-09
fix for sequencing error when BG intervenes between wizard and bolus

jebeck:
@jh-bate can you review?

The issue here is that the simulator expects events in exact sequence. When we bundle together wizard + bolus events, we pass them to the simulator at the same time, but in fact an event may have been sandwiched between them (and when it gets read next - uh oh! we choke on an out-of-sequence error). The fix is to apply the timestamp from the wizard (which is always earlier) to the bolus, so that they both get simulated before any event sandwiched between them. @HowardLook approved this general strategy, although I do want to ask him one more question (about a potential annotation) before this gets merged, but the substance is there and ready to review.


jh-bate:
LGTM - nice work on the extra specific test

jebeck:
Merging, as @HowardLook does not think an annotation is necessary.

PR #48, opened 2015-01-12
pulling in groups work

jh-bate:
@ianjorgensen here is what I have done so far 

- pulled in latest from master so that should already be done for you.
- add common as dep as that is what you where using for the groups and was failing the build. we are already using async so you shouldn't be adding another lib in with similar functionality https://github.com/tidepool-org/chrome-uploader/pull/48/files#diff-23ed28b93ce54157c776a7621ff8accbR98
- fixed the check to see if you should render the uploadsettings https://github.com/tidepool-org/chrome-uploader/pull/48/files#diff-f47eba8b229e5bbb334e51a7342e66f8R33


lib/core/api.js
```diff
@@ -84,6 +85,40 @@ api.user.logout = function(cb) {
...
+  var userId = tidepool.getUserId();
+
+  api.log('GET /access/groups/' + userId);
+
+  tidepool.getViewableUsers(userId, function(err, groups) {
+    if (err) {
+      return cb(err);
+    }
+
+    common.step([
```
jh-bate: @ianjorgensen you introduced common here - you should be using the already used async.

lib/state/appActions.js
```diff
@@ -46,6 +46,13 @@ appActions.bindApp = function(app) {
   return this;
 };
 
+appActions.changeGroup = function(cb) {
+  this.app.setState({
+    uploads: null, // clear upload state
+    targetId: groups[0].userid // todo: put something ref .value here.
```
jh-bate: @ianjorgensen  no reference is being passed to groups - this fails. check the build and also if you run gulp yo would see this too

lib/state/appActions.js
```diff
@@ -85,15 +92,19 @@ appActions.load = function(cb) {
...
-      var user = _.assign({}, account, {profile: profile});
+      var uploadGroups = results[2];
+      console.log(uploadGroups);
+      var user = _.assign({}, account, {profile: profile, uploadGroups: uploadGroups});
 
+      // todo add state for when the user cannot upload to any account
       self.app.setState({
         user: user,
-        targetId: user.userid,
+        targetId: uploadGroups[0].userid,
```
jh-bate: seems odd? so the first user in uploadGroups is always there and its the logged in user?

package.json
```diff
@@ -19,11 +19,12 @@
     "async": "0.9.0",
     "babyparse": "0.2.1",
     "commander": "2.5.0",
+    "common": "0.2.5",
     "lodash": "2.4.1",
     "moment-timezone": "0.2.4",
     "react": "0.12.0",
     "sundial": "1.0.0",
-    "tidepool-platform-client": "0.14.0"
+    "tidepool-platform-client": "git://github.com/tidepool-org/platform-client.git#ian/uploadGroup"
```
jh-bate: @ianjorgensen we obvioulsy need to change this but added while i was trying to test

package.json
```diff
@@ -19,11 +19,12 @@
     "async": "0.9.0",
     "babyparse": "0.2.1",
     "commander": "2.5.0",
+    "common": "0.2.5",
```
jh-bate: @ianjorgensen added common in order to test but we should really just be using the already imported async

ianjorgensen:
@jh-bate i updated the failing things and added some extra features i thought were missing, you can see them in the individual commits

ianjorgensen:
@jh-bate im looking into the failing test.

jh-bate:
@kentquirk this runs for me locally BUT i can only test with carelink data. You will see in all the drivers the groupId is now added for example https://github.com/tidepool-org/chrome-uploader/pull/48/files#diff-94d111f4d39df17bd170c0747ebd1a7aR807.

As I made some updates to the original PR do you want to have a quick look then we can merge and @ursooperduper can update the drop-down as part of her other work.

Ta

jh-bate:
bump @kentquirk  - see above

kentquirk:
@jh-bate -- The order of the arguments was reversed in the dexcom driver; I've fixed that and pushed it to this branch.

But when I run locally (all the servers are on master), the dexcom upload times out when it posts to the data/userid function (that function never returns). When I run against devel, every post to data/userid returns a 404. I haven't figured out what the difference is yet, but something's wrong. In both cases, I was testing a solo account (no upload rights except to itself).

I may get time to look again later. If so, I'll post again.

jh-bate:
thanks - i will have a look.

kentquirk:
With the pull request for jellyfish this is now properly functional. Merging.

PR #49, opened 2015-01-13
CareLink parsing fix: bolus & wizard sequencing

jebeck:
I believe this is ready to go. It doesn't technically have any dependencies, but in order for all users affected by this issue to be able to upload https://github.com/tidepool-org/jellyfish/pull/44 will have to be deployed as well. (That issue was just exposed when this issue was fixed; they are unrelated.)

I created a new annotation code as part of the latest commit, so there will be a tideline change to add some new text for that, but the default annotation will show up regardless, so that's not a dependency either.


lib/carelink/bolusNWizard.js
```diff
@@ -470,129 +472,164 @@ module.exports = function(timezone, model) {
...
-              if (square != null) {
-                datumsBySeqNum[square.uploadId][square.uploadSeqNum] = null;
+      if (findWizard == null) {
+        findWizard = makeFindWizardFn();
+      }
+      
+      if (parsed != null) {
+        switch(parsed.type) {
+          case 'bolusNormal':
+            if (isInCache(parsed)) {
```
kentquirk: In every one of these cases (and yes, I realize this is old code and it's just a layer deeper) you ask if it's in the cache...and then do nothing if it isn't in the cache. Shouldn't that be an error condition?

Similarly with the test for parsed -- what happens if parsed IS null? If that can't reasonably occur, why the test? If it can occur, is it correct to just ignore it?

lib/carelink/bolusNWizard.js
```diff
@@ -470,129 +472,164 @@ module.exports = function(timezone, model) {
   }
 
   var datumsBySeqNum = null, findNormal = null, findSquare = null, findWizard = null;
+  var pointer = 0, cachedObjects = null;
```
kentquirk: Your "buildCache" and cache manipulation routines don't touch "cachedObjects" -- but instead use "datumsBySeqNum". The "cachedObjects" is more of your result list. That's pretty confusing. Can you adjust the names of things so that they agree with each other, please?

kentquirk:
I tried to understand what you have here without actually deciding if you got all the logic correct. You have a bunch of test cases that seem to cover what you're after, and I just commented on the things I found confusing reading the code.

While I'd like to see it de-confused a bit, I'm happy otherwise -- so fix if you think it's worth it and merge at your convenience.

lib/carelink/bolusNWizard.js
jebeck: In response to first question: no. The problem that the cache is solving is that we sometimes need to process events *before* we get to them in order of timestamp/rows in the CSV. For example, we often come across a wizard, look for an associated bolus...if we find such a bolus, we process both and remove *both* wizard and bolus from the cache. Then the next row in the CSV to be processed might very well be the bolus we just removed from the cache. If it's no longer in the cache, we just skip to the next thing because we know we already processed it.

The test for `parsed` is a little different. I'd call this one a cheddarism - i.e., the pattern of initializing to null and then checking later to see if something is still null. If you look in the CareLink driver [here](https://github.com/tidepool-org/chrome-uploader/blob/jebeck/bolus-and-wiz-seq/lib/drivers/carelinkDriver.js#L51) and [here](https://github.com/tidepool-org/chrome-uploader/blob/jebeck/bolus-and-wiz-seq/lib/drivers/carelinkDriver.js#L233), you'll see that each row of the CSV gets passed to *every* processor. And the pattern for all the processors is to start with `parsed = null` then try to parse the CSV row with the local row-parser (built using `common.makeParser`), and each local row-parser only attempts to parse rows of certain types (whatever's in the domain of that processor) and returns `null` otherwise.

Does that help?

lib/carelink/bolusNWizard.js
jebeck: Oh yeah, these are now awful names. Working on it now.

jebeck:
@kentquirk I renamed some things and expanded a bit of the comments - LMK if it's a bit less confusing now. (I kept the `cachedObjects` name because it's a direct parallel with the settings processor.)

I'm *not* merging, but rather tagging on this branch and building for devel. Unless the decision around deployment has changed, I will wait until after this is all the way through to prod from the branch to update and merge into master.

lib/carelink/bolusNWizard.js
kentquirk: Great, thanks. No need to change anything now, but in general, I like stuff like that to be in the comments. Code tells you what but not why -- so it's sometimes valuable to add some "why".

kentquirk:
This is good, thanks.

PR #50, opened 2015-01-15
Right-click context menu to select server

kentquirk:
Dependent on a new feature in tidepool-platform-client, which is deployed to npm already.

Also contains the feature to have blip open a web page if chrome-uploader is NOT installed, and the uploader if it IS installed (when blip is running in Chrome only). 

After this change, a right click on the uploader will offer a "Server" menu, which has individual markings for local, development, staging, and production, with a check mark next to the current one.

The build process has changed to no longer require pre-setting the environment -- in fact, that now gets a warning message (it should be left at the default, which is production). However, developers can still pre-set it if you want to save clicks while testing (however, the initial state of the check mark in the right click menu will be wrong if you do). 

@jebeck or @jh-bate , please review


kentquirk:
@jebeck -- I've fixed things so that it loads properly for me. The problem doesn't seem to have been a race condition, but just a wrong object path.

main.js
```diff
@@ -31,21 +31,79 @@
...
+      BLIP_URL: 'https://blip-staging.tidepool.io'
+    },
+    Production: {
+      API_URL: 'https://api.tidepool.io',
+      UPLOAD_URL: 'https://uploads.tidepool.io',
+      BLIP_URL: 'https://blip-ucsf-pilot.tidepool.io'
+    }
+  };
+
+  console.log('will use', info.menuItemId, 'server');
```
jebeck: I'm not seeing this in the console, and it does feel really necessary to see in the console when you've switched environments. Can we fix?

jebeck:
Mostly this looks fine to me, but I have a few comments/questions (and also the comment above about fixing the `console.log`):

- When I deleted all my versions of the uploader from my Chrome apps and tried to construct the tidepool.org/uploader link from the link I get out of blip currently (app ID and token from the end of the URL), I got an error instead of a redirect to the Chrome Store (but I'm logged in properly and can get there just fine from the current blip link). Is that expected? Also there isn't yet a blip PR to change the link there, right? It did properly launch the loaded unpacked extension when I tested before I deleted that.

- I think I'd prefer it if we just *prevent* devs from setting ENV variables since it won't be reflected in the initial context menu; I'd rather we stick to just *one* way of setting the environment (i.e., the context menu), but my feelings around this are not super strong. @jh-bate do you have thoughts?

- A limitation of the way this works is that you can log in on one environment, *then* switch servers after log in and try to upload. All the POSTs fail with 401s, of course. Not sure if we care about this. It *should* only affect devs, but there's a chance overly ambitious users will find the context menu and fiddle with things, then kill their ability to upload, etc. If it's easy to make opening the context menu even harder (⌘+right click, for instance), that would be fine by me.

main.js
kentquirk: There are two consoles -- one for index, and one for "background" -- which is main.js. This console message gets printed there.

kentquirk:
* The uploader, when it's installed, registers to handle urls that look like "tidepool.org/uploader?token=blah". The old Chrome store URL is now on that page on tidepool.org. There's not yet a PR for changing Blip to use that; thanks for reminding me. The trello card is [here.](https://trello.com/c/EUBbDXD6)
* I think there's still a reasonable case for doing a build with other than the default environment (mainly so  you don't have to change every time you restart), which is why I left it. 
* I haven't yet found a way to restrict the right-click menu to the login screen. I think there probably is one -- but I'd rather not hold up this PR for that. [Trello.](https://trello.com/c/t5Vz6lwQ)

kentquirk:
@jebeck -- I had what I thought were race condition issues -- so I fixed it properly. It turns out that wasn't my problem...but it's better for it.

I do think this is ready for merge now.

kentquirk:
@jebeck -- I've merged in the stuff from #49 and also put in one more tweak that should make things fit better.

kentquirk:
I'm merging this so we can get it tested.

PR #51, opened 2015-01-16
use sundial everywhere

jebeck:
Ready for review @kentquirk 

I ended up doing Dexcom (and a small change in Abbott to fix the weirdness you discovered last week) because I was on a roll and there really wasn't that much to change. I did leave out Asante because I don't know what other changes might have to be made first when we get back to that - didn't want it to be wasted effort.

Double-checked the Dexcom changes on my own Dex and checked the Insulet, but of course couldn't test the Abbott.

Dependent on [sundial PR 12](https://github.com/tidepool-org/sundial/pull/12) - use `npm link` until that's reviewed, merged, and pubbed to npm.


lib/drivers/abbottFreeStyle.js
```diff
@@ -34,9 +34,9 @@ module.exports = function (config) {
                         '[012][0-9]:[0-5][0-9](?::[0-6][0-9])?)';
   function parseDateTime(match, index) {
     var fmt = 'MMMM DD YYYY HH:mm:ss';
-    var parsed = sundial.parseAndApplyTimezone(match[index], fmt);
+    var parsed = sundial.parseFromFormat(match[index], fmt);
     var dev = sundial.formatDeviceTime(parsed);
-    var utc = sundial.parseAndApplyTimezone(match[index], fmt, cfg.timezone).toISOString();
+    var utc = sundial.applyTimezone(dev, 'US/Eastern').toISOString();
```
kentquirk: Why is this hardcoded to 'US/Eastern' instead of cfg.timezone?

kentquirk:
This now passes tests since I deployed the new sundial and set the version number here. Feel free to merge once you've dealt with the question about US/Eastern.

lib/drivers/abbottFreeStyle.js
jebeck: Because I was testing in the console and C&P-ed wrong. o_O Good catch.

PR #53, opened 2015-01-21
Kjq/disable menu

kentquirk:
This removes the right-click server selection menu after logging in.

If you log out, it doesn't come back -- but I consider this acceptable as it's only a feature for us as developers; all you have to do is reload.

@jebeck -- pretty trivial.

### DIFF
--- a/lib/state/appActions.js
+++ b/lib/state/appActions.js
@@ -14,6 +14,7 @@
  * not, you can obtain one from Tidepool Project at tidepool.org.
  * == BSD2 LICENSE ==
  */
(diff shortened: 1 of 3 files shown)


kentquirk:
I guess I had a version change already committed on my branch. We'll need to make sure we don't go backwards when this is merged.

jebeck:
Can you fix the tests @kentquirk? Should just be checking for availability of `chrome` before trying to remove context menu, since `chrome` doesn't exist in our test environment(s). Otherwise LGTM.

PR #54, opened 2015-01-23
Kjq/dexcom calibration

kentquirk:
@jebeck -- this adds the calibration records to dexcom.

They're a lot like the glucose records in that they use the same API to fetch them, so I generalized some of the functions to be reusable, then added a deviceMeta calibration to objectBuilder.

The records as fetched by octopus look like this:

      {
        "id" : "r3u1gr719ur00ldsrdbl8gfohleg66kp",
        "_active" : true,
        "createdTime" : "2015-01-23T03:53:12.028Z",
        "uploadId" : "7udfIBEpSQ2QEFsZ4u14MF7a3yQ=",
        "_version" : 0,
        "subType" : "calibration",
        "time" : "2014-08-20T15:09:30.000Z",
        "deviceId" : "Dexcom G4 Receiver SM33821888",
        "source" : "device",
        "timezoneOffset" : -420,
        "type" : "deviceMeta",
        "units" : "mg/dL",
        "value" : 9.713808984329683,
        "_groupId" : "d7a69a917c",
        "deviceTime" : "2014-08-20T08:09:30"
      },

 My big question is that there is a *third* time value on calibration records. The documentation says this:

 `public UInt32 MeterTime;` // System time that meter value was taken (as indicated by user during manual entry).

I think they're saying that the user has entered a time manually -- and I'm not at all sure what to do with that time. It doesn't seem to be a useful value. So far, i'm just ignoring it but would value your opinion.

### DIFF
--- a/lib/drivers/dexcomDriver.js
+++ b/lib/drivers/dexcomDriver.js
@@ -199,7 +199,7 @@ module.exports = function (config) {
 
     for (var i = 0; i < header.nrecs; ++i) {
       var rec = struct.unpack(data, ctr, format, [
-        'systemSeconds', 'displaySeconds', 'meterValue', 'meterTime', 'crc'
+        'systemSeconds', 'displaySeconds', 'meterValue', 'meterTimeSeconds', 'crc'
       ]);
 
       rec.systemTimeMsec = BASE_DATE_DEVICE + 1000 * rec.systemSeconds;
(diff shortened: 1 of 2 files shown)


jebeck:
This looks good, and I just tried it from my receiver locally, and all looks well. Interestingly, it seems the receiver stores far more than 30 days of calibration records. (I get them going back to November.)

Re: the meter time - I suppose that's because the calibration UI has a timestamp that is pre-filled to the device time but will inevitably vary (usually by only some seconds) from the device time when you actually submit the calibration because it takes a bit of time to do the data entry. Not having a working system at the moment, I can't tell you whether that timestamp is ever user-editable. It might be. I just did some logging of the `meterTime` locally, and it looks like it is a Dexcom internal time, which we're not preserving anywhere else (because it's generally useless, as it is not the display/local time but has an unknown offset from UTC), so I'd say let's continue to do nothing with it here.

This is good to merge.

PR #55, opened 2015-01-23
Kjq/restore dexcom trend

kentquirk:
This restores the trend arrow information to Dexcom CBG records now that it has been approved by Dexcom. 

This branch already has #54 merged into it -- so should be merged after that one is. 

@jebeck ?

### DIFF
--- a/lib/drivers/dexcomDriver.js
+++ b/lib/drivers/dexcomDriver.js
@@ -590,12 +590,13 @@ module.exports = function (config) {
         // special and out-of-range values are not posted for now
         continue;
       }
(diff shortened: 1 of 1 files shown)


jebeck:
This also looks good. However, `trend` is not in our data model for `cbg` records. Can you do a PR to add it? Or potentially it should go in a `payload` object, as the number/type/name of arrows will vary by device, and the data model should be device-agnostic...

kentquirk:
I liked your idea to move it into payload, so I did. I'll add a note to the data model about it, but as it's now device-specific, it doesn't need to be a big deal.

PR #57, opened 2015-01-23
Add single-precision float support to struct.js

mgranberry:
This adds support for parsing single-precision floating-point numbers to struct.js

### DIFF
--- a/lib/struct.js
+++ b/lib/struct.js
@@ -139,7 +139,7 @@ module.exports = function() {
       dataview.setUint8(2, b[st+2]);
       dataview.setUint8(3, b[st+3]);
       return dataview.getFloat32(0, true);
-    }
+    };
     var extractBEFloat = function(b, st){
       var buffer = new ArrayBuffer(4);
       var dataview = new DataView(buffer);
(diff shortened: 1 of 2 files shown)


kentquirk:
Hmm...DataView is handy. There's one issue, though -- DataView's results are dependent on the endianness of the machine it runs on.

Given that all current macs and pcs are little-endian (no we don't miss the 68K), it's not really an issue today. But in a hypothetical future where the uploader might run on an ARM-based machine, there's a possibility that a big-endian architecture might rear its head again.

I believe that all the pre-existing code is not sensitive to the endianness of the platform it's running on, but this new code is.

In my ideal world, there would be a test during initialization to decide if the local architecture is big-endian, and it would swap extractFloat with extractBEFloat, and the same with store. An alternative would be a test during initialization that just throws an error if it's run on a big-endian machine. At the very least, it deserves a comment in the source about it.

mgranberry:
I wish I had a big-endian machine with enough RAM to test on, but these methods use the endian-safe setFloat32(value, endianness) and getFloat32(value, endianness) methods on DataView.  It is my understanding that typed arrays are native-endian, which is why I used DataView in the first place.  

I actually have another implementation that does it purely in math before I stumbled on DataView, but it requires more recent browser support for Math.fround() to arrive at bit-exact float decoding and this appears to be exactly what DataView is intended to do.

kentquirk:
Doh, my mistake. Thanks!

In that case, LGTM, will merge.

PR #58, opened 2015-01-24
Remove source attribute

kentquirk:
@jebeck -- these are the changes to remove 'source' as an attribute everywhere from the uploader. Note that appActions and appState have something they call 'source' but it's different, so I didn't touch them. Also removed it from all the tests where it was mentioned. Tests still pass.


kentquirk:
I decided to leave source in the setDefaults for objectBuilder. That way, we can still set it for Carelink and use the same test for whether old Carelink data needs to be deleted.

jebeck:
OK, this looks good. I'll leave it to you to merge when you're ready with all the other elements.

PR #59, opened 2015-01-25
Add device data

kentquirk:
Please do not merge yet.

This is the first pass of some code to add the device data to uploads; it will need to be merged after #58 and I've made it so the merge will cause a conflict, because CareLink will need to restore the source attribute.

@jebeck -- I'd like to hear what should go in the carelink fields for type, mfr, series, model. Can do it by talking if that's easier.


kentquirk:
@jebeck -- I believe this is now ready for merge. It passes all tests and seems to be working, and is current from master.

kentquirk:
@jebeck -- Actually -- carelink records are not being filled in with useful device info in the uploader record. I'd really appreciate your thoughts/advice or even code. carelinkDriver.js, line 247.

jebeck:
OK @kentquirk the device info is filled out as I believe we agreed for CareLink (incl. tests for documentation of how it works).

*However*, this code does not work for CareLink or Dexcom. I tried it and still saw old-style `type: upload` records for both in the resulting data exported from blip and then traced it as far as [here](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/core/api.js#L176) where I found that `sessionInfo` is not what we upload as the `type: upload` record, an `uploadSessionMeta` object is. I did not yet trace back to figure out where that object is generated.

kentquirk:
That (for some reason) is in platform-client.

It's confusing enough that you've now convinced me to actually pull it back into the uploader. I'm going to do that as part of this PR of the uploader, and then do a separate one for platform-client.

jh-bate:
@kentquirk are you going to remove the upload session creation from platform-client? Just need to think about the fact that it could/will not only be the uploader that push's data into the platform.

kentquirk:
@jh-bate Yes. I realize there will someday be other clients. But it's more important to keep the object builder together than it is to put that one feature in the platform-client. Right now, we've been having to rev 3 repositories every time we change the record formats; sometimes 4. At the time we decide to split platform-client into smaller pieces, we can move the entire objectBuilder into the public part of it, but for now we think it's better off unified.

PR #60, opened 2015-01-27
require units on cbg, smbg, wizard events

jebeck:
Here's the small builder change @kentquirk 

I made another small change because I happened to notice it - it was very bad that we were producing default `recommended` objects with values of 0 (because a 0 suggestion is a suggestion)! This change required a jellyfish change though: https://github.com/tidepool-org/jellyfish/pull/46

### DIFF
--- a/lib/objectBuilder.js
+++ b/lib/objectBuilder.js
@@ -100,15 +100,6 @@ module.exports = function () {
     };
   }
 
-  function _makeWithValue(typename) {
-    var rec = _.assign(_createObject(), deviceInfo, {
-      type: typename,
-      value: REQUIRED
-    });
-    rec._bindProps();
-    return rec;
-  }
-
   function makeSMBG() {
     var rec = _.assign(_createObject(), deviceInfo, {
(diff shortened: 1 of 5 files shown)


lib/objectBuilder.js
```diff
@@ -110,11 +110,23 @@ module.exports = function () {
...
+    var rec = _.assign(_createObject(), deviceInfo, {
+      type: 'cbg',
+      value: REQUIRED,
+      units: REQUIRED
+    });
+    rec._bindProps();
+    return rec;
   }
 
   function makeNote() {
```
kentquirk: Now that _makeWithValue is used in only one place, could you also just remove it from makeNote and eliminate the function?

kentquirk:
Tests are now failing. Please fix?

lib/objectBuilder.js
jebeck: Yeah, can do.

kentquirk:
I piled on the changes to use the new units behavior in dexcom and abbott.

PR #61, opened 2015-02-02
Kjq/dexcom hi lo

kentquirk:
This makes the dexcom driver store highs and lows and annotate them appropriately. @jebeck ?


jebeck:
LGTM. I'll let you merge and build. I just merged the relevant data model PR so that the docs are up-to-date.

PR #62, opened 2015-02-03
Kjq/add ketones

kentquirk:
@jebeck -- I can't test this as I don't have a PC here. But if you give it a once-over I'll do a test when I get home tomorrow.

### DIFF
--- a/lib/drivers/abbottFreeStyle.js
+++ b/lib/drivers/abbottFreeStyle.js
@@ -410,6 +410,15 @@ module.exports = function (config) {
         dataToPost.push(smbg);
       } else if (datum.readingType === 'ketone') {
         // for now we don't post these but we'll want to someday
+        var bloodKetone = cfg.builder.makeBloodKetone()
+          .with_value(datum.reading)
+          .with_deviceTime(datum.datetime.dev)
+          .with_timezoneOffset(datum.datetime.offset)
+          .with_time(datum.datetime.utc)
+          .with_units('mg/dL')
+          .done();
+        dataToPost.push(bloodKetone);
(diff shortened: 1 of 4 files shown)


lib/drivers/abbottFreeStyle.js
```diff
@@ -410,6 +410,15 @@ module.exports = function (config) {
         dataToPost.push(smbg);
       } else if (datum.readingType === 'ketone') {
         // for now we don't post these but we'll want to someday
+        var bloodKetone = cfg.builder.makeBloodKetone()
+          .with_value(datum.reading)
+          .with_deviceTime(datum.datetime.dev)
+          .with_timezoneOffset(datum.datetime.offset)
+          .with_time(datum.datetime.utc)
+          .with_units('mg/dL')
```
jebeck: Blood ketones are always in `mmol/L`.

jebeck:
Other than comment about units for blood ketones, LGTM.

kentquirk:
@jebeck, according to the spec from Freestyle:

"
[Glucose/Ketone Log Data] = 30 characters in the form “RRR  MMMM DD YYYY HH:MM C 0xSS” if Log is not empty, where:
•	RRR = glucose or ketone in units of mg/dL with leading zeros if less than 100. 
"

The meter I have is fetching three ketone readings, with values 8, 34, and 16 respectively.

Do you think this sounds plausible?

kentquirk:
Ok, it's verified. The device displays 0.4 mmol/L on the screen, but actually stores and reports 8 mg/dL through the software. I'm going to go ahead and merge this since it is working.

PR #63, opened 2015-02-03
Convert ketones to mmol/L before upload

kentquirk:


lib/drivers/abbottFreeStyle.js
```diff
@@ -409,13 +409,16 @@ module.exports = function (config) {
           .done();
         dataToPost.push(smbg);
       } else if (datum.readingType === 'ketone') {
-        // for now we don't post these but we'll want to someday
+        // These meters store ketone values as 18 * the mmol/L value of the ketones
+        // Because this is specific to these devices, we convert back
+        // in the driver and upload the data as mmol/L.
+        var ketoneValue = datum.reading / 18.0;
```
jebeck: Aren't we going to get some floating point oddness? Does the spec give any info on significant digits of ketone readings?

kentquirk:
@jebeck -- besides the conversion, now annotating out-of-range bgs and ketones.

lib/drivers/abbottFreeStyle.js
```diff
@@ -100,20 +100,42 @@ module.exports = function (config) {
       handler: function(match, packet) {
         var entry = {};
         var reading = match[1];
-        if (reading == 'HI ') {
-          entry.reading = 501;
-        } else if (reading == 'LO ') {
-          entry.reading = 20;
```
jebeck: If the threshold is 20, the value should be 19.

lib/objectBuilder.js
```diff
@@ -104,7 +104,8 @@ module.exports = function () {
     var rec = _.assign(_createObject(), deviceInfo, {
       type: 'smbg',
       value: REQUIRED,
-      units: REQUIRED
+      units: REQUIRED,
+      annotation: OPTIONAL
```
jebeck: Hmm, in the other drivers I've abstracted out some annotation functions instead of adding them as optional in the builder. This makes quite a bit of sense because the annotateEvent function needs to preserve existing annotations and check for duplicates before adding a new one. (This kind of thing can happen in basals.)

But we could just merge this and update later, since the common annotations functions are all the way over on my Insulet branch right now?

jebeck:
Other than comment about annotations, LGTM.

kentquirk:
Yes....I saw your code but it's in a branch and I didn't think cherry-picking it was worth it. So we can fix after you merge insulet.

PR #64, opened 2015-02-03
Sk/uploader pw change

ursooperduper:
@kentquirk 

This PR changes the password reset link for Blip which will take a logged in user to their account password to update their password, while taking logged-out users to the password reset request.

I didn't have a great way to test this in my local dev environment, so we'll need to do some testing on devel to ensure this is all working as designed.  

Resolves: https://trello.com/c/ilkdLJKh


kentquirk:
LGTM. Checked to make sure it agrees with blip, and it does, so I'm merging.

PR #65, opened 2015-02-04
[WIP] Tandem support

mgranberry:
This is still pretty rough and makes no attempt to decode log records or profiles, but it can send most commands and decode the responses.

Do not merge.


lib/drivers/tandemTslimDriver.js
```diff
@@ -0,0 +1,839 @@
...
+var struct = require('./../struct.js')();
+var sundial = require('sundial');
+
+module.exports = function (config) {
+    var cfg = config;
+
+    var SYNC_BYTE = 0x55;
+
+    var RESPONSES = {
+        81: {
```
kentquirk: I know this is incomplete, but why is this a raw number rather than VERSION_TE ?

If you're just trying to index by number, a helper function that searches for the responses by value would probably be clearer.

lib/drivers/tandemTslimDriver.js
```diff
@@ -0,0 +1,839 @@
...
+        0x67: { // LID_USB_ENUMERATED
+            value: 0x67,
+            name: 'USB Enumerated Event',
+            format: 'f............',
+            fields: ['negotiated_current_mA']
+        }
+    };
+
+    // This is a particularly weak checksum algorithm but that's what Insulet and Tandem use...
+    var weakChecksum = function (bytes, offset, count) {
```
kentquirk: Hmm. Maybe we should move this into crc.js. :(

lib/drivers/tandemTslimDriver.js
```diff
@@ -0,0 +1,839 @@
...
+        console.log('requesting log size');
+        var retval = [];
+        var entries;
+        var end_seq;
+        var seq;
+        function iterate(err, result) {
+            cfg.deviceComms.emitLog(true);
+            // TODO handle err with retransmissions
+            if (result.valid && seq < end_seq) {
+                if (result.descriptor === 125) {
```
kentquirk: Please define and use constants for stuff like this.

jebeck:
I'm gonna assume this is being superseded by #159 and closing.

PR #67, opened 2015-02-10
Fix md5 when it's used from node

kentquirk:
@jebeck -- I think this will fix your md5 issue.


jebeck:
Does the trick, thanks! Merging, and I cherry-picked it onto my branch, because gross conflicts that I didn't feel like sorting through yet (from the groups work)?

PR #68, opened 2015-02-11
add Insulet OmniPod

jebeck:
WIP - **don't merge**.

Opening the PR for ease of conversation with @mgranberry


jebeck:
@kentquirk I did the work to merge in updates from master and resolve conflicts. That included what we discussed before about replacing the addition of annotations in the builder with the common util I had here. So I did that for the PX and Dexcom in commit 62e32d87d90ea6fb34fe6b6135cfd1707c6cd7f8. Want to take a look at just that now (and test if you're able) to make sure all is well?

jebeck:
@kentquirk I believe this is now pretty much ready for review. I'm still working on some additional simulator tests, but we should probably start the review process. I'm going to call out a few things myself inline, so look for that.

lib/insulet/common.js
```diff
@@ -0,0 +1,22 @@
...
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ * 
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+module.exports = {
+  fixFloatingPoint: function(n, precision) {
+    return parseFloat((n).toFixed(precision));
```
jebeck: I'm not particularly attached to this method of avoiding floating point weirdness, but I do feel strongly about not storing floating point weirdness. If another method is better, LMK.

lib/drivers/insuletDriver.js
```diff
@@ -614,18 +644,47 @@ module.exports = function (config) {
     });
   };
 
+  // these aren't history records, so we have to find them separately
+  var findPumpAlarmRecords = function (recordlist) {
+    var result = [];
+    for (var i = 0; i < recordlist.length; ++i) {
+      // TODO: check with KJQ about log_index of -1???
+      // doesn't have a timestamp, so I think it's not a valid record
```
jebeck: @kentquirk want to weigh in on this one?

kentquirk: I think if it doesn't have a timestamp, it must not be valid. I can't imagine what we would do with it.

lib/insulet/common.js
kentquirk: This is exactly equivalent in results to my technique -- Math.floor(x * 100 + 0.5) / 100 -- it's quite a bit more expensive because yours does a conversion to and from strings.

docs/PWDSimulators.md
```diff
@@ -0,0 +1,43 @@
...
+- `basal`
+   + determining and setting `duration`
+   + determining and setting the `suppressed` basal object(s) for `temp` and `suspend` basals
+   + adding the `previous` event
+- `bolus`
+   + connecting `normal` and `square` components into a dual-wave bolus where applicable
+   + connecting boluses with accompanying `wizard` events, although this is more commonly accomplished in the device driver
+   + adding an `expectedNormal` and/or `expectedExtended` and `expectedDuration` when a bolus is terminated early (i.e., bolus volume programmed !== bolus volume delivered)
+- `deviceMeta` sub-type `status`
+   + adding the `previous` event
```
kentquirk: I love this document. Thanks.

lib/core/api.js
```diff
@@ -171,7 +171,8 @@ api.upload.accounts = function(happyCb, sadCb) {
 };
 
 function postBlockToPlatform(data, groupId, callback) {
-  if (!callback) {
+
+  if (typeof groupId === 'function') {
     callback = groupId;
```
kentquirk: I thought I said this to Jamie recently, so I'm sorry if I'm repeating myself. This form of optional arguments is challenging to maintain -- it's not obvious that it's an optional argument. It's useful in an API when you have third parties that may be using it -- but in this case, all the clients are within this app. I'd rather fix up all the references and remove the optionality.

lib/core/api.js
```diff
@@ -206,20 +208,31 @@ function postBlockToPlatform(data, groupId, callback) {
...
+  }
 }
 
 /*
  * process the data sending it to the platform in blocks and feed back progress to the calling function
  */
 api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb) {
-  if (!cb) {
+
+  if (typeof groupId === 'function') {
```
kentquirk: see above

lib/drivers/insuletDriver.js
```diff
@@ -19,17 +19,37 @@ var _ = require('lodash');
...
   };
 
+  // Insulet reports percentage temp basals as +/- integers [-100, 100+]
+  // i.e., -50 for 50% of the scheduled basal rate
+  var convertPercentage = function(x) {
+    // 0 means there *wasn't* a temp basal in this record
+    if (x === 0) {
+      return null;
+    }
+    return 1.0 + x / 100;
```
kentquirk: Although this is safe in JS because JS doesn't have integers, it still makes me nervous because of this:
x = 34;
y=1.0 + x / 100;
If you ran that code in any other language but JS, the answer would be 1.0, because the x/100 would take place in integer math (try it in Python or Ruby). Please replace the 100 with 100.0?

lib/drivers/insuletDriver.js
```diff
@@ -19,17 +19,37 @@ var _ = require('lodash');
...
+
   // all insulin unit readings are in .01 unit increments, so we divide by 100.0 to get units
   // (multiplying by 0.01 tends to cause floating point issues)
   var toUnits = function (x) {
     return x / 100.0;
   };
 
+  // Insulet reports percentage temp basals as +/- integers [-100, 100+]
+  // i.e., -50 for 50% of the scheduled basal rate
+  var convertPercentage = function(x) {
```
kentquirk: Could we maybe get a better name for this? convertPercentageToBasalFactor, perhaps? Or something? Esp since it returns null sometimes, when it's called it would help to know more about what it does.

lib/drivers/insuletDriver.js
```diff
@@ -257,7 +272,7 @@ module.exports = function (config) {
       'time_left_seconds'
     ], postprocess: function (rec) {
       rec.insulin_units_left = toUnits(rec.insulin_left);
-      rec.time_left_msec = rec.time_left_seconds * sundial.SEC_TO_MSEC;
+      rec.time_left_msec = rec.time_left_seconds * sundial.MIN_TO_MSEC;
```
kentquirk: Urk. If that time isn't in seconds (sorry!) its name should be changed.

lib/drivers/insuletDriver.js
```diff
@@ -614,18 +644,47 @@ module.exports = function (config) {
...
-    // we need to see if two boluses come from the same calculation (wizard) record
+    // we need to see if two (or more!) boluses come from the same calculation (wizard) record
     // if so, they're actually a dual bolus, and need to be consolidated
     // so we create a table of backlinks from the calc records to the bolus records that
     // refer to them
     var wizRecords = {};
     for (var b = 0; b < bolusrecs.length; ++b) {
-      var bolus = data.log_records[bolusrecs[b]];
-      var wiz_idx = bolusrecs[b] + bolus.detail.calculation_record_offset;
+      var bolus = data.log_records[bolusrecs[b]], wiz_idx;
```
kentquirk: Whoa -- I'm not sure what you intended here, but that's probably not what you got. Or if it is, I don't understand why the ", wiz_idx" is on this instruction.

lib/drivers/insuletDriver.js
```diff
@@ -637,6 +696,266 @@ module.exports = function (config) {
...
+        if (alarm.alarm.alarm_text != null) {
+          alarmValue = alarm.alarm.alarm;
+        }
+        else {
+          postalarm = null;
+        }
+      }
+      switch (alarmValue) {
+        // History - ALARM
+        case 21:
```
kentquirk: Can you please use the constants defined for the alarms?

case ALARM_TYPES.ALARM_KEY.value:

lib/drivers/insuletDriver.js
```diff
@@ -637,6 +696,266 @@ module.exports = function (config) {
...
+          postalarm = postalarm.with_alarmType('occlusion')
+            .with_payload({
+              alarmText: getNameForValue(ALARM_TYPES, alarmValue),
+              explanation: 'pump site occluded',
+              stopsDelivery: true
+            })
+            .with_status(postsuspend)
+            .done();
+          break;
+        case 18:
```
kentquirk: You could also consolidate a lot of these similar cases by extending the ALARM_TYPES data structure with a few more fields like "explanation" and "tidepoolAlarmCode".

lib/drivers/insuletDriver.js
```diff
@@ -644,10 +963,15 @@ module.exports = function (config) {
     ]);
     var wizRecords = linkWizardRecords(data, bolusrecs);
     var postrecords = [];
+
     for (var b = 0; b < bolusrecs.length; ++b) {
       var bolus = data.log_records[bolusrecs[b]];
       var wiz_idx = -1;
```
kentquirk: This -1 no longer has a meaning, since wiz_idx will always be reassigned below.

lib/drivers/insuletDriver.js
```diff
@@ -747,28 +1149,207 @@ module.exports = function (config) {
...
+      }
+    }
+    var lastBasal = postrecords[postrecords.length - 1];
+    annotate.annotateEvent(lastBasal, 'insulet/basal/fabricated-from-schedule');
+    return records.concat(postrecords);
+  };
+
+  // it turns out these records may not actually be implemented
+  // occlusions are only represented through pump advisory and hazard alarms
+  // TODO: maybe we should just delete this code?
```
kentquirk: Perhaps we should query Insulet on this one?

lib/insulet/insuletSimulator.js
```diff
@@ -0,0 +1,376 @@
...
+            // a resume is probably still impending
+            // but if we've already moved on to a basal that's not deliveryType: suspend
+            // then we need to wipe the slate clean
+            if (event.deliveryType !== 'suspend') {
+              setCurrStatus(null);
+            }
+          }
+        }
+        // currBasal.duration is a string when the builder object hasn't had
+        // .with_duration called yet (i.e., it's the string `**REQUIRED**`)
```
kentquirk: Should we perhaps add a helper function to objectBuilder that would let you call something like isAssigned('duration'), rather than depending on the implementation details of objectBuilder?

kentquirk:
I've completed my review. I don't think anything is really major but I'd like to see a few changes, please.

lib/core/api.js
jebeck: Yeah, I think this is where you commented on this before. Not quite sure why it's in the diff, but I'll investigate. I think it was something we put on hold to change because it will require a change to platform-client? Or am I misremembering?

I agree that the hidden optionality is confusing.

jebeck:
Thanks @kentquirk. I'm compiling a list of TODOs; will respond in more detail to specific issues if there's need.

lib/drivers/insuletDriver.js
jebeck: Done.

lib/drivers/insuletDriver.js
jebeck: I was confused by the `Factor` in your name, so I did `convertTempPercentageToNetBasal`.

lib/drivers/insuletDriver.js
jebeck: Done. It wasn't your fault - the spec was incorrect.

lib/drivers/insuletDriver.js
jebeck: OK, I'll put on separate lines.

lib/drivers/insuletDriver.js
jebeck: Yeah, good idea. Probably also time to add my few additional comments on the spec - will talk to @brandonarbiter about doing that.

lib/insulet/insuletSimulator.js
jebeck: Good idea. I'll add one.

lib/drivers/insuletDriver.js
jebeck: Yeah, I cleaned up the `buildAlarmRecords` with this suggestion and the above - much better, thanks :)

kentquirk:
Approved. Merge when ready.

PR #70, opened 2015-02-13
Jhbate/cli

jh-bate:
* cleanup test for groupId
* send to platform differently based on if we have the groupId specified or not

### DIFF
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -171,7 +171,8 @@ api.upload.accounts = function(happyCb, sadCb) {
 };
 
 function postBlockToPlatform(data, groupId, callback) {
-  if (!callback) {
+
+  if (typeof groupId === 'function') {
     callback = groupId;
     groupId = null;
   }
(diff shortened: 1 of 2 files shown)


jebeck:
Makes sense and works. Merging. Ta @jh-bate

lib/core/api.js
```diff
@@ -206,20 +208,31 @@ function postBlockToPlatform(data, groupId, callback) {
...
-
-  tidepool.uploadDeviceDataForUser(data, groupId, function(err, result) {
-    if (err) {
-      return sad(err, err);
-    }
-    return happy(result);
-  });
+  if(!_.isEmpty(groupId)){
+    api.log('postBlockToPlatform: for other user');
+    tidepool.uploadDeviceDataForUser(data, groupId, function(err, result) {
```
kentquirk: I'm not sure how I missed this, but I really dislike this pattern where uploadDeviceDataForUser hits a different endpoint depending on whether an argument in the middle of the argument list is present or absent.

I would have handled that in platform-client by leaving uploadDeviceData with no groupId, and uploadDeviceDataForUser with a groupId, and then having them both use a common helper routine.

This JavaScript pattern of testing the type of an argument and then shifting arguments over gives me nightmares. It's a pain to document and a pain to use. 

I'd like to see this fixed but I also understand if that ship has sailed for now.

jebeck: I think the ship has sailed at least for now; I don't have a very strong opinion here and would like to see if @jh-bate does, but I hope this branch will be merged before he gets back from his weekend. If you like, maybe create a Trello card so we don't forget to at least discuss it?

PR #71, opened 2015-02-13
clear React warnings and sort upload groups with logged-in user at top

jebeck:
@jh-bate just small changes I noticed while looking at other uploader React stuff.


jh-bate:
LGTM

PR #72, opened 2015-02-15
update CareLink deviceMeta status reasons to accord with revised data model

jebeck:
Dependency hell: required jellyfish v0.8.0 for Insulet demo broke CareLink. This fixes (although it also required additional tweaks to jellyfish: https://github.com/tidepool-org/jellyfish/pull/54)

This is already merged into the branch I'll be building for the devel Chrome store, and there will be a few small branches like that this weekend. I will be creating a list (incl. order in which I think they should be reviewed) by end of weekend.

### DIFF
--- a/lib/carelink/carelinkSimulator.js
+++ b/lib/carelink/carelinkSimulator.js
@@ -312,7 +312,7 @@ exports.make = function(config){
        * during an LGS. We are already handling these in our basal clock, so we just
        * push the clock forward and exit early when we come across one.
        */
-      if (currStatus != null && currStatus.status === 'suspended' && currStatus.reason === 'low_glucose') {
+      if (currStatus != null && currStatus.status === 'suspended' && (currStatus.payload && currStatus.payload.cause === 'low_glucose')) {
         ensureTimestamp(event);
         pushBasalClockForward(event.time);
         return;
@@ -552,7 +552,8 @@ exports.make = function(config){
         .with_time(event.time)
         .with_timezoneOffset(event.timezoneOffset)
         .with_deviceTime(event.deviceTime)
-        .with_reason(event.reason);
+        .with_reason(event.reason)
+        .with_payload(event.payload);
 
       if (event.deviceId != null) {
         deviceMetaResumeBuilder.set('deviceId', event.deviceId);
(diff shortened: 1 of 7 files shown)


jebeck:
Hey @jh-bate now that jellyfish v0.8.1 is deployed on devel and staging, master of the uploader will be broken for CareLink until this is merged. Can you take a look at it tonight?

lib/carelink/carelinkSimulator.js
```diff
@@ -312,7 +312,7 @@ exports.make = function(config){
        * during an LGS. We are already handling these in our basal clock, so we just
        * push the clock forward and exit early when we come across one.
        */
-      if (currStatus != null && currStatus.status === 'suspended' && currStatus.reason === 'low_glucose') {
+      if (currStatus != null && currStatus.status === 'suspended' && (currStatus.payload && currStatus.payload.cause === 'low_glucose')) {
```
jh-bate: seems an awful one liner check but prob good enough for now

lib/carelink/carelinkSimulator.js
```diff
@@ -861,7 +867,7 @@ exports.make = function(config){
       var event = ensureTimestamp(combineArguments(arguments));
       // threshold suspends involve several (usually 3-4) suspend events in succession
       // we only care about the first, so if the currStatus is already `suspended`, we return early
-      if (currStatus != null && currStatus.status === 'suspended' && event.reason === 'low_glucose') {
+      if (currStatus != null && currStatus.status === 'suspended' && (event.payload && event.payload.cause === 'low_glucose')) {
```
jh-bate: ditto on the one liner check

lib/carelink/suspend.js
```diff
@@ -33,23 +33,68 @@ module.exports = function (timezone) {
...
-                return { status: 'suspended', reason: 'low_glucose' };
-
+                return {
+                  status: 'suspended',
+                  reason: {
+                    suspended: 'automatic'
+                  },
+                  payload: {
+                    cause: 'low_glucose',
+                    code: reason
```
jh-bate: so the code references the reason above?

jh-bate:
made some comments will chat asap

lib/carelink/suspend.js
jebeck: `reason` is the raw CareLink string, so I named the attr `code`. Does that make it clearer?

PR #74, opened 2015-02-18
add device selection (and OmniPod)

jebeck:
This is getting close, so I'm opening the PR.


jebeck:
This is ready for review. If you want to review #68 first just for comments on OmniPod code @kentquirk, we could merge that to master and make the diff here simpler, for looking at the device selection code.

Once this is on master, we can comment out Insulet from the array of `uploads` in `lib/state/appState.js` and bump version for next likely Jaeb and prod deploy.

The `RESTRICT_DRIVERS` change is on your Asante branch; let's talk about how to proceed with merging.

jebeck:
Ignore previous comment about `RESTRICT_DRIVERS` - you'll see I cherry-picked commit. Merge conflicts are *only* because of version number change in package.json and manifest.json.

Ready for review.

kentquirk:
This seems odd to me:

image

There's a huge swath of empty space at the bottom of the screen, the window is as small as it will go, and the middle section has a scrollbar that wouldn't be needed if the box were bigger.

kentquirk:
Also...the device selection screen will need to grow a scrollbar, although we have time to add that later as it can grow to 7 devices before we need it.

image

kentquirk:
Other than the issues above, which are more UI than technical, I'm happy with this. You can merge at will, @jebeck,  or fix either or both of these first -- your call.

jebeck:
The extra space under the footer is a known issue that I believe @ursooperduper is working on. I explained how the height for the scrollable area got chosen by me over [here](https://github.com/tidepool-org/chrome-uploader/pull/75#discussion_r25045145) (short version: I was developing with accounts that have the groups dropdown visible, and in that case there really isn't as much wasted space).

Looking into adding scrolling into the device list before merge - good catch, I thought I'd already put it there.

jebeck:
OK, I ended up doing a bit more tweaking than I expected to address the issues @kentquirk raised. Much better now as far as wasted space - parametrized some dimensions in the CSS based on whether the groups dropdown is showing. I showed @HowardLook my changes live, and he approved, so merging.

@ursooperduper your PR should be updated from master after this comes in, please.

PR #75, opened 2015-02-18
Device selection UI styling

ursooperduper:
@jebeck 

This PR includes improved styling for: 
* Removing extra space between the footer and the bottom of the window. 
* Better differentiation of devices by adding a horizontal rule between items. 
* Updating button state to match the design. 

Resolves 3/4 items on the checklist here: https://trello.com/c/u1H9mqM1

### DIFF
--- a/lib/components/Upload.jsx
+++ b/lib/components/Upload.jsx
@@ -158,7 +158,7 @@ var Upload = React.createClass({
     return (
       <div className="Upload-button">
         <button
-          className="btn btn-secondary"
+          className="btn btn-primary"
           disabled={disabled}
           onClick={this.handleUpload}>{text}</button>
       </div>
--- a/styles/components/App.less
+++ b/styles/components/App.less
@@ -46,8 +46,7 @@
   border-radius: 2px;
   display: flex;
   flex-direction: column;
-  padding: 15px (30px - @App-mainPageBorder) 25px;
-  margin-bottom: 10px;
+  padding: 15px (30px - @App-mainPageBorder) 0;
 }
 
 .App-signup {
--- a/styles/components/App.less
+++ b/styles/components/App.less
@@ -35,7 +35,6 @@
 .App--login .App-page {
   width: 300px;
   height: 100%;
-  border: 1px solid red;
 }
 
 @App-mainPageBorder: 1px;
--- a/styles/components/App.less
+++ b/styles/components/App.less
@@ -19,7 +19,6 @@
 .App-header {
   padding: 0 25px;
   background: url(/T_icon_transparent.png) center center / 100px 100px no-repeat;
-
   min-height: 150px;
 }
 
@@ -30,12 +29,13 @@
   margin-left: auto;
   margin-right: auto;
   margin-bottom: 20px;
-  min-height: 100px;
-  max-height: 420px;
+  min-height: 470px;
 }
 
 .App--login .App-page {
   width: 300px;
+  height: 100%;
+  border: 1px solid red;
 }
 
 @App-mainPageBorder: 1px;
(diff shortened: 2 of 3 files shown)


lib/components/Upload.jsx
```diff
@@ -158,7 +158,7 @@ var Upload = React.createClass({
     return (
       <div className="Upload-button">
         <button
-          className="btn btn-secondary"
+          className="btn btn-primary"
```
jebeck: Can you include some screenshots of using the primary buttons for the uploads in the PR? My interpretation of my conversation with @HowardLook and @brandonarbiter was that they were looking for tweaks to the existing styles to bring out the state contrasts better, not switching to a different set, but it's possible that was just my interpretation.

styles/components/App.less
```diff
@@ -30,12 +29,12 @@
   margin-left: auto;
   margin-right: auto;
   margin-bottom: 20px;
-  min-height: 100px;
-  max-height: 420px;
+  min-height: 470px;
```
jebeck: Again I think there are undesirable side-effects to these changes - the footer is partially or completely cut off after login on some screens, for example.

styles/components/UploadList.less
```diff
@@ -34,14 +34,16 @@
...
   padding-right: 15px;
-  max-height: 260px;
+  min-height: 370px;
   overflow-y: scroll;
 }
 
 .UploadList-item {
-  margin-bottom: 30px;
+  padding: 15px 0;
+  border-bottom: 1px solid @gray-light;
```
jebeck: How do we feel about using a hairline separator for both separating the groups dropdown and between each upload list item? CC @HowardLook @brandonarbiter 

The separator is a little wider below the groups dropdown, but otherwise it's very similar (IMO).

screenshot 2015-02-18 22 37 54

HowardLook: I like the hairline.

I wonder if we can tighten up the vertical space between the type-ins and
buttons for each device group a bit, e.g. put username/password/import
closer together for CareLink, etc for choose file/upload for OmniPod.

lib/components/Upload.jsx
ursooperduper: chrome-uploader-upload-screen

The button style I chose was based on the design Sara and Brandon provided. Is there a new design I missed?

styles/components/App.less
ursooperduper: By screens, do you mean displays, like MacBook Air 13" vs 11"? If so, we should consider making the window small by default and allow people to resize it vertically. 

Also, there's a lot of wasted space in the Uploader right now. For example, the T icon mark is huge and could be much smaller. Likewise, it seems odd to include both the T icon and the Tidepool logo, you could remove the Tidepool logo and make the footer much smaller. But again, what's implemented in based on the approved design, so I don't want to overstep.

lib/components/Upload.jsx
jebeck: Ah, my bad. I didn't realize the button style that was in the version I branched off of wasn't to date with Sara's design.

styles/components/App.less
jebeck: I think there's misunderstanding here :/

I'll try to illustrate (from add-omnipod-select branch on origin). Here's what the UI looks like for someone who has multiple people they are allowed to upload to (so the groups dropdown is present) and uses a Medtronic pump, a Dexcom, and the PX meter (i.e., a likely Jaeb study participant).

screenshot 2015-02-19 17 51 53

The `max-height` I chose for the inner scrolling container was because of this case - I wanted to make sure that for such a person two devices are visible in their entirety, and they can scroll to see the third, but the third isn't cut off.

I neglected to notice, however, that there is still wasted space in the footer. If I can paraphrase @HowardLook's feedback, I think it was that we'd like to have the footer closer to the bottom of the app (and in the same place no matter where in the app you are - location in the app is what I actually meant by "screens") and try to increase the space given to the scrolling content. It would be great if three devices could fit in the upload list without engaging scrolling (maybe only possible for users that don't have permissions to upload to anyone else, so the groups dropdown isn't present).

Also FYI, I believe the decision to have two logos present was because with only the T logo there is no company identification anywhere in the app.

jebeck:
PR #81 cherry-picked from what was started here so that we could merge for a new build. Thanks for the help once again @ursooperduper!

PR #77, opened 2015-02-26
read mmol/L CareLink data

jebeck:
This PR contains changes to fully parametrize reading blood glucose(-related) data in CareLink for both the mg/dL and mmol/L cases. (This includes things like insulinSensitivity settings as well as BG values.)

In addition, I added console logging in lib/core/api.js and cleaned up some of the logging for the CLI tools to make debugging with them easier.

Ready for review.


lib/drivers/carelinkDriver.js
```diff
@@ -201,6 +203,52 @@ module.exports = function(simulatorMaker, api){
         var uploads = removeOverlaps(payload);
         progress(20);
 
+        function getUnits() {
+          var units = null;
```
kentquirk: Assuming that the colNames stuff has been called before you call getUnits (and it looks like it has) couldn't you just extract the units from colNames.CalBGForPH without bothering to do this search?

test/carelink/testCommon.js
```diff
@@ -79,4 +80,35 @@ describe('common', function() {
...
+  });
+
+  describe('convertBackToMmol', function() {
+    it('should be a function', function() {
+      expect(common.convertBackToMmol).to.exist;
+      expect(typeof common.convertBackToMmol).to.equal('function');
+    });
+
+    it('should return a floating-point number with one significant digit', function() {
+      expect(String(common.convertBackToMmol(99.08))).to.equal('5.5');
```
kentquirk: would you ever see a non-integral mg/dl value in real data?

kentquirk:
The rest of it looks ok but I don't think you want to merge this until the corresponding blip version has been merged -- otherwise it will block other uploader code from being merged and deployed.

lib/drivers/carelinkDriver.js
jebeck: Yeah, good point. I did the getUnits function before I discovered exactly how many columns were going to have alternating names. I'll change this.

test/carelink/testCommon.js
jebeck: Yup, this is an example inspired by real life. They are clearly converting *to* mg/dL in the `Raw-Values`. For some data types we do still get the actual mmol/L value in a separate column, but for the types that don't have a separate column (in the settings), we only see the conversion.

jebeck:
The blip has been merged and deployed to prod but I agree we want the blip out in the world first. I won't merge this until everything on staging for blip is cleared for prod (this should happen tonight - there's only analytics left to test) so that blip is ready to move to staging for testing.

jebeck:
As soon as Travis reports back that all is well, I am merging this to master, as its dependency (blip v0.5.12) is now on prod.

PR #78, opened 2015-02-26
fix split-extended annotation code

jebeck:
What it says on the tin. Caught during testing: https://trello.com/c/5f2WlFTb


jh-bate:
LGTM

PR #80, opened 2015-03-03
Insulet final touches

jebeck:
@kentquirk please review. This is (mainly) the fix for the bizarre duration on dual-wave boluses that were interrupted before the extended portion was even partially delivered.

### DIFF
--- a/lib/drivers/insuletDriver.js
+++ b/lib/drivers/insuletDriver.js
@@ -206,7 +206,24 @@ module.exports = function (config) {
       'immediate_duration_seconds'
     ], postprocess: function (rec) {
       rec.volume_units = toUnits(rec.volume);
-      rec.extended_duration_msec = rec.extended_duration_minutes * sundial.MIN_TO_MSEC;
+      // this occurs if an extended bolus is programmed but the bolus is interrupted
+      // before any of the extended portion is delivered!
+      if (rec.extended_duration_minutes === 65535) {
+        rec.extended_duration_msec = 0;
+      }
+      else {
+        // a zero *here* means that this isn't an extended bolus
+        // and we need to distinguish between two types of zeros -
+        // dual-wave boluses interrupted before any of the extended was
+        // delivered (i.e., the case above ^) or non-extended boluses (here)
+        // so here we use null instead of 0
+        if (rec.extended_duration_minutes === 0) {
+          rec.extended_duration_msec = null;
+        }
+        else {
+          rec.extended_duration_msec = rec.extended_duration_minutes * sundial.MIN_TO_MSEC;
+        }
+      }
       rec.immediate_duration_msec = rec.immediate_duration_seconds * sundial.SEC_TO_MSEC;
     }
     },
@@ -677,7 +694,7 @@ module.exports = function (config) {
         wiz_idx = bolusrecs[b] + bolus.detail.calculation_record_offset;
       }
       var r = wizRecords[wiz_idx] || {};
-      if (bolus.detail.extended_duration_msec !== 0) {
+      if (bolus.detail.extended_duration_msec !== null) {
         // the extended portion of a dual-wave bolus is split into two records
         // if it crosses local (deviceTime) midnight
         if (r.extended != null) {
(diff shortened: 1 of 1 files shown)


lib/drivers/insuletDriver.js
```diff
@@ -206,7 +206,24 @@ module.exports = function (config) {
       'immediate_duration_seconds'
     ], postprocess: function (rec) {
       rec.volume_units = toUnits(rec.volume);
-      rec.extended_duration_msec = rec.extended_duration_minutes * sundial.MIN_TO_MSEC;
+      // this occurs if an extended bolus is programmed but the bolus is interrupted
+      // before any of the extended portion is delivered!
+      if (rec.extended_duration_minutes === 65535) {
```
jh-bate: so we can guarantee that?

lib/drivers/insuletDriver.js
```diff
@@ -206,7 +206,24 @@ module.exports = function (config) {
...
+        // a zero *here* means that this isn't an extended bolus
+        // and we need to distinguish between two types of zeros -
+        // dual-wave boluses interrupted before any of the extended was
+        // delivered (i.e., the case above ^) or non-extended boluses (here)
+        // so here we use null instead of 0
+        if (rec.extended_duration_minutes === 0) {
+          rec.extended_duration_msec = null;
+        }
+        else {
+          rec.extended_duration_msec = rec.extended_duration_minutes * sundial.MIN_TO_MSEC;
```
jh-bate: ```sundial.MIN_TO_MSEC``` ohhh - I like the variables for magic numbers :)

lib/drivers/insuletDriver.js
```diff
@@ -677,7 +694,7 @@ module.exports = function (config) {
         wiz_idx = bolusrecs[b] + bolus.detail.calculation_record_offset;
       }
       var r = wizRecords[wiz_idx] || {};
-      if (bolus.detail.extended_duration_msec !== 0) {
+      if (bolus.detail.extended_duration_msec !== null) {
```
jh-bate: not worth just testing if its not empty? ```_.isEmpty(...) === false``` ??

lib/drivers/insuletDriver.js
```diff
@@ -873,7 +890,7 @@ module.exports = function (config) {
               .done();
           }
           wizRecords[wiz_idx].handled = true;
-        } else if (bolus.detail.extended_duration_msec !== 0) {
+        } else if (bolus.detail.extended_duration_msec !== null) {
```
jh-bate: as above - ```_.isEmpty(...) === false``` ?

lib/drivers/insuletDriver.js
jebeck: That's a good question. Do you mean guarantee that the *only* situation that produces this is an interrupted bolus? I don't know.

Or do you mean guarantee that an interrupted bolus (with no extended delivered, as specified) produces the special value 65535? Yes, I believe we can be confident in that. Quoting from @kentquirk "Look at the number -- it's 65535 -- that's 0xFFFF -- or -1 if you treat it as a 16-bit signed number." It's clearly a special value, not just a weird constant.

lib/drivers/insuletDriver.js
jebeck: This a debate we're going to have till the end of time! I like `_.isEmpty` for objects and arrays. I dislike it intensely for single values because I can never remember what its truth table actually is.

Also in this case we definitely *can't*, as the point is to distinguish between `null` and 0:
```
> _.isEmpty(0)
true
> _.isEmpty(null)
true
```

lib/drivers/insuletDriver.js
jh-bate: thanks for that

lib/drivers/insuletDriver.js
jh-bate: yeah but ... :)

good enough for me

jh-bate:
LGTM

kentquirk:
I know it's already merged...but LGTM too

PR #81, opened 2015-03-03
Device selection tweaks

jebeck:
Some styling enhancements for the new UI with device selection.


jh-bate:
LGTM code wise.

Do the ellipses hide anything we need to see for the long OmniPod name?? Or are we just doing the file pick on the only file we can pick for OmniPod?

jebeck:
People will only see one file available, yes. And we give an error message if you upload something with a non .ibf extension.

Also, the full filename is visible in a browser-default tooltip when you hover - the browser provides that for free.

PR #82, opened 2015-03-05
Jebeck/asante sim

jebeck:
Since it looks like my highest priority moving forward for the near-term future (i.e., next several days to a week) is going to be CareLink, I think this work (to make Asante demo-able) should be reviewed and merged to avoid long-running branches.

Ping @kentquirk 

This PR contains some general house-keeping to make the driver match the other drivers, builder fixes for some values that were in incorrect units, building of manual-entry SMBGs from wizard records, and a minimal simulator to get the basals (minus temp basals and suspends) going.


lib/drivers/asanteDriver.js
```diff
@@ -727,41 +690,42 @@ module.exports = function (config) {
...
           // process record
           retval.push(result.datarecord);
-          // console.log(result);
           // request next record
           var next = nextRecord();
-          if (retval.length >= 3000) {    // 3000 is bigger than any log's capacity
+          if (retval.length >= 3000) { // 3000 is bigger than any log's capacity
             next = stopSending();
-            console.log('cutting it short for debugging!');
+            // TODO: ask KJQ whether the cutting it short applies only to values
```
kentquirk: When I'm debugging, I typically change the number 3000 to 30, and I use that log message to remind me not to check it in.

lib/drivers/asanteDriver.js
```diff
@@ -1204,8 +1179,48 @@ module.exports = function (config) {
...
+          case 'settings':
+            simulator.settings(datum);
+            break;
+          case 'smbg':
+            simulator.smbg(datum);
+            break;
+          case 'wizard':
+            simulator.wizard(datum);
+            break;
+          default:
```
kentquirk: Given that all of these are the same signature, I think I would have done something like: 

````
var simfuncs = {
  basal: simulator.basal,
  bolus: simulator.bolus,
 ...
}; 

for (var i = 0; i < postrecords.length; ++i) {
        var datum = postrecords[i];
        if (!_.has(simfuncs, datum.type)) {
          // fail
        } else {
           simfuncs[datum.type](datum)
    }
}
````

In fact, looking at that now, I think the dispatcher belongs in the simulator.

kentquirk:
I made a couple of comments, but they're not critical. You may merge when ready.

lib/drivers/asanteDriver.js
jebeck: Yeah, we talked about this, but then I got confused when I actually read the comment, so I added the note. Maybe I'll adjust the comments a bit for clarity.

lib/drivers/asanteDriver.js
jebeck: Hmm, I can see your point. The signatures will not always be the same though. They only are for the simplest types. (Cf. [here](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/drivers/insuletDriver.js#L1530) in the Insulet driver; I expect similar, although not exactly the same, for Asante.)

I think moving the dispatcher sounds like a good idea, but I'd prefer to wait and do it for both Insulet and Asante at the same time (for parallelism).

jebeck:
Merging this. Note that it will be accessible from the UI on master - commit 4c31ead60523b958ef60df1371d49bf85eb11a10 will need to be reverted for the next release. (Probably I should just stash and unstash that change while I'm working in future.)

PR #83, opened 2015-03-08
Insulet settings bugfix plus addition of sequencing info

jebeck:
Fixes bug in storage of `bgTarget` settings. Also adds sequencing index info for Jaeb.

### DIFF
--- a/lib/drivers/insuletDriver.js
+++ b/lib/drivers/insuletDriver.js
@@ -883,10 +883,19 @@ module.exports = function (config) {
               )
               .with_duration(
                 data.log_records[ext].detail.extended_duration_msec + data.log_records[ext2].detail.extended_duration_msec
-              ).done();
+              )
+              // TODO: delete after conclusion of Jaeb study
+              .set('jaebPayload', {logIndices: [
+                data.log_records[imm].log_index,
+                data.log_records[ext].log_index,
+                data.log_records[ext2].log_index
+              ]})
+              .done();
           } else {
             postbolus = postbolus.with_extended(data.log_records[ext].detail.volume_units)
               .with_duration(data.log_records[ext].detail.extended_duration_msec)
+              // TODO: delete after conclusion of Jaeb study
+              .set('jaebPayload', {logIndices: [data.log_records[imm].log_index, data.log_records[ext].log_index]})
               .done();
           }
           wizRecords[wiz_idx].handled = true;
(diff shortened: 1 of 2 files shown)


lib/objectBuilder.js
```diff
@@ -96,7 +98,21 @@ module.exports = function () {
           throw new Error(util.format('Some arguments to %s(%j) were not specified!', this.type, valid.join(',')));
         }
 
-        return _.pick(this, function (value, key) {
+        // TODO: delete after conclusion of Jaeb study
```
kentquirk: a) Please mark the range of lines that should be deleted
b) However, I really don't like this code being in the general-purpose objectBuilder. You've got two apparently mutually-exclusive objects that are being combined into one payload and also allowing the payload itself to have a value. Couldn't/shouldn't this logic just live in the caller function?

lib/drivers/insuletDriver.js
```diff
@@ -883,10 +883,19 @@ module.exports = function (config) {
               )
               .with_duration(
                 data.log_records[ext].detail.extended_duration_msec + data.log_records[ext2].detail.extended_duration_msec
-              ).done();
+              )
+              // TODO: delete after conclusion of Jaeb study
```
kentquirk: Please mark the range of lines that should be deleted

lib/objectBuilder.js
jebeck: I don't like this code either, but sticking this in the object builder was by far the easiest way to get it in with minimal code changes elsewhere. Mainly the problem is that `.with_payload()` (or `.set('payload', obj)`) don't perform a merge with the current payload, and there are lots of places in both CareLink and Insulet code where we already were using the payload, so we couldn't just replace it with the "Jaeb payload". Doing a merge at the very end during `done` seemed like the easiest option.

The `jaebPayload` option is for CareLink (as you'll see when you look at that PR) and for the objects in Insulet that don't map 1:1...I was already using `.set('index'...)` in the Insulet code and so it was easier to just add an `else if` to merge those into the payload rather than go and change all our code that depends on the log index, which we will continue to need (for intermediate processing; it wasn't getting stored) even after the Jaeb study concludes, unlike the other stuff.

Happy to chat about this and try to come up with something more elegant...

I am updating with marks for end of deletion everywhere; good idea.

kentquirk:
LGTM

PR #84, opened 2015-03-08
add sequencing info for Jaeb

jebeck:
This is a start on reading time change events in CareLink (i.e., it adds parsing and simulation of deviceMeta subType `timeChange` events, but does not do any "bootstrapping" from them).

It also preserves the `Raw-Seq Num` and `Raw-Upload ID` in the payload for most objects, as requested by Jaeb.


kentquirk:
LGTM

PR #85, opened 2015-03-09
changes to add the UTC time for errors

jh-bate:
also a test to assert it remains so


jebeck:
This looks goods - I adjusted the format a bit and updated from master since there were a lot of PRs merged in between. I'll merge as soon as Travis reports back.

PR #86, opened 2015-03-09
add internalTime to the payload for Dexcom EGVs and calibrations

jebeck:
Add the Dexcom internal time to the payload. Cleaner than for the pumps, since there's no much less object-building going on.

I also made sure we're handling the UTC time translation into a string - happened to catch that that was happening in the transformation to JSON on the way to the server, which I think only by chance matched what we wanted!


kentquirk:
LGTM

PR #87, opened 2015-03-09
Jhbate/upload meta check

jh-bate:
WIP but should be done, discussing with @jebeck

### DIFF
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -267,12 +267,14 @@ api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb) {
         .with_payload(sessionInfo.payload)
         .done();
 
+      api.log('saving upload metadata');
+
       post_upload_meta(uploadItem, function(err, count){
         if(err){
-          api.log('error saving upload metadata so bailing ', err);
+          api.log('error saving upload metadata ', err);
           callback(err);
         }
-        api.log('upload metadata saved now moving onto processing the  ');
+        api.log('saved upload metadata ', count);
         callback(null, uploadItem);
       });
     },
@@ -283,6 +285,7 @@ api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb) {
       for (var i = 0; i < data.length; i += BLOCKSIZE) {
         blocks.push(data.slice(i, i + BLOCKSIZE));
       }
+      api.log('start uploading the rest of the data');
       //process then call cb
       async.mapSeries(blocks, post_and_progress, cb);
     }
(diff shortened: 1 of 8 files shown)


lib/core/api.js
```diff
@@ -217,23 +212,13 @@ function postBlockToPlatform(data, groupId, callback) {
...
-    api.log('postBlockToPlatform: as logged in user');
-    tidepool.uploadDeviceDataForUser(data, function(err, result) {
-      if (err) {
-        return sad(err, err);
-      }
-      return happy(result);
-    });
-  }
+
+  tidepool.uploadDeviceDataForUser(data, groupId, function(err, result) {
```
jebeck: Do you want to put an api.log back here? Maybe including logging of the groupId?

jebeck:
This did break the CLIs @jh-bate 

Can you fix please?

jh-bate:
sorry @jebeck - after saying I would test that I totally forgot about it.

jh-bate:
So both cli's now pass in the id of the user into the driver.

jebeck:
One of the log statements ends up with `undefined` both in CLIs and through UI: Api              | saved upload metadata  undefined

Trying to log the count?

jh-bate:
yeah it was but  we just care if there is an error so updated that

jebeck:
So this mostly works as I'd expect now. There's still an undefined being logged - maybe it's just a common issue for all the drivers? It occurs [here](https://github.com/tidepool-org/chrome-uploader/blob/jhbate/upload-meta-check/lib/drivers/carelinkDriver.js#L314) and [here](https://github.com/tidepool-org/chrome-uploader/blob/jhbate/upload-meta-check/lib/drivers/insuletDriver.js#L1632), for instance...

jebeck:
Oh and also the UTC time isn't being added to these errors...is that expected?

I think it'd be great if we could customize the error message for these too. Is that possible? "Upload failed due to bad metadata" or something?

jh-bate:
Yeah I think there are a few places like that in the uploader as there was a lot of logging done in order to see what was going on.

With regards to the other Q. Yes, I was going to tackle that as part of the card https://trello.com/c/FccsTSO7. If your happy with that answer then shall we merge this and tackle the nicer consistent messages for the next card

jebeck:
Leaving nicer messages for the next task is fine, but maybe remove the attempts at logging `result` in the drivers? I'd rather that not get left behind - I find it confusing (and potentially red herring) when there are undefined things being logged.

jh-bate:
cool - log statement removed from all drivers when there is an error

PR #88, opened 2015-03-13
remove invalid timestamps early in CareLink parsing

jebeck:
This removes any data with an invalid timestamp given the selected (or at present, default US/Pacific) timezone on upload - e.g., any `deviceTime`s between 2 and 3 a.m. on March 8th, 2015 (because the US/Pacific clock turns over straight from 1:59:59 to 3:00:00 for Spring Forward).

I would like to discuss with @HowardLook and @brandonarbiter ways of recording and/or exposing the data deletion happening here; possibly adding info to the upload metadata object about data that could not be processed would be appropriate.

Ping @jh-bate for review. Using the CLI tool, it's good to try this with a timezone that undergoes DST (as US/Pacific does) and with one that doesn't (e.g., Pacific/Honolulu) to observe the difference.

### DIFF
--- a/lib/carelink/basal.js
+++ b/lib/carelink/basal.js
@@ -18,11 +18,11 @@
 var common = require('./common.js');
 var parsing = require('./parsing.js');
 
-module.exports = function (timezone) {
+module.exports = function () {
   var parser = common.makeParser(
     {
       BasalProfileStart: [
-        common.makeCommonVals(timezone),
+        common.makeCommonVals(),
         {
           uploadSeqNum: parsing.asNumber('Raw-Seq Num'),
           deliveryType: 'scheduled',
@@ -31,7 +31,7 @@ module.exports = function (timezone) {
         }
       ],
       ChangeTempBasalPercent: [
-        common.makeCommonVals(timezone),
+        common.makeCommonVals(),
         {
           deliveryType: 'temp',
           percent: parsing.map(['Raw-Values', 'PERCENT_OF_RATE'], function(e){ return parseFloat(e) / 100.0; }),
@@ -39,7 +39,7 @@ module.exports = function (timezone) {
         }
       ],
       ChangeTempBasal: [
-        common.makeCommonVals(timezone),
+        common.makeCommonVals(),
         {
           deliveryType: 'temp',
           rate: parsing.asNumber(['Raw-Values', 'RATE']),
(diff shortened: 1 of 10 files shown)


kentquirk:
But isn't this exactly the wrong solution, given that the user almost certainly didn't change the device time at exactly 1:59:59? A few hours or days later, they'll jump time forward when (if) they reset the clock.

Proper handling of device time changes will make this point moot. Why would we add code to delete events that are not actually wrong?

jebeck:
It currently crashes CareLink uploads @kentquirk because the overlap results in out-of-sequence data.

This code will be removed once we "bootstrap" properly to UTC, but right now we have users who can't upload CareLink data.

jh-bate:
So the code looks fine to me, luckily you had the comment about not having the func in sundial :)

my questions comments are
- so would there be any potential impact on jeab?  
- rock and a hard pace I know in general it doesn't feel right deleting data but rather flag the records and then filter on display. It sounds like it might be some of this that you will be discussing with @HowardLook and @brandonarbiter though

jebeck:
Actually we wouldn't even probably need to filter on display; it's just that because of how the simulator works it's really hard to keep the simulator from crashing on the overlap. @HowardLook and I already discussed alternative strategies vs. deletion (and decided in favor of deletion since we don't expect this "fix" to last very long) although I wouldn't mind revisiting that discussion. The most obvious alternative would be to separate all the invalid timestamp data and treat it as a separate "device" (exactly the same strategy we use for when a user has multiple pumps operating simultaneously) - then process that "device" separately. If I get a chance to tinker with that strategy over the weekend, maybe I will and report back about how it goes.

I don't think this should impact Jaeb because it only happens for Spring Forward, and I assume they aren't looking at any retrospective data after the study starts.

jebeck:
So I just finished experimenting with the "non-deletion" version, but it just puts us in the same situation as Dexcom - we still end up deleting around half of the "overlapping" data, just a different half. (This is because the timestamps for CGM readings are identical hour over hour, except for the hour part - e.g. display time 2015-03-08T02:01:00 and 2015-03-08T03:01:00 produce cbg objects with identical IDs because the translated-into-UTC timestamps are identical due to moment's decision about how to deal with the invalid timestamps, and jellyfish rejects the upload of the second item with the same ID.)

This only really applies to CGM data (because of how the timestamps are identical hour-over-hour). The "non-deletion" version may be slightly preferable because of the fact that we preserve other data types that could have been deleted on my first version - e.g., boluses.

@HowardLook and @brandonarbiter I suggest we talk over options.

jebeck:
I also put the "non-deletion" version over [in another branch](https://github.com/tidepool-org/chrome-uploader/tree/jebeck/cl-invalid-timestamps-separate-upload) in case we decide to use it.

jebeck:
Closing this, as we are going with the strategy in #91.

PR #89, opened 2015-03-13
Jhbate/nice errors

jh-bate:
WIP @jebeck @kentquirk feel free to have a look.

really just need to get to to display the debug details also (we already log it) so we can get the more detailed user errors we want. 

* Create a common error type that all other errors are wrapped in
* Move app error messages into common location
* Build a "debug" message within the common error that contains utc time, error code, current step
  * so we can get better detail from errors created during parsing data (for instance) you can add a `step` to the error that will then be attached to the error in the appActions  
* Retain original error message so we don't loose any details
* Ensure wrapped error is also logged to server for consistency

I have asked the question about how we want to display these errors so its not overwhelming. I was thinking the message was the friendly part and then there was the debug with all the details we might want if the use sends in a screen shot. I was thinking of a modal but open to other suggestions.

### DIFF
--- a/lib/components/Upload.jsx
+++ b/lib/components/Upload.jsx
@@ -204,16 +204,7 @@ var Upload = React.createClass({
     }
     if (this.isUploadFailed()) {
       var uploadError = this.getUploadError();
-
-      if (getIn(uploadError, ['error', 'code']) && getIn(uploadError, ['error', 'message'])) {
-          return <div className="Upload-status Upload-status--error">{uploadError.error.message}</div>;
-      }
-
-      if (uploadError.code && uploadError.code === 404) {
-        return <div className="Upload-status Upload-status--error">{uploadError.message}</div>;
-      }
-
-      return <div className="Upload-status Upload-status--error">{'The upload didn\'t work.'}</div>;
+      return <div className="Upload-status Upload-status--error">{uploadError.message}</div>;
     }
     if (this.isBlockModeFileChosen()) {
       return <div className="Upload-status Upload-status--uploading"><p>{this.props.upload.file.name}</p></div>;
--- a/lib/state/appActions.js
+++ b/lib/state/appActions.js
@@ -42,6 +42,15 @@ appActions.trackedState = {
   SEE_IN_BLIP : 'Clicked See Data in Blip'
 };
 
+appActions.errorText = {
+  E_READING_FILE : 'Error reading file: ',
+  E_WRONG_FILE_EXT : 'Please choose a file ending in: ',
+  E_UPLOAD_IN_PROGRESS : 'Cannot start upload while an upload is in progress',
+  E_UNSUPPORTED_TYPE : 'Unsupported upload source type: ',
+  E_INVAILD_UPLOAD_INDEX : 'Invalid upload index: ',
+  E_DEVICE_NOT_CONNECTED : 'The device doesn\'t appear to be connected'
+};
+
 appActions.bindApp = function(app) {
   this.app = app;
   return this;
@@ -380,7 +389,7 @@ appActions.readFile = function(uploadIndex, targetId, file, extension) {
     var reader = new FileReader();
 
     reader.onerror = function(e) {
-      var error = new Error('Error reading ' + file.name);
+      var error = new Error(self.errorText.E_READING_FILE + file.name);
       inputError(error);
     };
 
@@ -401,7 +410,7 @@ appActions.readFile = function(uploadIndex, targetId, file, extension) {
     return true;
   }
   else {
-    var error = new Error('Please choose a file ending in ' + extension);
+    var error = new Error(self.errorText.E_WRONG_FILE_EXT + extension);
     error.code = 404;
     inputError(error);
     // the return is just for ease of testing
@@ -416,7 +425,7 @@ appActions.upload = function(uploadIndex, options, cb) {
   this._assertValidUploadIndex(uploadIndex);
 
   if (appState.hasUploadInProgress()) {
-    throw new Error('Cannot start upload while an upload is in progress');
+    throw new Error(self.errorText.E_UPLOAD_IN_PROGRESS);
   }
 
   var upload = this.app.state.uploads[uploadIndex];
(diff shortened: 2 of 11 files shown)


lib/state/appActions.js
```diff
@@ -42,6 +43,29 @@ appActions.trackedState = {
...
+  E_UNSUPPORTED_TYPE : 'Unsupported upload source type: ',
+  E_INVAILD_UPLOAD_INDEX : 'Invalid upload index: ',
+  E_DEVICE_NOT_CONNECTED : 'The device doesn\'t appear to be connected'
+};
+
+appActions.errorStage = {
+  STAGE_SETUP : { Code: 'E_SETUP' , Message: 'Error during setup', Step: '' },
+  STAGE_LOGIN : { Code: 'E_LOGIN' , Message: 'Error during login', Step: '' },
+  STAGE_LOGOUT : { Code: 'E_LOGOUT' , Message: 'Error during logout', Step: '' },
+  STAGE_DETECT_DEVICES : { Code: 'E_DETECTING_DEVICES' , Message: 'Error detecting devices', Step: '' },
```
jebeck: What's the difference between this `STAGE_DETECT_DEVICES` and the lower one `STAGE_DEVICE_DETECT`?

Maybe the first is because of some of the "legacy" code that's stuck around from back when we were trying to auto-detect plugged in devices (instead of having the device selection set-up menu)?

lib/components/Upload.jsx
```diff
@@ -204,16 +204,7 @@ var Upload = React.createClass({
...
-      if (getIn(uploadError, ['error', 'code']) && getIn(uploadError, ['error', 'message'])) {
-          return <div className="Upload-status Upload-status--error">{uploadError.error.message}</div>;
-      }
-
-      if (uploadError.code && uploadError.code === 404) {
-        return <div className="Upload-status Upload-status--error">{uploadError.message}</div>;
-      }
-
-      return <div className="Upload-status Upload-status--error">{'The upload didn\'t work.'}</div>;
+      return <div className="Upload-status Upload-status--error">{uploadError.message}</div>;
```
jebeck: Ahh, much cleaner :)

test/core/testUploaderError.js
```diff
@@ -0,0 +1,56 @@
...
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+/* global beforeEach, describe, it */
+
+var expect = require('salinity').expect;
+
+var UploaderError = require('../../lib/core/uploaderError.js');
```
jebeck: Thanks for adding these tests. I was gonna ask for them, but they're already here! \o/ Will look in more detail later, but so far all of this looks pretty good to me.

lib/components/Upload.jsx
jebeck: But the `message` doesn't have the UTC appended anymore. You're waiting for design input on how we want to display the additional info like UTC and stage, right?

lib/state/appActions.js
jh-bate: there is a part of the code that is about detecting multiple devices and then part for a specific device. As you say it may now be redundant - I will have a look

lib/components/Upload.jsx
jh-bate: yeah - and the top level I have split it into the `message` and also `debug`. The debug info contains all the extras such as the UTC datetime. Yeah so I want get design input so we dislapy all the info but it is split into the friendly `message` and the more detailed and helpful to us `debug` details

lib/state/appActions.js
jebeck: I'm pretty sure none of that "proactive" device detection code is in use. I'd be in favor of a cleanup to reduce developer confusion (and we can always go back into the git history to recover it), but I don't know if re-introducing proactive device detection is near enough on our roadmap that we should just keep it?

kentquirk: Device detection is a significant rathole, it turns out. Trying to identify a device safely and quickly when they don't share a protocol is a problem that goes up in complexity with the square of the number of devices you support. So yes, we can reasonably remove the code that tries to do it.

(Although the old saw "you can always go back in git history to recover it" is a lot harder than it looks. Reality is that if any significant amount of time has passed since you deleted it, it's rarely worth going back to try to reuse old code -- too much of the landscape will have changed.)

jh-bate: cool - thats not the objective of this PR though so we can do that separately.

jebeck: True @jh-bate, not the objective of this PR.

And yes @kentquirk I guess I meant more that we could go back and find it to look at for reference, not use it directly. I did that when reimplementing the file selection and reading for OmniPod files, but I will admit it took me a fairly long hunt one evening on GitHub to go back far enough in the history to find where that file selection had last existed...

lib/state/appActions.js
```diff
@@ -42,6 +43,29 @@ appActions.trackedState = {
   SEE_IN_BLIP : 'Clicked See Data in Blip'
 };
 
+appActions.errorText = {
+  E_READING_FILE : 'Error reading file: ',
+  E_WRONG_FILE_EXT : 'Please choose a file ending in: ',
+  E_UPLOAD_IN_PROGRESS : 'Cannot start upload while an upload is in progress',
+  E_UNSUPPORTED_TYPE : 'Unsupported upload source type: ',
+  E_INVAILD_UPLOAD_INDEX : 'Invalid upload index: ',
```
kentquirk: INVALID, please

jebeck:
As we discussed yesterday @jh-bate I tried this with some of our currently failing uploads. Errors during data processing are *not* being caught by this. Is there a way we can catch them without having to go through all the drivers and simulators and replacing all `Error` instances with `UploaderError`?

kentquirk:
I've added a commit to allow setting port patterns by device and OS, and extended the Dexcom port pattern.

kentquirk:
I've reviewed @jh-bate 's stuff again and it LGTM. Since I've added code, I'm not going to merge. @jebeck if you could give my fix a quick once-over and merge, we can push it to the store.

PR #90, opened 2015-03-15
check the account isn't for someone else

jh-bate:
simple fix for displaying the correct name for who we are uploading the data for. see https://trello.com/c/8lxY82lW

@jebeck @kentquirk


jebeck:
LGTM. Merge at will.

PR #91, opened 2015-03-19
invalid local timestamps as a separate upload

jebeck:
OK, so now I think this is the better solution than #88. We add a millisecond (by simple string substitution - replacing .000Z with .001Z - because we never have millisecond info from CareLink timestamps) to all the timestamps that are invalid in the chosen timezone to prevent duplicates, then upload all these as a separate "device" to prevent the simulator from crashing on overlapping data. This does mean there will be an additional upload metadata object for users whose data has this problem, but I don't actually think that's a bad thing.

This involved adding the ability to set a default annotation (the text for which will not appear in blip until a tideline change goes through), which is something we may or may not want to keep in the long term.

CC @kentquirk

### DIFF
--- a/lib/carelink/basal.js
+++ b/lib/carelink/basal.js
@@ -18,11 +18,11 @@
 var common = require('./common.js');
 var parsing = require('./parsing.js');
 
-module.exports = function (timezone) {
+module.exports = function () {
   var parser = common.makeParser(
     {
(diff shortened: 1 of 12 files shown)


kentquirk:
Ok; very similar to #92 so I'm happy.

PR #92, opened 2015-03-20
fix issues around (U.S.) DST changeover for Dexcom

jebeck:
This PR depends on [jellyfish 58](https://github.com/tidepool-org/jellyfish/pull/58). It allows us to recover data that wasn't uploaded due to invalid timestamps (given the hardcoded timezone) and uses the same "add a millisecond" strategy as #91 to differentiate "identical" UTC timestamps.

It depends on the jellyfish PR because sometimes the invalid member of the pair was uploaded (without the millisecond addition), and so we needed to force storage of a new version of all the datapoints within the one-hour UTC that produced a data overlap.

I've tested this on my own receiver that suffered from this problem, and it fixes the data exactly as expected.

### DIFF
--- a/lib/carelink/common.js
+++ b/lib/carelink/common.js
@@ -131,6 +131,22 @@ exports.convertBackToMmol = function(n) {
   return Math.floor(inMmol * 10 + 0.5) / 10;
 };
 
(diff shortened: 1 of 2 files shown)


kentquirk:
LGTM

PR #93, opened 2015-03-24
fudge a millisecond only on spring forward overlap datapoints that would have been dropped before

jebeck:
So I think this actually accomplishes the original goal of completely fixing the data. I really need to test on B's receiver again to be sure, but I think I'm now always only fudging a millisecond on the datapoint that was considered a duplicate before (and thus not stored). (I think I can guarantee this because I found the code where @kentquirk was logging duplicates, and I'm leveraging the order the "pairs" come in, which is deterministic across multiple uploads from the same receiver, but *not* deterministic in terms of which of the pair comes "first.")

### DIFF
--- a/lib/drivers/dexcomDriver.js
+++ b/lib/drivers/dexcomDriver.js
@@ -20,6 +20,7 @@ var async = require('async');
 var sundial = require('sundial');
 var annotate = require('../eventAnnotations');
 var common = require('../carelink/common');
+var util = require('util');
 
 var crcCalculator = require('../crc.js');
 var struct = require('../struct.js')();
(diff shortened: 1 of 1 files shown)


jebeck:
Checked on two more receivers, and it works. Ready for review @kentquirk

lib/drivers/dexcomDriver.js
```diff
@@ -631,6 +628,9 @@ module.exports = function (config) {
           threshold: 400
         };
       }
+      if (!common.isValidLocalTimestamp(datum.displayTime, datum.displayUtc, sundial.getOffsetFromZone(datum.displayUtc, cfg.timezone))) {
+        datum.annotation = {code: 'spring-forward'};
```
jh-bate: cool - so any data that this change touch's will be annotated with this?

lib/drivers/dexcomDriver.js
```diff
@@ -645,6 +645,7 @@ module.exports = function (config) {
       }
       if (datum.annotation) {
         annotate.annotateEvent(cbg, datum.annotation);
+        cbg._forceUpdate = true;
```
jh-bate: and this is the flag we add to say this must be updated / or we can force an update?

jh-bate:
I was going to say are there corresponding tests but there is no test suite for the dexcomDriver. If you could just clarify my questions above other than that code wise LGTM

lib/drivers/dexcomDriver.js
jebeck: Only the data that has the invalid local timestamp will be annotated (that will also help us identify it as opposed to the other half of the overlap). Because of the Dexcom sort order, the data points in the non-invalid-timestamp half of the data are actually the ones being edited :/ I thought about introducing an annotation there to reflect the fact that we're adding a millisecond to the timestamp, but then I thought that info is already recoverable because the `deviceTime` is there without the millisecond addition. Open to changing this if others feel strongly about having an annotation whenever we add the millisecond (I think it should be a different one though, so that `spring-forward` only identifies data that originally had timestamps that are invalid in the local timezone).

lib/drivers/dexcomDriver.js
jebeck: Yep, see https://github.com/tidepool-org/jellyfish/pull/58, which is already on staging.

PR #94, opened 2015-03-24
Kjq/update deps

kentquirk:
@jh-bate @jebeck I think this is ready.


jebeck:
Confirmed that tests pass for me locally in both node 0.10.37 and node 0.12.0.

@jh-bate were talking about this a bit yesterday because he had started a similar branch. Is there a reason you're leaving out the lodash update from 2.4.1 to 3.5.0? It does have a breaking change that affects us, but I think it's just one thing - `_.has` has been renamed to `_.includes`.

kentquirk:
Yes, I saw that lodash had breaking changes and backed them out; hadn't gone back yet to figure out what they were. I can take a look at that.

jebeck:
Also React is now on 0.13. There were far more changes that went into 0.12 (nothing that particularly affected us, I think, although I probably should have been looking more closely at the console when I checked blip yesterday, we may have a couple of deprecation warnings in there - there were quite a few changes in 12 that affected a lot of external React libraries). Do you want to stay on 0.12 since that's what blip is on?

kentquirk:
I saw that react just went to 0.13 and I felt like it could be an infinite treadmill. I don't object to going to it but I thought that 0.12 was still current enough. Let's do it as a separate thing rather than as part of this.

jebeck:
I tagged and published a new sundial - v1.1.8. It'd be nice to include here, as it removes the console warnings about deprecated methods in the higher version of moment(-timezone).

jh-bate:
I will do the update of sundial

jh-bate:
updates LGTM. 
note: lodash is on an older version but I think thats ok for now - I would rather not hold this up

PR #95, opened 2015-03-26
Insulet bugfixes #1 (wizard-building, basal schedule names)

jebeck:
Fixes two bugs in Insulet data processing:
- at the beginning of the PDM's data records, a bolus record may refer to a wizard record that does not exist in the PDM's memory (presumably out of space); in such cases we stop trying to find and build such a non-existent wizard record
- `.` and `$` are illegal characters in object keys for mongo's serialized JSON (BSON), but they can occur in the basal schedule names for Insulet, so we replace the former with `-` and the latter with an empty string (actually I don't know if you can add a `$` via the Insulet PDM interface but both those characters are illegal, so it seemed prudent to deal with both)

One user was experiencing both of these issues, hence covering both in the same PR, because I was trying to get that data to upload.


jebeck:
Updated this as discussed @kentquirk 

The new logic (checking for empty `wiz.detail`) covers both cases of weird wizard record retrieval I've seen in real data.

lib/insulet/insuletSimulator.js
```diff
@@ -346,6 +346,7 @@ exports.make = function(config){
       // (although within their own type all the basals are always in order)
       // end result: we have to sort events again before we try to upload them
       var orderedEvents = _.sortBy(_.filter(events, function(event) {
+        delete event.index;
```
jh-bate: do we need a comment as to why we are deleting this?

lib/drivers/insuletDriver.js
```diff
@@ -508,6 +508,8 @@ module.exports = function (config) {
       var prog = struct.unpack(basalProgramsHeader.rawdata, prgOffset,
                                fixedRecords.basal_programs_name.format,
                                fixedRecords.basal_programs_name.fields);
+      prog.name = prog.name.replace(/\./g, '-');
```
jh-bate: I guess the thing that worries me with the drivers is that lack of tests around them. In this instance maybe we could have a func that is used but can also be independently tested without taking the rest of the drive into consideration. We know that if the name comes in as this format we need to update it to be that format otherwise bad things can happen.

Thoughts?

lib/insulet/insuletSimulator.js
jebeck: Will add a comment, although this is going away very soon. Once we have UTC bootstrapping, I won't need the log index to get the right order of pump events.

lib/drivers/insuletDriver.js
jebeck: In the main, I agree, but in this particular case I'm not sure what tests are going to gain us. The real thing that would need testing is making sure we're not building *any* objects that have keys that use the illegal characters. Separating out the regex replace functions into a testable module is only going to allow use to test that we can remove illegal characters when we want do; it won't do anything to ensure that we're removing them in all the necessary places.

We should be adding to the jellyfish validation to look for these illegal characters in keys, and at that point we can have ingestion tests. Does that sound like an OK long-term solution?

lib/drivers/insuletDriver.js
```diff
@@ -1512,7 +1520,15 @@ module.exports = function (config) {
...
       // first sort by log index
       postrecords = _.sortBy(postrecords, function(d) {
-        return d.index;
+        if (d.index) {
+          return d.index;
+        }
+        else if (d.payload && d.payload.logIndices) {
+          return d.payload.logIndices[0];
+        }
+        else {
```
kentquirk: I can't find anywhere that indicates that returning undefined from a sort comparator has well-defined behavior. What are you intending this to do?

lib/insulet/insuletSimulator.js
```diff
@@ -346,6 +346,10 @@ exports.make = function(config){
       // (although within their own type all the basals are always in order)
       // end result: we have to sort events again before we try to upload them
       var orderedEvents = _.sortBy(_.filter(events, function(event) {
```
kentquirk: I believe it's technically OK for filter to modify the elements it touches but it does feel a bit wrong to me. I think I'd at least make note of it in a comment.

Also, because you've nested sort and filter, it's hard to tell what's going on. Any chance of pulling filter out into a named local function?

lib/drivers/insuletDriver.js
jebeck: Ah damn, let me look at this some more. It appears the `undefined`s are being sorted to the end by lodash, but I don't want that - these are non-log records (certain alarms). Let me play around a bit.

lib/insulet/insuletSimulator.js
jebeck: Sure, will do.

lib/drivers/insuletDriver.js
jebeck: OK, because I shifted the deletion of `index` later, I actually didn't need to change the behavior of this sort. Going back to just sorting by `d.index` (about to push the reversion).

jebeck:
OK, I think I addressed your concerns @kentquirk - thanks for the catch on the sortBy...

kentquirk:
LGTM now.

PR #96, opened 2015-03-26
Jhbate/errors ui

jh-bate:
@jebeck ready for an initial look. Still need to get the styling sorted out

- updates to the way an UploadError is initialised
- any error(s) from uploads shown above the upload items

TODO:
- styling of individual error messages
- styling on background of failed upload item


lib/components/UploadList.jsx
```diff
@@ -30,6 +30,27 @@ var UploadList = React.createClass({
...
+    });
+
+    if(_.isEmpty(uploadsWithError)) {
+      return null;
+    }
+
+    return _.map(uploadsWithError, function(upload){
+      return (
+        <div key={upload.key} className="UploadList-error-item">
+          <div className="UploadList-error-name">{upload.name}</div>
```
jebeck: Maybe this section should have a headline at the top. "Upload Errors" or something? Either that or maybe make this "Upload error in " + upload.name

lib/core/uploaderError.js
```diff
@@ -17,29 +17,43 @@
...
 
-  this.debug = [
-    message || stage.Message,'|',
+  return [
+    message || otherError.message ,'|',
     UploaderError.UTC_TIME, sundial.utcDateString(),'|',
     UploaderError.CODE, stage.Code,'|',
-    UploaderError.STEP, theStep
+    UploaderError.CODE_MSG, stage.Message,'|',
+    theStep
```
jebeck: I haven't been able to produce one yet that actually has a `theStep`...? (There's always nothing after the last pipe.)

lib/components/UploadList.jsx
jh-bate: yeah - thats one of the things along with colours and layout I wanted to tackle in a UI review with @brandonarbiter

lib/core/uploaderError.js
```diff
@@ -17,29 +17,43 @@
 
 var sundial = require('sundial');
 
-// Create a new object, that prototypally inherits from the Error constructor.
-function UploaderError(message, stage, error) {
-
-  this.name = this.constructor.name;
-  this.message = message;
+// if another error is passed in then attach it
```
jebeck: Ah, I think this is the cause of some of the redundancy I complained about in the Trello card. Want to elaborate on why you think this is useful?

lib/core/uploaderError.js
jh-bate: So you will see it used in the tests but it might be a redundant feature? The intention was if you wanted to attach step information from a driver where you either catch or throw an error that would be meaningful in diagnosing what went wrong then this is a way to do that.

lib/core/uploaderError.js
jh-bate: I don't want to loose any history that may be attached to the original message we caught - this is a simple way to do that.

jebeck: I worry that this has the potential to be confusing. I really don't like the way the redundancy shows up in the server logs (nor the formatting? is there anything to be done about the insane proliferation of `\`?? They multiply like rabbits when there's more than one level of error)

jh-bate: So one of our goals is to have good info so we can track down and solve issues. I would rather have more than less at this stage but if over time it is always redundant then we can axe it or if there is a better way to expose it then we can adapt at that time. 

I can separately look at removing the \ as I agree that makes it hard to read.

jebeck:
Starting to take a look at this. I just added a commit to deal with incorrect CareLink creds better.

Now I was checking out what the result would look like for the recent CareLink parsing bug. It's...not helpful at all:
screenshot 2015-04-27 15 29 17

jh-bate:
Agreed - I can do those a specific type

jebeck:
And a weird regression - Last upload being reported in the button??
screenshot 2015-04-27 15 44 48

jebeck:
screenshot 2015-04-27 15 47 23

I don't really get why the "Detail" (here "Error uploading data") in the last part of the details is *not* the part that actually gets exposed to the user - it's more generic and user-friendly. If we do want to keep the current non-hidden-under-the-details link text what it is (i.e., here "Some arguments..."), can we just delete the final "Detail" section of what gets logged? It's the opposite of detailed...

jebeck:
Also, stack traces? And am I supposed to see the logged errors in my local server.log? Because I don't...

jebeck:
Superceded by #115, so closing.

PR #98, opened 2015-03-27
Handle (and ignore) bogus exception thrown by menu code.

kentquirk:
Addresses the bug: https://trello.com/c/yG4zNKeW

The real bug appears to be in Chrome. But I was able to make it not matter by catching the exception and ignoring it.

@jebeck @jh-bate ?

### DIFF
--- a/main.js
+++ b/main.js
@@ -33,7 +33,6 @@
 
 var contexts = [ 'page' ];
 
-
 var contextMenus = [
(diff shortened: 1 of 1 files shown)


jebeck:
Just tested myself compared to one of the store versions, and works! Awesome catch @kentquirk 

Good to merge.

PR #99, opened 2015-03-27
Insulet bugfix #3: error building settings bgTarget

jebeck:
Contains more robust searching for previous target BG or threshold when there isn't a new one for a particular start time.


kentquirk:
LGTM

PR #100, opened 2015-03-28
OnetouchUltra2 - Don't merge yet

patovala:
We are pushing this branch for your consideration, kind of dirty yet, I'll finish the data pattern matching in a couple of hours. Up to now this code:
  - Connects to the OneTouchUltra2 and requests PC communication
  - Collects the results stored in the device
  - Parses and checksums the data header.
TODO:
  - Parse the data stream
  - Build the smbg object

### DIFF
--- a/lib/drivers/oneTouchUltra2.js
+++ b/lib/drivers/oneTouchUltra2.js
@@ -80,9 +80,6 @@ module.exports = function (config) {
     if (buffer.len() < 1) {
       return false;
     }
-
-    console.log('buffer.bytes() ', buffer.bytes());
-
     // there's enough there to try, anyway
     var packet = buildPacket(buffer.bytes());
     if (packet.packet_len !== 0) {
@@ -91,7 +88,6 @@ module.exports = function (config) {
     }
 
     if (packet.valid) {
-      console.log('found packet: ', packet, ' len:', packet.packet_len, ' valid:', packet.valid);
       return packet;
     } else {
       return null;
@@ -113,10 +109,8 @@ module.exports = function (config) {
   var byteConcat = function (a, b){
       var aLength = a.length,
       result = new Uint8Array(aLength + b.length);
-
       result.set(a);
       result.set(b, aLength);
-
       return result;
   };
 
@@ -173,34 +167,11 @@ module.exports = function (config) {
   var prepBGData = function (progress, data) {
     cfg.builder.setDefaults({ deviceId: data.id });
     var dataToPost = [];
-    // var bytes = data.bytes;
-    // var lines = [];
-
-    // // split up into lines by searching for '\r\n'
-    // var startIndex = 0;
-    // var endIndex = _.indexOf(bytes, CR, startIndex + 1);
-
-    // var startLine = startIndex;
-    // while (startLine < endIndex) {
-    //   var endLine = _.indexOf(bytes, CR, startLine);
-    //   if (endLine === -1) {
-    //     endLine = endIndex - 1;
-    //   }
-    //   if (bytes[endLine + 1] != LF) {
-    //     // something borked, we should discard the packet
-    //     console.log('hmmm...bad packet with CR but not LF', bytes);
-    //     packet.packet_len = endLine;
-    //     return packet;
-    //   }
-
-    //   var line = String.fromCharCode.apply(String, bytes.buffer.slice(startLine, endLine));
-    //   lines.push(line);
-    //   startLine = endLine + 2;
-    // }
 
     //This is converting to a string the data bytes array
     var lines = String.fromCharCode.apply(null, data.bytes);
-    //Separet each data by \n
+
+    //Separe each data row by \n
     var splites = lines.split('\n');
 
     // extract header
@@ -213,17 +184,7 @@ module.exports = function (config) {
         checksum = toMatch[4],
         calcchks = 0;
 
-    console.log(
-        'nrecords:', nrecords,
-        'serialno:', serialno,
-        'oum:', oum,
-        'checksum:', checksum,
-        'calcchks:', calcchks
-      );
-
-    // Need another way to check witout endline.
     var bytes = data.bytes.subarray(0, 27, 0);
-    //var bytes = data.bytes.subarray(0, _.indexOf(data.bytes, CR, 0));
 
     calcchks = _.reduce(bytes, function(s1, e){ return s1 + e;}, 0);
 
@@ -233,7 +194,6 @@ module.exports = function (config) {
      }
 
     // get data from lines
-    //var dpattern = /P "(\w+)","(.+)","(.+)","(\d+)","(\w)","(\d+)", (\d{2}) ([[:xdigit:]]+)/;
     var dpattern = /P "(\w+)","(\d{2}\/\d{2}\/\d{2})","(\d{2}:\d{2}:\d{2})   ","..(\d{3}).","(.)","(\d{2})",\W00\W(.+)/;
     _.forEach(splites.slice(1), function(l){ // <-- jump the header
         var ms = l.match(dpattern);
--- a/lib/drivers/oneTouchUltra2.js
+++ b/lib/drivers/oneTouchUltra2.js
@@ -34,23 +34,14 @@ module.exports = function (config) {
   var serialDevice = config.deviceComms;
 
   //BYTE START DEFINITION HEX
-  var MEC = [0x11, 0x0d, 0x0a];
-  var DMF = MEC.concat([0x44, 0x4d, 0x46]);
-  var DMV = MEC.concat([0x44, 0x4d, 0x3f]);
-
-  var buildPacket = function (length, data) {
-    var datalen = length;
-    var buf = new ArrayBuffer(datalen);
-    var bytes = new Uint8Array(buf);
-    //var ctr = struct.pack(bytes, 0, 'bbbbbb', data);
-    var ctr = struct.pack(bytes, 0, 'bbbbbb', 0x11, 0x0d, 0x0a, 0x44, 0x4d, 0x46);
-    return buf;
-  };
+  var MEC = '\x11\x0d';
+  var DMF = MEC.concat('\x44\x4d\x46');
+  var DMV = MEC.concat('\x44\x4d\x3f');
 
   var otu2MessageHandler = function (buffer) {
 
     //OneTouchUltra2 returns ascii, so let's see
-    console.log('DEBUG OTU2 buffer', buffer.bytes());
+    console.log('DEBUG OTU2 buffer', buffer);
     buffer.discard(buffer.length);
     return null;
     /*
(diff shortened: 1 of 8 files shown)


patovala:
I guess we need to sync our fork with yours also... will do over the weekend.

jh-bate:
Awesome - thanks a lot.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+
+            if(createchk !== parseInt(chk, 16)){
+              console.log('hmmm...bad checksum', createchk, chk);
+              return null;
+            } else {
+              var bdt = buildDateTime(dor, tor);
+
+              var smbg = cfg.builder.makeSMBG()
+                              .with_deviceId('OneTouchUltra2 ' + serialno)
+                              //.with_deviceSerialNumber(serialno)
```
jh-bate: looks like this shouldn't be commented out

jh-bate:
So all the tests pass but there are jshint errors see lib/drivers/oneTouchUltra2.js

Travis link https://travis-ci.org/tidepool-org/chrome-uploader/builds/56632709

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+      clearTimeout(abortTimer);
+      clearInterval(listenTimer);
+
+      callback(null, pkt);
+    }, 1000);     // spin on this one not so quickly
+  };
+
+  var probe = function(cb){
+    console.log('attempting probe of oneTouch Ultra2 - don\'t probe anything');
+    //otu2CommandResponse(DM, cb);
```
jh-bate: commented out??

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+
+            var lineWithoutChecksum = 'P "' + dow + '","'+ dor + '","' + tor + '   ", "' + rf + ' ","' + alv + '","' + umc + '", 00 ';
+            var arrayBytes = [0];
+
+            for (var i = 0; i < lineWithoutChecksum.length; ++i)
+            {
+                arrayBytes.push(lineWithoutChecksum.charCodeAt(i));
+            }
+
+            createchk = _.reduce(arrayBytes, function(s1, e){ return s1 + e;}, 0);
```
jh-bate: i  think this might be your jshint issue here. prob should be var createchk. You can use npm jshint to do the check

fabricioflores: @jh-bate yes, it was the jshint issue. But we started with ```gulp jshint```

jh-bate: thanks @fabricioflores - you might have already seen this but you can also use a jshint watch while you develop https://github.com/tidepool-org/chrome-uploader/blob/master/package.json. Personally I just run the jshint after or by using an installed add-in in your editor (I use Sublime)

patovala: Thanks @jh-bate for your remarks, I didn't realise the npm task for jshint, I'll respond to them ASAP.

lib/drivers/oneTouchUltra2.js
patovala: you mean the otu2commandResponse? We realise there was no need (a strong one) to just fire a command to the device on start, since for each command we are sending a DM for collecting data later.

lib/drivers/oneTouchUltra2.js
patovala: That didn't work on first instance, we were just guessing and then decided to feed the serial as soon as we get a data row parsed. This could have sense on other devices but this one particularly sends a header with serialno and other info on each data gathering.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+      data.processData = true;
+      //console.log('processData:', data);
+      cb(null, data);
+    },
+
+    uploadData: function (progress, data, cb) {
+      progress(0);
+      var sessionInfo = {
+        deviceTags: ['bgm'],
+        deviceManufacturers: ['OneTouch'],
```
kentquirk: The manufacturers field here should be ['LifeScan'], please.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+var struct = require('../struct.js')();
+
+module.exports = function (config) {
+  var cfg = _.clone(config);
+  cfg.deviceData = null;
+  var serialDevice = config.deviceComms;
+
+  //BYTE START DEFINITION HEX
+  var DM = '\x11\x0d';
+  var DMDateTime = DM.concat('\x44\x4d\x46');
```
kentquirk: For ease of debugging (looking at data dumps, etc) it's helpful not to use the hex codes for easily printable characters. 
I'd rather have you put this as 'DMF' instead of '\0x44\x4d\x46', and DM@, DMP, DMSU? and DMST? etc.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+  };
+
+  var otu2ErrorHandler = function(info) {
+    // we don't care too much for timeout errors
+    if(info.error && info.error!=='timeout'){
+      console.log('DEBUG OTU2ERROR:', info);
+    }
+    if (info.connectionId && info.error) {
+      if (info.error == 'timeout') {
+        return -9; // TODO check this
```
kentquirk: Returning a value from the error handler does nothing. The only thing you can do here is log the result. See https://developer.chrome.com/apps/serial#event-onReceiveError

lib/drivers/oneTouchUltra2.js
kentquirk: @patovala we try not to check in commented-out code unless there's a good reason to keep it there, in which case the reason should also be documented.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+    var lines = String.fromCharCode.apply(null, data.bytes);
+
+    //Separe each data row by \n
+    var splites = lines.split('\n');
+
+    // extract header
+    var hpattern = /P (\d{3}),"(\w{9})","(\w+\/\w+).*" (\w+)/;
+
+    var toMatch = hpattern.exec(splites[0]);
+    var nrecords = toMatch[1],
```
kentquirk: You don't check here to make sure that the pattern matched before you start using the matches. Please test for failure in case someone plugs in a device that isn't compatible.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+
+    //set missing serial number and id
+    data.serialNumber = serialno;
+    data.id = data.id + serialno;
+
+    // var bytes = data.bytes.subarray(0, 27, 0);
+    var stringWithoutChk = 'P ' + nrecords + ',"' + serialno + '","' + oum + ' "';
+    var bytes = [0];
+
+    for (var i = 0; i < stringWithoutChk.length; ++i)
```
kentquirk: Ok, this is ... strange. You assemble a string and then check the checksum of the assembled string. This goes against the very concept of a checksum. Why did you do this instead of something like the subarray command that you commented out? Please calculate the checksum correctly.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+    if(calcchks !== parseInt(checksum, 16)){
+         console.log('hmmm...bad checksum', calcchks, checksum);
+         return null;
+     }
+
+    // get data from lines
+    var dpattern = /P "(\w+)","(\d{2}\/\d{2}\/\d{2})","(\d{2}:\d{2}:\d{2})   ","..(\d{3}).","(.)","(\d{2})",\W00\W(.+)/;
+    _.forEach(splites.slice(1), function(l){ // <-- jump the header
+        var ms = l.match(dpattern);
+        if(ms){
```
kentquirk: Please log the result if this test fails.

lib/drivers/oneTouchUltra2.js
kentquirk: That's why there are separate steps for fetchData, processData, and uploadData -- you should be doing the assembly of data in the processData step, so that you have all the information you need at that point.

lib/drivers/oneTouchUltra2.js
```diff
@@ -0,0 +1,386 @@
...
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+/**
+ * IOET's code start here
+ * pvalarezo@ioet.com
+ *
+ * The registers looks like:
+ * P "SAT","03/21/15","16:45:24   ","  081 ","N","03", 00 09B6
```
kentquirk: Can you also show an example of the header, please?

lib/state/appActions.js
```diff
@@ -276,6 +276,7 @@ appActions.addOrRemoveTargetDevice = function(e) {
       return device === e.target.value;
     });
   }
+
```
kentquirk: It doesn't matter very much, but please try not to check in meaningless changes. You should be reviewing the files you check in to look for stuff like this and revert them before committing.

lib/drivers/oneTouchUltra2.js
patovala: mmm... yes,  we were trying to avoid "magic numbers"  like the 27 on line 192, in the device spec the character position seems to be determined, but to be sure we recreated the whole line, I think it should be a better way to do it.

kentquirk: I think you're saying that you think reassembling the string is a better way to do it. I really disagree. You can use a search or a pattern match to find the position of the checksum at the end of the string, and then calculate the checksum of the other bytes. Please do that.

kentquirk:
Can you please re-merge from master and make sure it passes? Sending separate email.

lib/drivers/oneTouchUltra2.js
kentquirk: I'm sorry to insist, but this is really the wrong way to do it. The whole point of a checksum is to validate the actual bytes received to make sure that what you receive is what was sent. Doing this by parsing and then reassembling the string goes against the entire idea. Please write a loop over the original data to calculate the checksum.

lib/drivers/oneTouchUltra2.js
kentquirk: If you're going to test that there was a match, you should have some action if there was NOT a match (as it might be an indication that something has broken). Please log a failure here.

lib/drivers/oneTouchUltra2.js
patovala: let me double check I fixed this long time ago, maybe the rebase I did yesterday could make this lost.

lib/drivers/oneTouchUltra2.js
kentquirk: Actually, you should just delete the commented-out line. As you mention, it's not correct -- so please don't leave it in there.

lib/drivers/oneTouchUltra2.js
patovala: fixed on last push

kentquirk:
I think we finally have a winner. Thank you.

HowardLook:
Possible to change the title? It says "Don't merge yet".

PR #101, opened 2015-03-30
Jhbate/tz picker

jh-bate:
TODO: UI review with @brandonarbiter 

- using autocomplete drop downlist (see https://github.com/JedWatson/react-select)
- bound to appActions to update TZ on selection in settings and added to app state


jh-bate:
@jebeck initial look through feels all good. Was just going to run locally first then give the a-ok

jh-bate:
LGTM

PR #102, opened 2015-04-01
regroup alarm switch statement according to proper alarmTypes

jebeck:
At some point the switch statement for Insulet alarms got consolidated a bit *too much* - almost everything was getting created an an `other` alarmType, which is incorrect.

I also changed some of the alarm names again to match the spec exactly - it's too confusing for me looking back and forth when trying to verify that everything is being handled if they aren't exactly the same.

Discovered this in the course of working on the UTC bootstrapping because there are some challenges dealing with the alarms and their timestamps...

I'd like to get this in ASAP, so whoever can get to it @jh-bate @kentquirk?


jh-bate:
LGTM

PR #103, opened 2015-04-02
OmniPod UTC "bootstrapping"

jebeck:
WIP.

Still some TODOs - tests and making sure this is "turn-off-able" so that we can merge but wait to "activate" until we have all devices. All the important logic is in `TimezoneOffsetUtil`, and I think it should be generalizable to all the drivers.

Also still need to investigate how it'll work with non-fresh accounts.

It'd be great if we could chat about this tomorrow @kentquirk.


jebeck:
Closing this in favor of #106

PR #104, opened 2015-04-06
ensure we're using the correct start time for Insulet bgTarget settings

jebeck:
Fixes a bug with Insulet bgTarget settings under certain circumstances (namely, the user only has a single setting for the "correct above" high threshold but multiple target BGs at different times of day). In such circumstances, the start time for the high threshold (midnight) was being applied to all the settings instead of the start time for the target BGs.


jh-bate:
LGTM

PR #105, opened 2015-04-06
update lodash to v3.6.0

jebeck:
What it says on the tin.

It was increasingly frustrating working with a 2.x version of lodash because they don't allow you to view the documentation for older versions and some method names have changed.


jh-bate:
sat through the demo LGTM

PR #107, opened 2015-04-07
use node 0.12 in Travis build

jebeck:
The latest production uploader was already built using node 0.12.0, so Travis should be using it too!

Ping @kentquirk for review/merge.


kentquirk:
LGTM

PR #108, opened 2015-04-08
fix (my) bad lodash update from has to includes

jebeck:
Misreading of the lodash changes on my part :/

HT @kentquirk for finding the problem very quickly. This fixes both of the uploader UI bugs.

I should probably add a test of the tiny `getIn` module on this branch, huh?


kentquirk:
Cool. Thanks for the tests.

PR #109, opened 2015-04-17
Clean up clean up

jh-bate:
@kentquirk if you could have a look at this clean that @jebeck and I did.

- remove mock mode and mocked components
- remove running in browser and also server to do so
- move images into separate folder
- cleanup readme to the essentials 
- removed redundant bin/parse.js file


jh-bate:
@kentquirk is you get a change to give this a quick look it would be appreciated - thanks Jamie

README.md
```diff
@@ -2,62 +2,10 @@
 
 [![Build Status](https://travis-ci.org/tidepool-org/chrome-uploader.png)](https://travis-ci.org/tidepool-org/chrome-uploader)
 
-These notes were updated on Tuesday, December 30, 2014.
```
kentquirk: I don't care too much, but the reason I put such things in is that people can easily know how current the notes are.

kentquirk:
Looks great to me. It was badly needed. I'll let you merge it.

PR #110, opened 2015-04-27
Jebeck/bugfix carelink suppressed

jebeck:
Ping @jh-bate for review.


jebeck:
Hmm, restarting Travis build that failed. Looks like it just failed to bring in an optional dependency and gave up? Odd.

lib/carelink/carelinkSimulator.js
```diff
@@ -1006,6 +1010,16 @@ exports.make = function(config){
      */
     getEvents: function(){
       return events;
+    },
+
+    /**
+     * Return a boolean reflecting whether or not this simulator instance has seen
```
jh-bate: Should we explain why we care? If we are only using it because of the isolated issue that occurred should we explain that

jh-bate:
Code LGTM - just think we need some explanation in the code as to why we are checking this now so it still makes sense at a latter date

jebeck:
OK, I expanded the comment, merging.

PR #111, opened 2015-04-30
set index for PX BG and ketone entries

jebeck:
This is ready for review. Ping @jh-bate.

### DIFF
--- a/lib/drivers/abbottFreeStyle.js
+++ b/lib/drivers/abbottFreeStyle.js
@@ -429,7 +429,9 @@ module.exports = function (config) {
           .with_timezoneOffset(datum.datetime.offset)
           .with_time(datum.datetime.utc)
           .with_units('mg/dL')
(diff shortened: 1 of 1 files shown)


lib/drivers/abbottFreeStyle.js
```diff
@@ -429,6 +429,7 @@ module.exports = function (config) {
           .with_timezoneOffset(datum.datetime.offset)
           .with_time(datum.datetime.utc)
           .with_units('mg/dL')
+          .set('index', i)
```
jh-bate: is this temporary? if so should we make  comment to that effect

jh-bate:
LGTM

lib/drivers/abbottFreeStyle.js
jebeck: Oh yeah I'll add the same comments I've used elsewhere to the effect of 'delete after Jaeb'.

PR #112, opened 2015-04-30
ensure delete of index at top level (should only be in payload.logIndices)

jebeck:
Index was appearing in two places for the PX data...

Also added a comment in the objectBuilder about why the deletion can't happen when `.done()` is called (because it took me a min to remember why I didn't put it there).


jh-bate:
LGTM - just waiting for build

PR #114, opened 2015-05-04
OneTouchUltraMini - fixes from IOET team

patovala:
Creating pull request for your consideration, fixes OneTouchUltraMini driver and minor fixes on OneTouchUltra2


jebeck:
Can you update with a merge from master and/or fix the merge conflicts and update this PR please?

lib/drivers/oneTouchMiniDriver.js
```diff
@@ -437,20 +438,27 @@ module.exports = function (config) {
...
-        .with_units('mg/dL')
-        .done();
-      dataToPost.push(smbg);
+    if (data.bgmReadings.length > 0) {
+      for (var i = 0; i < data.bgmReadings.length; ++i) {
+        var datum = data.bgmReadings[i];
+        var smbg = cfg.builder.makeSMBG()
+          .with_value(datum.glucose)
+          .with_deviceTime(datum.displayTime)
+          .with_timezoneOffset(TZOFFSET / 60)
```
jebeck: I know this is an oversight from before you took over the code, but timezone offset should not be a constant - it varies according to whether Daylight Savings Time is in effect. Can you refactor the code to generate the display and UTC timestamps as well as the timezone offset in the same way as is done [here](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/drivers/oneTouchUltra2.js#L56) in the `buildDateTime`, particularly with the three calls to sundial's `formatDeviceTime`, `applyTimezone`, and `getOffsetFromZone` methods? Thanks!

CC @brandonarbiter @HowardLook

lib/drivers/oneTouchMiniDriver.js
```diff
@@ -560,7 +568,18 @@ module.exports = function (config) {
 
     uploadData: function (progress, data, cb) {
       progress(0);
-      cfg.jellyfish.post(data.post_records, progress, cfg.groupId, function (err, result) {
+      var sessionInfo = {
+        deviceTags: ['bgm'],
+        deviceManufacturers: ['LifeScan'],
+        deviceModel: 'OneTouchultraMini',
```
jebeck: Can you make the deviceModel `OneTouch UltraMini`?

PR #115, opened 2015-05-07
KISS user-visible errors

jebeck:
In order to facilitate my work on adding CGM settings for Dexcom in the uploader, I branched off #96 to finish up the expansion of user-visible errors. Along the way - because I was testing a whole lot of different error states to make sure they appeared in a useful way - I fixed a few bugs that I noticed or were already outstanding but easy to fix.

The only thing on [this checklist](https://trello.com/c/kcC9Flqc) that's left to do is the changing of logging of errors by groupId instead of userId. I'm having a hard time finding where to make that change (in the uploader? platform-client?) and if it's the latter, whether it's possible to change it only for uploader errors, like we want.

List of bugfixes:
- 'Last upload' only appears if the upload was actually successful
- Upload button *and* functionality of said button is properly disabled when an upload is already in progress (I believe this is a regression that may have come in with the groups work o_O)
- The weird thing in the logging where strings we're being treated as arrays and stuff was logging one letter at a time is fixed, I believe, although it'd be good to verify this on a live highwater since I had to checkout a pre-AWS version locally to be able to see my logs locally
- The show/hide details for the errors would open and close *all* the detail sections if there were multiple uploads with errors; now it works as expected, only showing and hiding for the one you're clicking on

I also added a "debug" mode (added via an environment variable in config/local.sh, and using webpack's [DefinePlugin](http://webpack.github.io/docs/list-of-plugins.html#dependency-injection)) that reverts to the old behavior of throwing the errors instead of catching them. I think this is the easiest and best way (possibly the *only* way, at least without going down a significant rathole of hacks) to expose file numbers and line numbers of errors to developers while developing; without this, it is really, *really* hard to figure out what's failing. (The stack trace that I've added to the debug string for error logging only includes function names, which are sometimes useful and sometimes not, but the method of getting the stack traces it uses won't point to file numbers and line numbers properly due to module bundling. Which is to say: that stack trace is not enough to facilitate developer happiness.)

What I included in the debug string for logging (and in the details if the user shows them) is the following information:
- the detailed `message` from the error
- UTC time
- the error `code` identifying where in the app flow (login, file selection, CareLink fetch, etc.) the error occurred (from original #96 and earlier work)
- the error type - e.g., just plain `Error` or if it's another kind of JavaScript error that we caught but didn't throw, it'll be `RangeError`, etc. - I added this, and I think it's really useful; I also put in some logic to identify the errors coming back from failed jellyfish POSTs and gave those an error type 'POST error'
- the uploader version
- (if available) a stringified version of the error's stack trace, just listing the function names in the stack since nothing else is useful when the modules are bundled; tried this with a bunch of errors and it's moderately useful

### Examples

Handle incorrect CareLink creds:
handle carelink incorrect creds

Handle erroring out of upload because a POST to jellyfish fails (likely data failing validation):
handling of jellyfish post error

Handle an error caught but not thrown by our code, in this case naming a blank file foo.ibf and trying to upload it as an OmniPod file won't fail at the `FileReader` (because `FileReader` can read an empty file perfectly fine), but it will fail in one of the earliest stages of the Insulet driver:
handling of caught error not thrown by us text file with ibf extension

Handle selection of non-.ibf file for OmniPod:
handle selection of non- ibf file for omnipod

Handle device not connected:
handle dexcom not connected

Handle Dexcom connection error during the course of upload (when I was figuring out the debug mode I managed to produce this because of briefly borking the driver manager flow!):
handle dexcom connection error

Handle an error in the object builder during data processing in any device driver:
handle object builder error

@jh-bate whenever you're free from the OAuth deadlines, I'd appreciate a look at this. I'd like to get it in ASAP (i.e., before UTC bootstrapping).


lib/state/appActions.js
```diff
@@ -127,8 +160,8 @@ appActions.load = function(cb) {
     setHostsWithCallback,
   ], function(err, results) {
     if (err) {
-      var error = new UploaderError(err.message, appActions.errorStage.STAGE_SETUP, err);
-      return cb(error);
+      appActions.addMoreInfoToError(err, 'STAGE_SETUP');
```
jh-bate: I think we  should just just pass in the actual stage 
 
``appActions.addMoreInfoToError(err, appActions.errorStage.STAGE_SETUP);``

lib/state/appActions.js
```diff
@@ -53,18 +54,50 @@ appActions.errorText = {
...
+}
 
+function getErrorName(err) {
+  if (err.name) {
+    return err.name;
+  }
+  return 'POST error';
+}
+
+appActions.addMoreInfoToError = function(err, stage) {
```
jh-bate: as below I think we should just pass in the actual stage 

``appActions.addMoreInfoToError(err, appActions.errorStage.STAGE_SETUP);``

lib/state/appActions.js
```diff
@@ -18,11 +18,12 @@
 
 var _ = require('lodash');
 var async = require('async');
+var stacktrace = require('stack-trace');
```
jh-bate: sweet!

config/local.sh
```diff
@@ -1,4 +1,4 @@
 export API_URL='http://localhost:8009'
 export UPLOAD_URL='http://localhost:9122'
 export BLIP_URL='http://localhost:3000'
-export MOCK='false'
+export BUILD_DEV=true
```
jh-bate: do you think we should document what this flag allows or is intended for?

lib/drivers/carelinkDriver.js
```diff
@@ -221,6 +221,12 @@ module.exports = function(simulatorMaker, api){
         progress(20);
 
         function getUnits(colName) {
+          if (colName == null) {
+            var err = new Error('No column labels; check CareLink credentials.');
```
jh-bate: So standard [js errors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) don't have a code / friendlyMessage. Do you think given that you are throwing a generic error here and then attaching the details to it that it could still be worth having an uploader error type ?

lib/serialDevice.js
```diff
@@ -255,7 +255,10 @@ module.exports = function(config) {
         } else {
           connection = null;
           clearPacketHandler();
-          cb(new Error('Could not connect to a matching device port ' + deviceInfo));
+          var deviceDebugInfo = 'driverId ' + deviceInfo.driverId + ', vendorId ' + deviceInfo.vendorId +
+            ', productId ' + deviceInfo.productId + ', portPattern ' + deviceInfo.portPattern +
+            ', usbDevice ' + deviceInfo.usbDevice;
+          cb(new Error('Could not connect to a matching device port: ' + deviceDebugInfo));
```
jh-bate: and if is useful to add the detail above why isn't it here?

webpack.config.js
```diff
@@ -1,8 +1,19 @@
...
+var webpack = require('webpack');
 
-if ((!process.env.API_URL || !process.env.UPLOAD_URL || !process.env.BLIP_URL) &&
-     process.env.MOCK != 'true') {
+var definePlugin = new webpack.DefinePlugin({
+  __DEV__: JSON.stringify(JSON.parse(process.env.BUILD_DEV || 'false'))
+});
+
+if (process.env.BUILD_DEV === 'true') {
+  console.log('~ ~ ~ ~ ~ ~ ~ ~ ~ ~');
```
jh-bate: nice decoration!

lib/components/Upload.jsx
```diff
@@ -194,17 +203,17 @@ var Upload = React.createClass({
...
     }
     if (this.isUploadSuccessful()) {
-      return <div className="Upload-status Upload-status--success">{'Done!'}</div>;
+      return <div className="Upload-status Upload-status--success">{this.props.text.UPLOAD_COMPLETE}</div>;
     }
     if (this.isUploadFailed()) {
-      var uploadError = this.getUploadError();
-      return <div className="Upload-status Upload-status--error">{uploadError.message}</div>;
+      //nothing to show here
+      return <div className="Upload-status Upload-status--error"></div>;
```
jh-bate: can we just get rid of this? not sure

PR #119, opened 2015-05-14
small debug string creation fix

jebeck:
Discovered in the course of diagnosing issues with the OneTouch driver errors (not) surfacing.

Borked the commit message, should say: properly provide debug info for jellyfish messages with only a `reason` instead of a `message`.

Ping @jh-bate for a quick review.


jh-bate:
LGTM - thanks for the comments

PR #120, opened 2015-05-15
filter out groups we can't upload to

jh-bate:
FYI this was failing login as an exception was occurring.

This is to cater for the use case that an account it not setup as a data account BUT has upload rights to another linked account.

### DIFF
--- a/lib/components/UploadSettings.jsx
+++ b/lib/components/UploadSettings.jsx
@@ -26,25 +26,40 @@ var UploadSettings = React.createClass({
     targetId: React.PropTypes.string,
     isUploadInProgress: React.PropTypes.bool
   },
-  render: function() {
-    // we're already doing a check to see if we want to render in App.jsx
-    // but this is an extra measure of protection against trying to render
-    // when we don't have the groups to do so
+
+  availableUploadGroups: function(){
+    //can only upload for yourself
     if (_.isEmpty(this.props.user.uploadGroups) || this.props.user.uploadGroups.length <= 1) {
       return null;
     }
-    var self = this;
 
-    // sort users alpha by full name
-    var sortedGroups = _.sortBy(this.props.user.uploadGroups, function(group) {
-      if(group.profile.patient.isOtherPerson){
+    //only groups we can upload too, e.g. some people simply aren't `patients` and might be setup without data storage
+    var available = _.filter(this.props.user.uploadGroups, function(group) {
+      return _.isEmpty(group.profile.patient) === false;
+    });
+
+    //and now return them sorted them by name
+    return _.sortBy(available, function(group) {
+      if (group.profile.patient.isOtherPerson) {
         return group.profile.patient.fullName;
       }
       return group.profile.fullName;
     });
+  },
+
+  render: function() {
+    var sortedGroups = this.availableUploadGroups();
+
+    if (_.isEmpty(sortedGroups)){
+      // we're already doing a check to see if we want to render in App.jsx
+      // but this is an extra measure of protection against trying to render
+      // when we don't have the groups to do so
+      return null;
+    }
 
     var options = _.map(sortedGroups, function(group) {
-      if(group.profile.patient.isOtherPerson){
+      //some people simply aren't `patients` and might be setup without data storage
+      if(_.isEmpty(group.profile.patient) || group.profile.patient.isOtherPerson){
         return (
           <option key={group.userid} value={group.userid}>{group.profile.patient.fullName}</option>
         );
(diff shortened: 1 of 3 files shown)


jebeck:
This works UI-wise but it *uploads to the wrong account*.

How I tested: created a new local account (Mary Doe) with no data storage. Gave upload permission to Mary Doe for Jane Doe's account. In the uploader, the dropdown only shows me Jane Doe, as the changes on this branch are intended to do. I uploaded to Jane Doe, went to refresh in blip - nothing. Then I went and set up a data storage account for Mary Doe (which really already existed, just wasn't marked as active or whatever). And voila! Mary Doe magically has data (because that's where it went when I tried to upload to Jane).

P.S. This branch also needs an update from master - there have been a few small changes.

jh-bate:
@jebeck should be good to go now - tested locally and all worked as expected

jebeck:
How did you test locally @jh-bate? I still get the exact same undesired result with the testing situation I described in my previous comment... (Logging in as Mary Doe, who has signed up only to receive Jane Doe's invitation to view & upload on behalf of Jane and does not have her own data storage set up.)

Digging a little deeper into the code changes now to understand what's not behaving...

lib/components/UploadSettings.jsx
```diff
@@ -26,25 +26,40 @@ var UploadSettings = React.createClass({
...
-    var sortedGroups = _.sortBy(this.props.user.uploadGroups, function(group) {
-      if(group.profile.patient.isOtherPerson){
+    //only groups we can upload too, e.g. some people simply aren't `patients` and might be setup without data storage
+    var available = _.filter(this.props.user.uploadGroups, function(group) {
+      return _.isEmpty(group.profile.patient) === false;
+    });
+
+    //and now return them sorted them by name
+    return _.sortBy(available, function(group) {
+      if (group.profile.patient.isOtherPerson) {
```
jebeck: In my [testing scenario](https://github.com/tidepool-org/chrome-uploader/pull/120#issuecomment-104797939), which I believe is the same kind of case as we're trying to set up for Jaeb, there is no `isOtherPerson` property on Jane Doe's entry in Mary Doe's `uploadGroups`.

Is the difference between the results we're seeing @jh-bate the fact that you're manipulating account set-ups directly in the database? Because I'm creating accounts through blip locally...

lib/components/UploadSettings.jsx
```diff
@@ -26,25 +26,40 @@ var UploadSettings = React.createClass({
...
+      // we're already doing a check to see if we want to render in App.jsx
+      // but this is an extra measure of protection against trying to render
+      // when we don't have the groups to do so
+      return null;
+    }
 
     var options = _.map(sortedGroups, function(group) {
-      if(group.profile.patient.isOtherPerson){
+      //some people simply aren't `patients` and might be setup without data storage
+      if(_.isEmpty(group.profile.patient) || group.profile.patient.isOtherPerson){
```
jebeck: I think `_.isEmpty(group.profile.patient)` is redundant (and in fact will never be true) given the logic in `availableUploadGroups`?

lib/state/appActions.js
```diff
@@ -236,16 +236,39 @@ appActions.login = function(credentials, options, cb) {
...
+appActions._getDefaultTargetId = function(loggedInUser){
+  //default to the logged in users id
+  var userid = loggedInUser.userid;
+
+  //even though we have multiple groups there may still only be one upload account
+  if (_.isEmpty(loggedInUser.uploadGroups) === false || loggedInUser.uploadGroups.length > 1) {
+
+    var possibilities = _.filter(loggedInUser.uploadGroups, function(group) {
+      //if the account isn't marked with patient details we don't upload to it
+      return _.isEmpty(group.profile) === false || _.isEmpty(group.profile.patient) === false;
```
jebeck: I don't think this logic is right... As far as I can tell, `group.profile` is never empty, so you're always returning `true` in this filter and not filtering anything out.

lib/state/appActions.js
```diff
@@ -236,16 +236,39 @@ appActions.login = function(credentials, options, cb) {
   });
 };
 
+//make sure we have the correct default targetId
+appActions._getDefaultTargetId = function(loggedInUser){
+  //default to the logged in users id
+  var userid = loggedInUser.userid;
+
+  //even though we have multiple groups there may still only be one upload account
+  if (_.isEmpty(loggedInUser.uploadGroups) === false || loggedInUser.uploadGroups.length > 1) {
```
jebeck: This is really confusing to read, and I'm still working out whether the logic is right. Wouldn't `!_.isEmpty(loggedInUser.uploadGroups)` (and no `||...`) accomplish the same thing? Because what you're asking here is `uploadGroups.length !== 0 || uploadGroups.length > 1`, which is just the same thing is `uploadGroups.length > 0`.

jh-bate:
closing this as it has been super-seeded by https://github.com/tidepool-org/chrome-uploader/pull/122

PR #121, opened 2015-05-16
properly deal with errors when jellyfish is down

jebeck:
What it says on the tin...

### DIFF
--- a/lib/state/appActions.js
+++ b/lib/state/appActions.js
@@ -68,12 +68,12 @@ appActions.errorStages = {
 };
 
 function extractMessage(err) {
-  // if there's no message, just try to stringify the whole obj
-  if (!err.message && typeof err.error === 'string') {
-    return err.error;
+  // if there's no message, return `error` or 'Unknown error message'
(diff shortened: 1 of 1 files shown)


lib/state/appActions.js
```diff
@@ -68,6 +68,10 @@ appActions.errorStages = {
 };
 
 function extractMessage(err) {
+  // if there's no message, just try to stringify the whole obj
+  if (!err.message && typeof err.error === 'string') {
```
jh-bate: is there a reason why not continuing on with the if else flow you have below??

also do you think this reads a bit better ?

``if (_.isEmpty(err.message) && typeof err.error === 'string' ) { ...``

jebeck:
Adjusted @jh-bate. LMK if you think this is an improvement.

I did change to `_.isEmpty` but in general I'm not that much of a fan unless it's an object or an array you're checking - I find it impossible to remember/unpredictable how it behaves with other things. :/

jh-bate:
LGTM

PR #122, opened 2015-05-23
upload target fix

jebeck:
I think this fixes the issues I'm still seeing in #120, but I'm putting it up separately in case there's still confusion about the properties of the accounts we're concerned with (although I went over these with Nick to check my testing scenario, and he thought what I was doing sounded right).

The key changes (relating to my comments on bits of the logic in #120) are in commit b36f1740a560e7a531b207d055d8217345490abc - I'd look at that diff. Then I started working on surfacing the situation when someone who has no data storage set up for their own account and no permission to upload on behalf of someone else logs in, but I quickly realized that's a larger task because we can't surface the error through the upload errors - it should just be a different "page" of the app I think. So for now this just logs a warning in the console.

### DIFF
--- a/lib/components/UploadSettings.jsx
+++ b/lib/components/UploadSettings.jsx
@@ -58,8 +58,8 @@ var UploadSettings = React.createClass({
     }
 
     var options = _.map(sortedGroups, function(group) {
-      //some people simply aren't `patients` and might be setup without data storage
-      if(_.isEmpty(group.profile.patient) || group.profile.patient.isOtherPerson){
+      // for child accounts
+      if (group.profile.patient.isOtherPerson){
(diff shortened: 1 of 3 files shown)


jebeck:
I also think we might want to expand the tests - I'm a little concerned about how the changes I made didn't break any of them. But I only have one thing left to do on the TZ picker so I may wait until Sunday or Monday.

jh-bate:
My accounts were generated, the one's in via blip you can't make as "upload only" accounts. I will setup and test both ways to make sure its still true

jh-bate:
LGTM tested using both accounts created in blip and generated by command line (as are used in studies )

PR #123, opened 2015-05-27
Bayercontournext

patovala:
We are publishing the bayer contour next driver HID for your consideration.


jebeck:
Some initial comments:
- error handling looks good this time :)
- there's an `NaN` showing briefly for the progress bar percentage somewhere around 50%, can you look into that?
- also please read the BG units of measure from the BG records (should be `mg/dL` but there's an outside chance someone somewhere will have a `mmol/L` device, and we want to handle that correctly)
- and please read the thresholds for LO/HI BG readings and store any LO or HI values plus an annotation on the object as described [here](http://developer.tidepool.io/data-model/v1/annotations/) - look for `bg/out-of-range` (You could also look at the Abbott driver starting [here](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/drivers/abbottFreeStyle.js#L112) for a rough example.)

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+  };
+
+  var getAnnotations = function (annotation){
+    if (annotation == '>') {
+      return [{
+        code: 'bg/out-of-range',
+        value: 'high',
+        threshold: 600
```
jebeck: According to my read of the spec, you should be able to get the threshold out of the header record as glucose HI / LO range rather than hard-coding as 600. Can you do that? Again to guard against devices that could be in mmol/L...

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+      }];
+    } else if (annotation == '<') {
+      return [{
+        code: 'bg/out-of-range',
+        value: 'low',
+        threshold: 20
```
jebeck: Same comment as above.

lib/state/appState.js
```diff
@@ -59,6 +59,11 @@ appState.getInitial = function() {
       name: 'OneTouch UltraMini',
       key: 'onetouchmini',
       source: {type: 'device', driverId: 'OneTouchMini'}
+    },
+    {
+      name: 'BayerContourNext',
```
jebeck: Can you put spaces in here - `Bayer Contour Next`? Although this makes me realize we should refactor so the display name is only in one place. (This one gets displayed in the device list, the one from `deviceInfo.js` gets displayed in the upload list.) I'll put the refactor on my TODO. ;)

jebeck:
One last thing which I forgot to mention the other day (my fault, sorry!)...for the research team we're adding this meter for, they'd like an additional field for sorting data that's not the timestamps. Can you add the index of the record from the `data.bgmReadings` array to each object following the model [here](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/drivers/abbottFreeStyle.js#L432) - lines 432-434?

The end result (with some stuff the `objectBuilder` is doing) will be the index in a `payload.logIndices` on each event, like so if the index is `i`:
```
{
  type: 'smbg',
  value: 100,
  ...
  payload: { logIndices: [i] }
}
```

Thanks!

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+        data.processData = true;
+        cb(null, data);
+    },
+
+    uploadData: function (progress, data, cb) {
+      progress(0);
+      var sessionInfo = {
+        deviceTags: ['bgm'],
+        deviceManufacturers: ['Bayer'],
+        deviceModel: 'BayerContour Next',
```
jebeck: Just `Contour Next` for deviceModel please, since Bayer is in the manufacturers.

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+
+var isBrowser = typeof window !== 'undefined';
+var debug = isBrowser ? require('../bows')('BCNextDriver') : debug;
+
+
+module.exports = function (config) {
+  var cfg = _.clone(config);
+  var hidDevice = config.deviceComms;
+
+  var STX = 0x02;
```
kentquirk: technically, STX, ETB, ETX, and EOT are all ASCII_CONTROL characters as well (http://ascii.cl/control-characters.htm). They should probably also go into the collection below.

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+  };
+
+  var probe = function(cb){
+    debug('attempting probe of Bayer Contour Next');
+  };
+
+  var bcnPacketHandler = function (buffer) {
+    // we need to get the REPORT_BYTES here for the first communication
+    // time if they are not set
+    var discardCount = 3;
```
kentquirk: I don't see how this can be correct. If you want to always discard the first 3 bytes of the packet, why the conditional? And why not just set the minimum packet size to 9 and handle those 3 bytes in the packet processor? At the very least this doesn't make sense and it seems like it must be wrong.

If it's NOT wrong, then you need to put a comment explaining why you're doing something so strange.

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+
+    if (packet.valid) {
+      return packet;
+    } else {
+      return null;
+    }
+  };
+
+  var buildPacket = function (command, cmdlength) {
+    var datalen = cmdlength + 4; // We need just 5 bytes for now
```
kentquirk: bad comment?

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+    var buf = new ArrayBuffer(datalen);
+    var bytes = new Uint8Array(buf);
+    var ctr = struct.pack(bytes, 0, 'bbb', REPORT_BYTES.reportID,
```
kentquirk: You're only packing 3 bytes into the struct, but you're passing at least 5 bytes worth of data. Why?

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+                          REPORT_BYTES.hostID, REPORT_BYTES.deviceID,
+                          STX, command);
+    if (cmdlength) {
+      ctr += struct.pack(bytes, 3, 'bb', cmdlength, command);
```
kentquirk: Instead of 3 here you probably want "ctr".

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+        }
+    };
+  };
+
+  var buildAckPacket = function() {
+    return buildPacket(ASCII_CONTROL.ACK, 1);
+  };
+
+  var buildHeaderPacket = function() {
+    var cmd = 0x3a;  // just anything 'x'
```
kentquirk: what does this comment mean? 0x3a is a ':' character.

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+    if (packet_len > plen) {
+      return packet;
+    }
+
+    //discard the length byte from the begining
+    var tmpbuff = new ArrayBuffer(packet_len);
+    struct.copyBytes(tmpbuff, 0, bytes, packet_len, 1);
+    packet.bytes = tmpbuff;
+
+    // we now have enough length for a complete packet, so calc the CRC
```
kentquirk: You're not calculating a CRC, so maybe edit the comments?

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+    packet.valid = true;
+
+    return packet;
+  };
+
+
+  var buildHeaderCmd = function() {
+    return {
+      packet: buildAckPacket(),
+      parser: function (result) {
```
kentquirk: You use this same parser in multiple places -- maybe it should become its own function so you don't repeat yourself.

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+                          function(e){
+                            return String.fromCharCode(e);
+                          }).join('');
+        result.payload = tostr;
+        return tostr;
+      },
+    };
+  };
+
+  var parseHeader = function (s){
```
kentquirk: Please put an example of the header in a comment. This split/shift/pop stuff is dizzying.

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,510 @@
...
+      });
+
+      if((sum % 256) !== parseInt(check, 16)){
+          return null;
+      }else{
+          return data[0];
+      }
+  }
+
+  var parseDataRecord = function (str, callback){
```
kentquirk: Same here -- examples of the format would be helpful

lib/drivers/oneTouchMiniDriver.js
```diff
@@ -486,6 +486,9 @@ module.exports = function (config) {
 
     connect: function (progress, data, cb) {
       debug('in connect!');
+
+      data.deviceInfo.deviceId = data.id;
```
kentquirk: why is this change part of a ContourNext pull request? This should probably be a completely separate PR.

test/testHidDevice.js
```diff
@@ -0,0 +1,83 @@
...
+ * Test for new hidDevice facilities
+ * */
+
+/* global beforeEach, describe, it */
+
+var expect = require('salinity').expect;
+
+var hidDevice = require('../lib/hidDevice.js');
+
+describe('hidDevice.js', function(){
```
kentquirk: I started the pattern of "exists" testing because I was building multiple things that needed to conform to an interface. But it feels like this has now become a pattern that make it look like there are some tests but without actually doing anything useful.

This is as much for Tidepool as IOET, but it seems like it would be nice to at least make a stab at putting some useful test code in place -- actually test SOMETHING. For a HID device, the test won't be able to go all the way to the device, but some of those functions at least could have something more meaningful in them.

lib/drivers/bayerContourNext.js
patovala: We found in the first device we got, three bytes not mentioned in the reference, we needed to assert if all devices behave the same before refactor this code. The second device we got has the same 3 bytes which are ABC, so we will refactor this.

lib/drivers/bayerContourNext.js
patovala: we need to add ABC always before sending anything, so this constructor only adds those bytes

lib/drivers/bayerContourNext.js
patovala: ok, good catch

lib/drivers/bayerContourNext.js
patovala: The documentation states we should send anything to start the conversation, but at the end we are only using ACKs, I think we can remove this function safely.

jebeck:
Can you update this from master so it's mergeable?

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,501 @@
...
+    if(data){
+      var result = data.trim().match(/^.*\d+R\|(\d+).*Glucose\|(\d+)\|(\w+\/\w+).*\|{2}(.*)\|{2}(\d{12})$/).slice(1,6);
+      callback(null, result);
+    }else{
+      throw( new Error('Invalid record data') );
+    }
+
+  };
+
+  var getAnnotations = function (annotation, data){
```
jebeck: What you're passing to this function as `annotation` is never `>` or `<`, but rather I see `>/M0/T1` from a HI value. Consequently the annotations are not being added.

(You can produce a HI value with a bit of juice on a test strip!)

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,501 @@
...
+      var robj = {};
+      if (err) {
+          debug('Failure trying to read record', record);
+          debug(err);
+          return callback(err, null);
+      } else {
+          robj.timestamp = parseInt(r[4]);
+          robj.annotations = getAnnotations(r[3], data);
+          robj.units = r[2];
+          robj.glucose = parseInt(r[1]);
```
jebeck: I was actually surprised to see a HI value come through with a numeric value of 601; in my read of the spec I expected those values to be non-numeric. We *do* want HI values to have the value threshold + 1 and LO values to have the value threshold - 1. Do you have reason to believe that's how the device is actually representing them?

lib/drivers/bayerContourNext.js
```diff
@@ -0,0 +1,501 @@
...
+        cb(err, data);
+      });
+    },
+
+    processData: function (progress, data, cb) {
+        progress(0);
+        data.bg_data = processReadings(data.bgmReadings);
+        data.post_records = prepBGData(progress, data);
+        var ids = {};
+        for (var i = 0; i < data.post_records.length; ++i) {
```
jebeck: Why are you doing the dup check below? Have you been seeing duplicates? Is that due to something on the device?

(We do expect the possibility of duplicates by `time` because of the way the UTC timestamps are generated (can have dups if display time is changed), but they should be quite rare.)

jebeck: Within this loop can you also delete the `index` so it only appears in the `logIndices`? See comment [here](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/objectBuilder.js#L111). This is non-ideal, but it's only for this study, so...

lib/drivers/bayerContourNext.js
patovala: mmm... I quite don't understand this and maybe is because we don't see the whole picture. The threshold values are been gathered from the header l:164, normally we get here two values ie. 20, 600, and according to the manual, this values represents GLU LO = 20 and GLU HI = 600.  So please could you elaborate a little more specially about 'threshold + 1'.

jebeck: When someone tests and the value shows on the meter as LO or HI it represents a value that is either lower than the low threshold for recording numerical values on the meter (in this case, 20) or higher than the high threshold (in this case, 600). Our data model does not allow for non-numerical values in the `value` field, so we store threshold - 1 for LO values (e.g., 19 in this case) and threshold + 1 for HI values (e.g., 601 in this case), and then we add the annotation so that it's clear some data manipulation may have taken place.

The markers for being over or under threshold are the < and > markers. I would like to see code that recognizes reading with the < and > markers and *ensures* that the values stored in the `value` field are threshold - 1 in the LO (<) case and threshold + 1 in the HI (>) case. It appears that this is already what the meter is storing, at least for a HI value (which had a stored value of 601), but I found that surprising after looking at the spec, and I don't want to be unpleasantly surprised by LO values being potentially different. Does that help?

lib/drivers/bayerContourNext.js
patovala: We have not seen duplicates either, but remember we only have two devices in our hand so difficult to assert, so based on the other drivers implementations available we thought it was a good idea to check it.

Regarding the logIndices and the comment, if we remove the index here you won't get the logIndices filled either, in objectBuilder, or I'm missing something?

jebeck: After you've called the objectBuilder's `done()` method, the object builder doesn't touch the objects anymore, so no, you won't lose the logIndices if you wait until after `done()` is called to delete the `index`.

gniezen:
Hi @patovala and others from IOET! I'm going to close this pull request as I've created a minor fork in order to make some changes and will merge that as a separate pull request. Thanks for all your work on this. I think this driver looks good and we should get it into testing/QA.

PR #126, opened 2015-06-08
handle outdated uploader version error

jebeck:
The only TODO left here is the link to the page on tidepool.org or the knowledge base instructing how to update the uploader (there is a TODO comment in App.jsx about this). Since no one should ever see the error page where that link is displayed until we make the minimum version something higher than the default 0.99.0, this *can* be merged and published, and should be.

### DIFF
--- a/lib/components/UpdatePlease.jsx
+++ b/lib/components/UpdatePlease.jsx
@@ -22,11 +22,18 @@ var UpdatePlease = React.createClass({
     link: React.PropTypes.string.isRequired
   },
   render: function() {
+    var text = {
+      OUT_OF_DATE: 'Your uploader is out-of-date.',
+      TO_UPDATE: 'To update it, please follow ',
+      LINK_TEXT: 'these instructions',
+      TRY_AGAIN: 'Then try your upload again!'
+    };
+
     return (
       <div className="UpdatePlease">
-        <p>Your uploader is out-of-date.</p>
-        <p className='most-important'>Please follow the instructions <a href={this.props.link}>here</a> to update it.</p>
-        <p>Then try your upload again!</p>
+        <p>{text.OUT_OF_DATE}</p>
+        <p className='most-important'>{text.TO_UPDATE}<a href={this.props.link}>{text.LINK_TEXT}</a>.</p>
+        <p>{text.TRY_AGAIN}</p>
       </div>
     );
   }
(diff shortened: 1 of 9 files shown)


lib/components/UpdatePlease.jsx
```diff
@@ -0,0 +1,35 @@
...
+var React = require('react');
+
+var UpdatePlease = React.createClass({
+  propTypes: {
+    link: React.PropTypes.string.isRequired
+  },
+  render: function() {
+    return (
+      <div className="UpdatePlease">
+        <p>Your uploader is out-of-date.</p>
```
jh-bate: shall we use the have default text but allow it to be overridden? e.g. https://github.com/tidepool-org/chrome-uploader/blob/master/lib/components/Upload.jsx#L41

lib/state/appActions.js
```diff
@@ -62,6 +62,7 @@ appActions.errorStages = {
   STAGE_PICK_FILE : { code: 'E_SELECTING_FILE' , friendlyMessage: 'Error during file selection' },
   STAGE_READ_FILE : { code: 'E_READING_FILE' , friendlyMessage: 'Error trying to read file' },
   STAGE_UPLOAD : { code: 'E_UPLOADING' , friendlyMessage: 'Error uploading data' },
+  STAGE_METADATA_UPLOAD : { code: 'E_METADATA_UPLOAD' , friendlyMessage: 'Error uploading upload metadata'},
```
jh-bate: Do you think this should be should be worded a little differently? Just thinking the rest of the messages make sense in general but this might be a little more cryptic

jh-bate:
@jebeck I agree on your merge and deploy early stance.

A couple of questions / comments about user messages first.

lib/state/appActions.js
jebeck: This never gets exposed to the user because of the immediate redirect to the "Your uploader is out-of-date" page. It does look like this (if you put in a 10-second timeout on the redirect at least...):
screenshot 2015-06-08 16 10 14

I was thinking we might want to generalize the same error to other upload metadata validation problems.

jebeck:
I factored out the text for the first issue and commented on the second. LMK if it looks good to you now @jh-bate

jh-bate:
LGTM

PR #127, opened 2015-06-19
Fix license clause

darinkrauss:
@jh-bate


jh-bate:
LGTM just double checked based on this http://tidepool.org/2013/12/19/tidepool-license-update/

PR #106, opened 2015-04-07
"bootstrap" to UTC for all devices where possible

jebeck:
WIP. **Don't merge.** Waiting to be able to implement "bootstrapping" for Dexcom in order to merge this.

Closing PR #103 (for OmniPod only) because we'll do all devices at once from this PR instead.

PR #128, opened 2015-06-20
"bootstrap" to UTC for all devices where possible + data model updates

jebeck:
This will be the real PR for bootstrapping to UTC. Supercedes #106, which I will close shortly.


jebeck:
#131 will replace this.

PR #129, opened 2015-06-30
proposal for factoring out all CareLink CSV formatting details

jebeck:
Keen to get your opinion on this @jh-bate. Put it on a separate branch in case it's not the way we want to go forward.

What I'm envisioning here is that we could have several formatting specifications in the `lib/carelink/csv/` folder. Maybe another spec would be for an international version (I know someone Euro e-mailed about differences in their CSVs that went beyond just different column names, also things like `,` for the decimal separator, which we could address with a different parsing function, hence including the calls to the `parsing` utilities in the refactoring). Either some part of the UI (down the road) could be responsible for choosing the proper CSV format, or the code might be able to test what format of file it thinks it has.

LMK what you think...


lib/carelink/csv/minimed.com.js
```diff
@@ -0,0 +1,18 @@
...
+  pumpTypesToRead: {
+    BGCapturedOnPump: {
+      RAW_TYPE: 'BGCapturedOnPump',
+      RAW_VALUES: {
+        AMOUNT: parsing.asNumber([RV, 'AMOUNT']),
+        ACTION_REQUESTOR: parsing.extract([RV, 'ACTION_REQUESTOR'])
+      }
+    }
+  },
+  cgmTypesToRead: {
```
jebeck: Separating out pump types from CGM types is something that I had to do in the UTC bootstrapping work, so this is just looking forward.

lib/drivers/carelinkDriver.js
```diff
@@ -55,7 +57,7 @@ function initializeProcessors(opts) {
   // re-ordering.
   return [
     require('../carelink/settings.js')(opts),
-    require('../carelink/smbg.js')(opts),
+    require('../carelink/smbg.js')(_.assign(opts, csvFormat.pumpTypesToRead.BGCapturedOnPump)),
```
jebeck: Would handle this more elegantly in the final version I hope - this is just my hack for doing just one of the types this way.

jebeck:
Closing - although I will cherry-pick the commit with updates to tests. I was uncertain about whether the gymnastics involved in the architectural change for the driver was correct and @jh-bate also had doubts that this is the right time to try to make such changes. I will proceed with a smaller series of band-aids on the model of a3ebb0fa5174c867ae776dfbca2b25d4d2ac9b4f for settings.

PR #130, opened 2015-07-01
updates for new CareLink CSV format

jebeck:
Ready for review @jh-bate. Sorry for the messy diff, but the tests needed lots of updating... :/


jh-bate:
LGTM - nice use of the constants and also docs too.

To consider - if we had csv's of the old format that we had to reprocess then we would need to be versioning the driver. It doesn't matter at the moment but we should keep it in mind I feel

jebeck:
Mm, good point @jh-bate about re-processing older files, although I agree that at the moment we shouldn't do any gymnastics to make that possible. But I'll version bump to v0.150.0 as a small hack to make the division really clear & easier to spot, we can check out a < v0.150.0 older tag then if we ever need to re-process.

Will merge and publish to devel Chrome store now.

PR #131, opened 2015-07-18
bigger straps for those boots

jebeck:
**WIP**

This will probably replace #128 with expanded bootstrapping to UTC and the addition of a `conversionOffset` on all data.


jh-bate:
@jebeck LGTM

PR #133, opened 2015-07-24
delete Asante-related code

jebeck:
Still unclear whether we should be deleting Asante-specific thing(s) in the `lib/crc.js`. Given the comments about example code from J&J it may also apply to the OneTouch/Lifescan (== J&J) meters.

@gniezen please have a more extensive look at some point.

This PR is not a high priority but will be useful for clean-up purposes.


gniezen:
I can confirm that the things in `lib/crc.js` is being used in `oneTouchMiniDriver.js`. It seems that Asante and J&J used the same method to calculate the CRC, and the implemented code is based on example code from Asante and J&J. 

As such, I think we can merge this PR as is.

jebeck:
Thanks very much for remembering to look at this! It was starting to make me itchy having so many open things for my brain to juggle. Merging, and I'll update the bootstrapping branch myself from master.

PR #136, opened 2015-08-06
ContourNext USB implementation

gralmurdok:
We are publishing the bayer contour next USB implementation for your consideration.


gniezen:
When testing this on Windows 8.1, it seems that I cannot get data from the Bayer Contour Next (not the Bayer Contour Next USB) anymore - `receivePacket` times out. Is it happening for you as well?

gralmurdok:
We've done some tests on windows 8.1 but nothing happened.

gniezen:
Just to be sure, you've been able to upload from the Bayer Contour Next on Windows 8.1 as well, not only the Bayer Contour Next USB?

If it's not a Windows version issue, I wonder if it could be a Chrome version issue. What version of Chrome are you running on the Windows 8.1 machine?

gralmurdok:
We've done upload data from the BayerContourNext, we are using the chrome v 44.0.2403.155, no problems

gniezen:
I'm using Chrome v 44.0.2403.130, so it's probably not that then. It seems that the device disconnects and reconnects after the first packet is sent (based on the Windows sounds), which is why `receivePacket` then gets an empty packet until timeout.
capture

gralmurdok:
it probably could be a device issue, we've done more tests specially on Windows 8.1, and we can't get these message errors, we have a normal workflow (we are restored the 4 bytes, and we not have a disconnect/reconect problem)
imagen1

gniezen:
Thanks for the screenshot. I can confirm that I'm using exactly the same meter (Bayer7350) with the same firmware versions (01.14\01.03\04.18). Can you have a look in Device Manager to check what driver you're using on Windows 8.1?

gralmurdok:
yeah, we are using the default HID driver for Windows, no specyfic driver needed to work
imagen3
image2

gniezen:
I can confirm that I'm also using the default HID driver for Windows. There does seem to be a small difference in the version number, i.e., 6.3.9600.17041 instead of 6.3.9600.16384. Does your Windows 8.1 installation have all the newest updates?

Sorry for all the hassle, it's just that it's driving me crazy that we seem to have the exact same setup, but that it doesn't work this side. I've tried various fixes (e.g. changing the USB report header) but nothing seems to work for the Bayer Contour Next (although it works fine on OS X). The Bayer Contour Next USB works fine on both. I'm just worried that if I have this issue on Windows 8.1, someone else is going to have the same problem.

gralmurdok:
no problem :smile: , we are using two versions of Windows 8.1, one with all updates, and the other with a clean installation (no updates), in both we have no problems, and we are using a Windows 8 machine (no 8.1) all are working fine.
I'm also worried but we can't get the error, if you determine the origin of the issue please notify us to help fixing it

gniezen:
OK, as the Windows 8.1 error is an issue with the Bayer Contour Next, and not the Bayer Contour Next USB, it should not stop us from merging this pull request.

I've made some changes after testing the Bayer Contour Next USB:
- If there's a lot of data stored on the device, `abortTimer` times out before all the data is received. I've changed `abortTimer` so that it re-sets after every packet.
- I've added more error handling, so that if a device disconnects, we correctly handle the `chrome.runtime.lastError` that is thrown

I tried submitting these changes to your PR, but GitHub created a separate branch at:
https://github.com/tidepool-org/chrome-uploader/tree/pr/136

I'd rather merge this PR than create a new PR for the pr/136 branch, so I'm hoping you can merge that branch into your PR, check that you agree with my changes and verify that everything is still working?

gralmurdok:
yeah, we've done merging the pr/136 branch into this PR, we've done some tests in Win, Linux and Mac.
All working fine :smile:

gniezen:
We've recently merged a lot of changes into the uploader to make it UTC timezone-aware. As such I've merged this into a new PR (#144) with the required changes to make it timezone-aware. I've tested it against the development server on both Windows and Mac and everything looks good. @gralmurdok, please check that it works for you as well against the latest version of the server, and comment in the new PR if there are issues.

PR #138, opened 2015-08-12
Abbott Free Style Lite implementation

gralmurdok:
We are publishing the Abbott Free Style Lite driver implementation for your consideration.


lib/core/device.js
```diff
@@ -50,7 +52,9 @@ device._deviceComms = {
   'DexcomG4': serialDevice,
   'OneTouchMini': serialDevice,
   'AbbottFreeStyle': serialDevice,
+  'AsanteSNAP': serialDevice,
```
gniezen: We've removed the Asante drivers from the uploader, so it should not appear here.

gniezen:
I've made a number of changes and merged in other changes from `master`. Can you merge this branch and check that everything is still working on your side as well:
https://github.com/tidepool-org/chrome-uploader/tree/ioet-freeStyleLite

We can then finalise the review and merge this PR, instead of me having to create another PR.

gralmurdok:
ok, we have tested the device now and it works properly :smile:

gniezen:
I'm closing this PR, as I've now moved these changed to a new PR in #158. Thanks guys!

PR #139, opened 2015-08-13
add KB link

jebeck:
So the dogfooding revealed a bit of a brain fart. The current prod uploader didn't yet have the link to the KB article for how to update the uploader (because it didn't exist when we first added the "you need to update your uploader page") but the current prod uploader is the one we expect people to be get getting the "outdated uploader" error with when we start enforcing a minimum uploader version with the release of BtUTC.

So this is a quick fix to add the link that we will try to publish today to prod, and hopefully many people will get the update over the weekend!

### DIFF
--- a/lib/state/appState.js
+++ b/lib/state/appState.js
@@ -54,12 +54,12 @@ appState.getInitial = function() {
     //   name: 'OneTouch UltraMini',
     //   key: 'onetouchmini',
     //   source: {type: 'device', driverId: 'OneTouchMini'}
+    // },
+    // {
+    //   name: 'Bayer Contour Next',
+    //   key: 'bayercontournext',
+    //   source: {type: 'device', driverId: 'BayerContourNext'}
     // }
-    {
-      name: 'Bayer Contour Next',
-      key: 'bayercontournext',
-      source: {type: 'device', driverId: 'BayerContourNext'}
-    }
   ];
 
   if (config.CARELINK) {
--- a/lib/state/appState.js
+++ b/lib/state/appState.js
@@ -68,6 +68,7 @@ appState.getInitial = function() {
 
   return {
     dropMenu: false,
+    howToUpdateKBLink: 'https://tidepool-project.helpscoutdocs.com/article/6-how-to-install-or-upgrade-the-tidepool-uploader-gen',
     page: 'loading',
     user: null,
     targetId: null,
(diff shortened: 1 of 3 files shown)


lib/state/appState.js
```diff
@@ -68,6 +68,7 @@ appState.getInitial = function() {
 
   return {
     dropMenu: false,
+    howToUpdateKBLink: 'https://tidepool-project.helpscoutdocs.com/article/6-how-to-install-or-upgrade-the-tidepool-uploader-gen',
```
darinkrauss: Do we want to have a redirecting url here in case the end url changes?  It would be easier to change the redirecting url than the wait for the uploader to be distributed.

lib/state/appState.js
```diff
@@ -54,12 +54,12 @@ appState.getInitial = function() {
     //   name: 'OneTouch UltraMini',
     //   key: 'onetouchmini',
     //   source: {type: 'device', driverId: 'OneTouchMini'}
+    // },
+    // {
+    //   name: 'Bayer Contour Next',
+    //   key: 'bayercontournext',
+    //   source: {type: 'device', driverId: 'BayerContourNext'}
     // }
-    {
```
darinkrauss: Is there a reason this was commented out?

darinkrauss:
Two comments, but otherwise :+1:

lib/state/appState.js
jebeck: We haven't QA-ed it yet. It's not going to be tested until after the BtUTC release.

lib/state/appState.js
jebeck: If so, we need to set this up. Are you volunteering to spend time wrangling whatever needs to be done (probably mucking around with the WP site) to do that now? We need to get this out as fast as possible, no matter what kind of link is in it (as there is no link at all currently), and we have yet to decide what level of testing process we're comfortable with.

darinkrauss: :) Okay, how about pushing this out now and adding a new card to move to a redirecting url later?

jebeck: Sure, can you create that card @darinkrauss?

HowardLook:
I have also reviewed this. Looks good. Change is small enough and risk is low enough that I am comfortable pushing this directly to production.

lib/state/appState.js
darinkrauss: For future reference, the new Trello card:

https://trello.com/c/PHhrxxYn/97-2-uploader-create-redirect-url-endpoint-for-how-to-update-kb-article-in-uploader <https://trello.com/c/PHhrxxYn/97-2-uploader-create-redirect-url-endpoint-for-how-to-update-kb-article-in-uploader>

PR #140, opened 2015-08-15
use sundial's new getTimezones

jebeck:
Fixes an issue caused by the overly complicated gymnastics trying to fit our old timezone list into react-select's data model.

Also removed the 'x' for "clearability" in the dropdowns and ensured long timezone names will be elided in the selection box (both per @skrugman's guidance).

A remaining task is to make the dropdown wider so that long timezone names don't break onto two lines in the dropdown list - that turned out to be a much bigger task than anticipated due to some unfortunate magic numbers in the CSS.

### DIFF
--- a/lib/components/TimezoneSelection.jsx
+++ b/lib/components/TimezoneSelection.jsx
@@ -46,7 +46,6 @@ var TimezoneSelection = React.createClass({
         onChange={this.props.onTimezoneChange}
         options={opts}
         placeholder={'Type to search...'}
-        // if for whatever reason we don't have the label, just display the actual timezone name
         value={this.props.targetTimezone} />
     );
   },
--- a/lib/components/TimezoneSelection.jsx
+++ b/lib/components/TimezoneSelection.jsx
@@ -40,13 +40,6 @@ var TimezoneSelection = React.createClass({
       .concat(sortByOffset(timezones.hoisted))
       .concat(sortByOffset(timezones.theRest));
 
-    var targetTimezoneLabel = null;
-    if (this.props.targetTimezone !== null) {
-      targetTimezoneLabel = _.find(timezones, function(tz) {
-        return tz.value === self.props.targetTimezone;
-      });
-    }
-
     return (
       <Select clearable={false}
         name={'timezoneSelect'}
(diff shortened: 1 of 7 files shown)


lib/components/TimezoneSelection.jsx
```diff
@@ -24,26 +24,37 @@ var TimezoneSelection = React.createClass({
...
+    var opts = sortByOffset(timezones.bigFour)
+      .concat(sortByOffset(timezones.unitedStates))
+      .concat(sortByOffset(timezones.hoisted))
+      .concat(sortByOffset(timezones.theRest));
+
+    var targetTimezoneLabel = null;
+    if (this.props.targetTimezone !== null) {
+      targetTimezoneLabel = _.find(timezones, function(tz) {
+        return tz.value === self.props.targetTimezone;
+      });
```
darinkrauss: I may be missing something here, but the `_.find` returns the tz object, not the label, right?  Wouldn't you need to access the `label` property of the tz object returned from `_.find`?

lib/components/TimezoneSelection.jsx
```diff
@@ -24,26 +24,37 @@ var TimezoneSelection = React.createClass({
...
+    var timezones = sundial.getTimezones();
+    var opts = sortByOffset(timezones.bigFour)
+      .concat(sortByOffset(timezones.unitedStates))
+      .concat(sortByOffset(timezones.hoisted))
+      .concat(sortByOffset(timezones.theRest));
+
+    var targetTimezoneLabel = null;
+    if (this.props.targetTimezone !== null) {
+      targetTimezoneLabel = _.find(timezones, function(tz) {
+        return tz.value === self.props.targetTimezone;
```
darinkrauss: Do we need self here?  Won't this work just fine?  (I think we can get rid of self.)

darinkrauss:
@jebeck @jh-bate Looks good.  A couple quick comments, one which may just be cleanup. If you are good with those, then :ship:

lib/components/TimezoneSelection.jsx
jebeck: Well there were multiple things wrong with this (/o\), which means the fallback to `this.props.targetTimezone` was always in play, and as it turns out, that's all that's needed! Now that our label-value pairs are unique, react-select does the equivalent of the `_.find`, so I was able to just delete it.

lib/components/TimezoneSelection.jsx
jebeck: This got deleted, but for future reference inside lo-dash predicate functions you do need self.

PR #141, opened 2015-08-17
ensure we check for whether logged-in user has data storage

jebeck:
Bug fix for https://trello.com/c/16dIWMWm

Something I started (and just finished updating the tests for) when I couldn't sleep last Sat...

PR #142, opened 2015-08-18
slightly smaller straps for those boots

jebeck:
**WIP**

Introduces `clockDriftOffset`, separate from `conversionOffset`. The former is stored but not factored into the generation of the `time` field (to prevent uploading of duplicates). `conversionOffset` is reserved for changes larger than anything conceivable as a timezone offset change (i.e., device set to wrong date, month, year).


jebeck:
And...misfire. Branched off #141 instead of master. Closing this to create new.

PR #144, opened 2015-08-18
IOET Contour Next USB driver

gniezen:
This PR contains all the work from IOET’s original PR (#136) plus the necessary changes to make the driver BtUTC-aware.


Sjlmejia:
hi @gniezen we are reviewed the latest changes, and all worked fine.

lib/core/deviceInfo.js
```diff
@@ -37,5 +37,6 @@ module.exports = {
                                  'Blood glucose and ketone meter'),
   'InsuletOmniPod': infoBuilder('Insulet OmniPod', 'Upload .ibf file from PDM.'),
   'OneTouchUltra2': infoBuilder('OneTouch Ultra2', ''),
-  'BayerContourNext': infoBuilder('Bayer Contour Next', '')
+  'BayerContourNext': infoBuilder('Bayer Contour Next', 'Blood glucose meter'),
```
jebeck: I'll probably take out the 'Blood glucose meter'...we've only been adding a "subtitle" if there's something unusual about the device, like it reading ketones or whatever.

lib/drivers/bayerContourNext.js
```diff
@@ -34,6 +36,9 @@ module.exports = function (config) {
   var cfg = _.clone(config);
   var hidDevice = config.deviceComms;
   var HID_PACKET_SIZE = 64;
+  // With no date & time settings changes available,
+  // timezone is applied across-the-board
+  cfg.tzoUtil = new TZOUtil(cfg.timezone, new Date().toISOString(), []);
```
jebeck: Ah, this doesn't matter with across-the-board timezone application, but just as a reminder in bootstrap-able devices we should try to provide the most recent datum on device timestamp (if available) as 2nd arg of the TZOUtil constructor, not current datetime from `new Date()`. The edge case being if you upload just after Fall Back, say, but the most recent data is from just prior to Fall Back - we want to start with the pre-Fall Back offset (obtained by querying for the offset given the pre-Fall Back timestamp and the timezone), not the post-Fall Back offset.

lib/drivers/bayerContourNext.js
```diff
@@ -161,17 +167,33 @@ module.exports = function (config) {
...
         rawrecords: records,
-        lowT: parseInt(threshold[1]),
-        hiT: parseInt(threshold[2])
       };
 
+      if(threshold){
+        devInfo.lowT = parseInt(threshold[1]);
+        devInfo.hiT = parseInt(threshold[2]);
+      }else{
+        devInfo.uThres = true;
```
jebeck: I'd prefer a slightly more explicit variable name here - the full `unreportedThres` maybe? I remembered this from our conversation but coming new to the code I probably wouldn't know what it meant without looking for it elsewhere.

lib/drivers/bayerContourNext.js
```diff
@@ -161,17 +167,33 @@ module.exports = function (config) {
...
         serialNumber: sNum[1],
         nrecs: pString[6],
         rawrecords: records,
-        lowT: parseInt(threshold[1]),
-        hiT: parseInt(threshold[2])
       };
 
+      if(threshold){
+        devInfo.lowT = parseInt(threshold[1]);
+        devInfo.hiT = parseInt(threshold[2]);
```
jebeck: Ditto with these - I'd prefer the slightly more verbose `lowThres` and `hiThres` but don't feel super strongly about it.

jebeck:
Just a couple comments, but looks merge-able. Will merge soon regardless of response to comments since we want to have something ready to build and test.

PR #146, opened 2015-08-18
yet another BtUTC branch

jebeck:
Because after rubber-ducking with @darinkrauss I regretted the decision to revert the adjustments to the `computeMillisInCurrentDay` function in the CareLink simulator. Closing #143 in favor of this version.

ETA: We are putting off the decision as to whether `clockDriftOffset` (and also `conversionOffset`) will be optional or required attributes. The state of this branch is that `clockDriftOffset` will be left off Abbott PX events. (It will always be zero here anyway since the Abbott PX is a BGM that does not have any time change records, clock drift or otherwise.) `conversionOffset` is currently required and appears everywhere, although that could change in the future; if we decide to make it optional, we will likely stop marking it where it will be consistently 0 (again, on BGMs including the Abbott PX).


darinkrauss:
FYI - I looked at the non-test changes and it looks good to me. Of course, I don't yet know the larger picture of the uploader, but the changes do look reasonable.

jh-bate:
LGTM

PR #149, opened 2015-09-02
Bayer Contour USB

gniezen:
The Bayer Contour USB works with same driver as:

- Bayer Contour Next
- Bayer Contour Next USB

All that was needed was to add the right VID/PID combination.


jebeck:
LGTM, merging.

PR #153, opened 2015-09-08
Fix for Chrome 45 on Windows

gniezen:
This fixes the regression bug in Chrome 45 on Windows, where the Dexcom and Precision Xtra wasn't detected anymore.

### DIFF
--- a/lib/core/device.js
+++ b/lib/core/device.js
@@ -217,7 +217,11 @@ device.detectUsb = function(driverId, cb) {
     vendorId: driverManifest.vendorId,
     productId: driverManifest.productId
   };
-  chrome.usb.getDevices(identification, function(results) {
+  chrome.serial.getDevices(function(results) {
+    self.log('Connected device(s): ');
+    for(var i in results) {
+        self.log(results[i].path);
+    }
     var devices = _.map(results, function(result) {
       var retval = {
         driverId: driverId,
--- a/lib/drivers/abbottFreeStyle.js
+++ b/lib/drivers/abbottFreeStyle.js
@@ -373,8 +373,9 @@ module.exports = function (config) {
       // once we've sent the command, start listening for a response
       // but if we don't get one before the end of the timeout, give up
       // max timeout is 1000 * 32 bytes / 19200 baud / 10 bits/char == about 16 sec
-      listenForPacket(18000, commandpacket.parser, function(err, result) {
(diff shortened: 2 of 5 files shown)


lib/drivers/abbottFreeStyle.js
```diff
@@ -373,8 +373,9 @@ module.exports = function (config) {
       // once we've sent the command, start listening for a response
       // but if we don't get one before the end of the timeout, give up
       // max timeout is 1000 * 32 bytes / 19200 baud / 10 bits/char == about 16 sec
```
jebeck: Does your change match the comment here still? If it doesn't, can you update the comment?

lib/core/device.js
```diff
@@ -217,7 +217,11 @@ device.detectUsb = function(driverId, cb) {
     vendorId: driverManifest.vendorId,
     productId: driverManifest.productId
   };
-  chrome.usb.getDevices(identification, function(results) {
```
jebeck: So from what I understand of your explanation from elsewhere, there are disadvantages to doing the `chrome.serial.getDevices` when multiple devices might be connected. If needing to use it is a limitation that only affects Windows users so far, could we keep `chrome.usb.getDevices` for Mac users? Maybe we don't make that change today since we want to get this out ASAP, but what do you think?

lib/drivers/abbottFreeStyle.js
gniezen: Yes, I'd like to keep the info on what a max timeout will be, even if we only set it to 3s.

lib/core/device.js
gniezen: I think this is something that we need to look into for Chrome 46, as it looks like Google does want to improve `chrome.serial.getDevices`: https://code.google.com/p/chromium/issues/detail?id=522217

If they implement that, we can get the vendor ID and product ID using `chrome.serial.getDevices`, which will be the right way of doing things.

jebeck: Sounds good.

gniezen: Just revisited this and it looks like we'll finally be able to get vendor ID and product ID using `chrome.serial.getDevices` when Chrome 48 get released.

Current Chrome 46:
screen shot 2015-11-11 at 15 53 36

Chrome 48 (canary build):
screen shot 2015-11-11 at 15 53 22

This will resolve the `WARNING: More than one device found for ...` issue we've been having.

PR #154, opened 2015-09-10
add a default case to catch alarm codes not documented in the spec

jebeck:
In response to a bug report.

A small change to catch OmniPod alarm codes not documented in the spec and finish building proper objects with them. Since these aren't documented, leaving the `payload` (which usually contains more details) blank.

### DIFF
--- a/lib/drivers/insuletDriver.js
+++ b/lib/drivers/insuletDriver.js
@@ -896,6 +896,13 @@ module.exports = function (config) {
             .with_status(postsuspend)
             .done();
           break;
+        // for alarm codes not documented in the spec
+        default:
+          if (postalarm) {
+            postalarm = postalarm.with_alarmType('other')
+              .done();
+          }
+          break;
       }
       if (postalarm != null) {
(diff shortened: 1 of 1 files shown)


lib/drivers/insuletDriver.js
```diff
@@ -896,6 +896,13 @@ module.exports = function (config) {
             .with_status(postsuspend)
             .done();
           break;
+        // for alarm codes not documented in the spec
+        default:
+          if (postalarm) {
+            postalarm = postalarm.with_alarmType('other')
```
darinkrauss: Is it okay for an 'other' alarm type to NOT have a payload with alarmText, explanation, and stopsDelivery (as seen above on line 825)?  If so, LGTM!  :shipit:

jebeck: Yes, because these alarm codes aren't documented in the spec, we don't want to expose any info we shouldn't in the payload. They will just be "other" alarms until when/if we get more info in the spec we've been given to implement.

PR #155, opened 2015-09-11
increase timeout from 3s to 10s

jebeck:
What it says on the tin.


gniezen:
Reviewed. Looks fine to me!

PR #156, opened 2015-09-15
Hide unavailable devices

gniezen:
This PR will hide the Precision Xtra on Mac as its not supported at the moment. The other Abbott meters (FreeStyle Lite and FreeStyle Freedom Lite) can easily be included in this check after PR #138 has been merged.

New unit tests have been written that first sets the OS to either Windows or Mac and then checks which devices are available.


lib/state/appState.js
```diff
@@ -89,6 +89,17 @@ appState.getInitial = function() {
   };
 };
 
+appState.hideUnavailableDevices = function() {
+  var currentOS = this.app.device._os; // "mac", "win", "android", "cros", "linux", or "openbsd"
```
jebeck: I had to trace this through to figure out that `_os` gets added as a property of `device` in `lib/core/device.js`...it seemed an odd path, and I'd certainly be in favor of promoting the `_os` attribute up a level to `this.app._os` if you agree.

lib/state/appState.js
```diff
@@ -89,6 +89,17 @@ appState.getInitial = function() {
...
+
+  if(currentOS === 'mac') {
+    this.app.state.uploads = _.reject(this.app.state.uploads, function(d){
+      return d.key === 'precisionxtra'; // TODO: || d.key === 'abbottfreestylelite' || d.key === 'abbottfreestylefreedomlite'
```
jebeck: Do you think as we add more and more devices we should actually just add an attribute to each device's object in `this.app.state.uploads` specifying which OSes it's *not* compatible with and use that for the filtering?

jebeck:
A couple of questions re: code organization, but nothing that blocks merging, so merge at will @gniezen

lib/state/appState.js
gniezen: I agree with your reasoning, but I think the issue is that `chrome.runtime.getPlatformInfo` is asynchronous, so requires a callback. `App` is a React class, so I don't think we can use callbacks there (?), and `appState.js` doesn't use callbacks either. 

I can move it to `appActions.js`, but then we'll have to mock the `chrome` object, which would introduce a new dependency (`sinon-chrome`). More effort than it's worth?

lib/state/appState.js
gniezen: I think the list of unsupported devices per OS will be quite short. So far it's just the Abbott devices that turned out to have a very different Mac implementation, so I'm hoping that we will be able to supported both Mac and Windows for each new device that we roll out.

If this turns out not to be the case, I agree that it's better to store it as an attribute in `uploads`.

lib/state/appState.js
jebeck: Yeah, sounds like more effort than it's worth. Maybe just add a comment so that other people reading the code from a fresh perspective (like me yesterday) have a hint as to where `this.app.device._os` is being set?

lib/state/appState.js
jebeck: Mkay, fine for now, we'll reassess later if we have a lot of devices like this.

lib/state/appState.js
gniezen: On second thought, I see we're already using `typeof chrome !== 'undefined'` in `appActions.js` to ignore chrome-specific code in the test environment. I'm going to add a `getOS()` function to `appActions.js`, which will create `this.app._os` on load.

gniezen:
@jebeck: Will you just double-check my latest changes based on your feedback before I merge?

jebeck:
OK, had another look. My only question now is about how we _also_ still have `this.app.device._os`...could we consolidate and have one source of truth for platform info? Or is that more effort than it's worth?

PR #158, opened 2015-09-17
IOET's Freestyle Lite & Freestyle Freedom Lite driver

gniezen:
I'm creating a new PR based off #138 to be able to integrate changes faster.


lib/drivers/abbottFreeStyleLite.js
```diff
@@ -0,0 +1,538 @@
...
+        var testType = match[3];
+
+        if(testType[3] === '1') {
+          debug('Marking as control test');
+          entry.control = true;
+        }else{
+          entry.control = false;
+        }
+
+        //this works as a reference, from the standard HI/LO values
```
jebeck: What do you mean "as a reference"? Just not sure I understand this comment. I do see from the checklist that reading out-of-range thresholds from device is not implemented - presumably not in spec?

jebeck:
LGTM, just one question about a comment I didn't quite understand @gniezen

lib/drivers/abbottFreeStyleLite.js
gniezen: This was a comment from IOET. I've just checked the spec again - it looks like if the HI/LO flag is set, the actual value sent will be the threshold, e.g. for a reading of 567 (with a range of 20-500), the returned value will be 500 with the HI flag set.  I think they assumed that it will always be 20-500, but I don't think this will always be the case. Will look into it.

gniezen: Sorry, just realised I've been looking at the wrong spec just now. According to the *right* spec, the threshold is hardcoded as 20 for LO and 500 for HI (for all FreeStyle Family meters), so the code is correct. I'll change the comment for readability.

PR #160, opened 2015-09-25
move logic for creating a bucket for CGM data inside check for uploads we're reading

jebeck:
What it says on the tin.

In response to a support issue. The issue here is that if there are overlapping (in time) uploads from the same model of device, we only take the very last upload. So in this case where there was CGM data in an early upload but not in the most recent, we ended up creating a bucket for CGM data that did not exist in the most recent upload, and we expected it to be non-empty.


gniezen:
LGTM.

PR #161, opened 2015-10-02
Bypassing the detect function for Dexcom to fix serial port issue on El Capitan

gniezen:
On El Capitan, something goes wrong during serial port disconnect and reconnect. This is an issue because between the `detect` and `process` functions we do a disconnect/reconnect.

Bypassing the default `detect` function should not cause any issues, as it does a *connect->getConfigInfo->disconnect->cleanup*, which is repeated directly afterwards in the `process` function with a *connect->getConfigInfo->fetchData->processData->uploadData->disconnect*. This can be verified in `driverManager.js` from line 72 onwards.

Known bug(s):
- If the Dexcom is not unplugged and reconnected after upload, it won't connect again and may need to be power-cycled

### DIFF
--- a/lib/drivers/dexcomDriver.js
+++ b/lib/drivers/dexcomDriver.js
@@ -718,7 +718,7 @@ module.exports = function (config) {
     for (var i = 0; i < pagedata.length; ++i) {
       var page = pagedata[i].parsed_payload;
       for (var j = 0; j < page.data.length; ++j) {
-        var reading = _.pick(page.data[j], 'displaySeconds', 'displayDate', 'internalTime', 
+        var reading = _.pick(page.data[j], 'displaySeconds', 'displayDate', 'internalTime',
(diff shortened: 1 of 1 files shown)


lib/drivers/dexcomDriver.js
```diff
@@ -857,9 +857,11 @@ module.exports = function (config) {
 
 
   return {
-    // using the default detect for this driver
-    // detect: function(cb) {
-    // },
+    //bypassing the detect function for now
+    detect: function (obj, cb) {
+      debug('Dexcom Detect!');
```
jebeck: If we're still doing this bypass even after a more thorough dig into the El Capitan changes, maybe it'd be better to have the debug message 'Bypassing Dexcom detect' or something, seems a little odd to me to have added a debug message for what's basically a no-op ;) But I'm gonna merge so Howard can test the build.

PR #164, opened 2015-10-11
Add new production endpoint; allow access to tidepool.org; required for new cluster

darinkrauss:
@jh-bate @jebeck The permissions change took quite some time to track down.

### DIFF
--- a/main.js
+++ b/main.js
@@ -50,6 +50,14 @@ var contextMenus = [
   },
   {
     type: 'radio',
+    id: 'Prd',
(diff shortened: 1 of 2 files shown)


jh-bate:
so now in the dropdown we have existing prod and new prod?

jh-bate:
LGTM 

fyi I assume for the time being it will still default to old prod is you don't select and environment??

darinkrauss:
Yes, "Production" is still the one selected by default.

PR #151, opened 2015-09-02
ioet Contour feature

gralmurdok:
We are publishing the bayer contour driver serial for your consideration, this driver can be used to gather data with the bayer USB data cable, at the moment we are tested it with the Bayer Contour Next EZ

PR #168, opened 2015-10-22
IOET's Bayer Contour driver (UPLOAD-24)

gniezen:
This is an updated version of PR #151. It looks like the Bayer USB cable (used with the Bayer Contour, Contour Next EZ and Contour Link) only comes with drivers for Windows.

After merging with the latest master branch, this PR does download data from the Contour device, but breaks halfway during processing. Probably as it was only tested with the Contour Next EZ (see comment https://github.com/tidepool-org/chrome-uploader/pull/151#issue-104520945). 

Seems to work fine with Contour Next EZ, but driver should be updated to accurately reflect various device names.


gniezen:
@pazaan Wow, I can't believe it. After 4.5 years, this PR is finally ready for review. :sweat_smile: 

Details: [UPLOAD-24]

PR #169, opened 2015-10-23
correct building of date & time changes, albeit separately

jebeck:
@gniezen you'll need to start here.

There is a lot to be done before we even get to a stub simulator... This corrects the building of `timeChange` objects, but unfortunately when you change the date & time, we can only get those as two separate events, which may or may not work for bootstrapping - I'm not sure yet.

Next steps are to apply bootstrapping throughout the driver. The current code does not quite follow the strategy discussed in [the BtUTC doc](https://github.com/tidepool-org/chrome-uploader/blob/master/docs/BootstrappingToUTC.md) and is incorrect and/or misleading in some places. We can discuss today.

### DIFF
--- a/lib/drivers/tandemTslimDriver.js
+++ b/lib/drivers/tandemTslimDriver.js
@@ -1042,49 +1042,50 @@
   };
 
   var buildTimeChangeRecords = function (data, postrecords) {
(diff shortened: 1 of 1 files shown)


jebeck:
Rebasing a new branch on top of tandem-mvp, closing this.

PR #171, opened 2015-10-23
Tandem date & time changes, stub simulator

jebeck:
Here you go @gniezen 

As I was working on this earlier, I found it useful to comment out the calls to all `buildFooRecords()` functions except for the date & time changes...

Replaces #169

PR #172, opened 2015-10-23
Tandem date time changes

jebeck:
Here you go @gniezen

As I was working on this earlier, I found it useful to comment out the calls to all `buildFooRecords()` functions except for the date & time changes...

Replaces #169 #171

### DIFF
--- a/lib/drivers/tandemTslimDriver.js
+++ b/lib/drivers/tandemTslimDriver.js
@@ -1291,8 +1291,6 @@ module.exports = function (config) {
       if (bolus.total_bolus_size !== undefined && bolus.target_bg !== 0 && bolus.target_bg) {
         var wizard_record = cfg.builder.makeWizard()
           .with_deviceTime(bolus.wizardDeviceTime)
-          .with_timezoneOffset(record.timezoneOffset)
-          .with_conversionOffset(record.conversionOffset)
           .with_recommended({
             carb: bolus.food_bolus_size,
             correction: bolus.correction_bolus_size,
@@ -1310,7 +1308,9 @@ module.exports = function (config) {
           })
           .with_bolus(record)
           .with_units('mg/dL')
-          .done();
+          .set('index', bolus.index);
+        cfg.tzoUtil.fillInUTCInfo(wizard_record, bolus.jsDate);
+        wizard_record = wizard_record.done();
         records.push(wizard_record);
       }
     }
--- a/lib/drivers/tandemTslimDriver.js
+++ b/lib/drivers/tandemTslimDriver.js
@@ -1145,15 +1145,47 @@ module.exports = function (config) {
       var tc = timeChangeLogs[i];
       console.log(tc);
       var tc_base = sundial.floor(tc.rawTimestamp, 'day').valueOf();
-      var timechange = cfg.builder.makeDeviceEventTimeChange()
-        .with_change({
-          from: sundial.formatDeviceTime(tc_base + tc.time_prior),
-          to: sundial.formatDeviceTime(tc_base + tc.time_after),
-          agent: 'manual'
-        })
-        .with_deviceTime(tc.deviceTime)
-        .set('index', tc.index);
-      postrecords.push(timechange);
+
+      // Look for a corresponding date change event
+      var found = false;
+      var index = data.log_records.indexOf(tc);
+      var start = index < 5 ? 0 : index - 5; // look at 5 entries before
+      var end = (index + 5) < (data.log_records.length - 1) ? index + 5 : data.log_records.length - 1; // and 5 entries after
+      for(var k = start; k <= end; k++) {
+        var event = data.log_records[k];
+        if(event.header_id == PUMP_LOG_RECORDS.LID_DATE_CHANGED.value) {
+          console.log('found a corresponding date change:', event);
+          found = true;
+          var datetimechange = cfg.builder.makeDeviceEventTimeChange()
+            .with_change({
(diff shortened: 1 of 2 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zXgrhdC4&#x2F;17-1-tandem-clean-up-old-across-the-board-timezone-non-btutc-timestamp-generation"> &lt;1&gt; [tandem] clean up old across-the-board-timezone non-BtUTC timestamp generation</a>

lib/drivers/tandemTslimDriver.js
```diff
@@ -1133,50 +1131,87 @@ module.exports = function (config) {
...
+      [PUMP_LOG_RECORDS.LID_DATE_CHANGED],
+      data.log_records
+    );
+
+    for (var i = 0; i < timeChangeLogs.length; ++i) {
+      var tc = timeChangeLogs[i];
+      console.log(tc);
+      var tc_base = sundial.floor(tc.rawTimestamp, 'day').valueOf();
+
+      // Look for a corresponding date change event
```
jebeck: So unfortunately in my data, the most recent instance of this pairing up of time changes and date changes actually has *two* time changes paired with a single date change (probably fiddled with the time multiple times in the UI or something), so we're gonna have to figure out a more general method of grouping date & time changes rather than just assuming they occur in _pairs_. Going to give it a try myself now.

lib/drivers/tandemTslimDriver.js
```diff
@@ -1133,50 +1131,89 @@ module.exports = function (config) {
...
+      data.log_records
+    );
+
+    for (var i = 0; i < timeChangeLogs.length; ++i) {
+      var tc = timeChangeLogs[i];
+      var tc_base = sundial.floor(tc.rawTimestamp, 'day').valueOf();
+
+      // look for a corresponding date change event
+      var found = false;
+      var index = tc.index;
```
jebeck: You were using the array index rather than the (persistent) Tandem log index for all this @gniezen...I'm curious why? (Also, I changed it, but we can discuss if there's something I'm missing.)

lib/drivers/tandemTslimDriver.js
```diff
@@ -1133,50 +1131,89 @@ module.exports = function (config) {
...
+      var lastIndex = data.log_records[data.log_records.length - 1].index;
+      var SEARCH_THRESHOLD = 5;
```
jebeck: I factored out a `SEARCH_THRESHOLD` constant just in case we need to change it later.

lib/drivers/tandemTslimDriver.js
```diff
@@ -1133,50 +1131,89 @@ module.exports = function (config) {
...
+      var start = index < SEARCH_THRESHOLD ? 0 : index - SEARCH_THRESHOLD; // look at x entries before
+      var end = (index + SEARCH_THRESHOLD) < lastIndex ?
+        index + SEARCH_THRESHOLD : lastIndex; // and up to x entries after
+      for (var k = start; k <= end; k++) {
+        var event = _.find(dateChangeLogs, {index: k});
```
jebeck: So it occurred to me that this was really all that was needed to have the `_.without` work properly. However, it's not the most efficient approach (but since `dateChangeLogs` is never likely to be a _huge_ array I don't see much of a problem).

With the cluster of 3 events in my data, I ended up just pairing two and having the additional time change just a separate change (a clock drift offset). I think that's actually...fine, but we can discuss. (There did turn out to be additional complications re: having the date & time changes separate.)

lib/drivers/tandemTslimDriver.js
```diff
@@ -1529,6 +1557,18 @@ module.exports = function (config) {
...
+        where a pump shut-down appears in records to be processed
+        we only attempt to process and upload the data following
+        the most recent device shut-down.
+      */
+      data.log_records = _.takeRightWhile(data.log_records, function(rec) {
+        if (rec.change_type && rec.change_type === 'pump_shut_down') {
+          return false;
+        }
+        return true;
+      });
```
jebeck: Here's how I handled the device shut down issue. Not too hard, thankfully :)

jebeck:
OK, I made a bunch of comments mostly for us to discuss @gniezen 

Merging now, then will do a new branch from tandem-mvp for start of basals/simulator work.

lib/drivers/tandemTslimDriver.js
gniezen: Because there may be gaps in the Tandem log index, as we don't process _all_ events. I just wanted to make sure I was looking at exactly 5 events before and after the current one _that we processed_. As we're now comparing to the date events in `dateChangeLogs` instead of `data.log_records`, we'll have to use the Tandem log index (and I think this is ok).

I suppose the if-statement in line 1163 (`if (event.header_id == PUMP_LOG_RECORDS.LID_DATE_CHANGED.value)`) can be removed as it should always be true now?

jebeck: But since we fetch records in batches, couldn't we end up with our array of records not sorted by Tandem log index, if we change something at a later date re: fetching, for example?

And yes, good catch re: line 1163. Feel free to remove for clean-up whenever. Either remove entirely, or replace with an `else` (paired with the `if (!event)`)?

gniezen: Done in commit https://github.com/tidepool-org/chrome-uploader/commit/491f497334856bdbb877d7ba48aa35abb29144a1

PR #177, opened 2015-10-29
Increments progress bar after every 1000 records received from Tandem pump

gniezen:
Does what it says on the tin.


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ZEU2QHCw&#x2F;42-tandem-progress-bar"> [tandem] progress bar</a>

jebeck:
This seemed to increment within a quite small range - like it jumped to 47% then increment slowly to 48%, then 49%, then bigger jumps when done fetching. So we might still want to fiddle later, but better than it was! Merging into basal-simulator now to try to reduce branches.

PR #179, opened 2015-10-29
Daylight Savings Message that shows once per user on upload screen

GordyD:


lib/components/DaylightSavingMessage.jsx
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Determine whether the message should be displayed to the user or not
+   * 
+   * @return {boolean}
+   */
+  shouldDisplayOverlay: function() {
+    // for now show if the user has not previously acknowledged this warning
+    // in future though, we could make this show if
+    //  - the current date is within X days of DST starting/ending
+    //  - and acknowledged is false or not within X days
```
jebeck: Yeah we're going to need to adjust this for this deploy I think. We're planning to publish today, and we don't want people who update right away to acknowledge it on Saturday and then not see the warning after the switch on Sunday! Let's add `if (new Date().valueOf() >= [UTC value of switchover in selected timezone])`. Need to think about how to find that 'UTC value of switchover in selected timezone' - might not be an appropriate function in sundial.

lib/state/appActions.js
```diff
@@ -257,6 +258,13 @@ appActions.login = function(credentials, options, cb) {
       return cb(err);
     }
     self._logMetric(self.trackedState.LOGIN_SUCCESS);
+    
+    var acks = localStore.getItem('acknowledgments');
```
jebeck: I don't understand why you're loading these before the `appState.getInitial()`. Can you explain why retrieving the acknowledgements is different from how we retrieve `devices`?

lib/state/appActions.js
```diff
@@ -817,4 +825,24 @@ appActions.reset = function(uploadIndex) {
...
+  if (!update) {
+    update = {};
+  }
+  
+
+  if (!update[key]) {
+    update[key] = {};
+  }
+
+  update[key][this.app.state.targetId] = Date.now();
```
jebeck: Does `Date.now()` stringify to an ISO string automatically? If not, I'd rather explicitly store an ISO string or a `valueOf()` integer.

GordyD: The Date.now() method returns the number of milliseconds elapsed since 1 January 1970 00:00:00 UTC.

lib/state/appActions.js
GordyD: I `devices` see is retrieved in `_afterLoginPage`. I could move this logic into that function as it is over ever called after this point and before when we render the page.

jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;O96HU1ws&#x2F;47-uploader-ui-updates-to-display-dst-device-time-update-suggestion-and-store-acknowledgement"> [uploader] UI updates to display DST device time update suggestion and store acknowledgement</a>

lib/state/appActions.js
GordyD: As for explaining why. I am using the `initialState.acknowledgments` to be the defaultValue if `localStorage.acknowledgments` is not set. Because it is a nested object, with possibly more objects inside it I didn't want to redefine the initial empty structure of this object, so I am re-using it.

lib/components/DaylightSavingMessage.jsx
```diff
@@ -0,0 +1,104 @@
...
+*/
+
+var React = require('react');
+var sundial = require('sundial');
+
+var DaylightSavingMessage = React.createClass({
+  propTypes: {
+    onAcknowledge: React.PropTypes.func.isRequired,
+    acknowledged: React.PropTypes.number,
+    timezone: React.PropTypes.any
```
jebeck: Why PropType `any`?? Should be string, no?

lib/state/appActions.js
jebeck: OK, I get that - re: the nested object and re-use. If you could move to `_afterLoginPage` along with `devices` that'd be great, just to keep things together.

lib/state/appActions.js
jebeck: Awesome, fine.

lib/state/appActions.js
```diff
@@ -817,4 +822,43 @@ appActions.reset = function(uploadIndex) {
...
+ * @param {String} isoString
+ * @return {boolean}
+ */
+appActions.shouldDisplayDaylightSavingsMessage =  function(isoString) {
+  // for now show if the user has not previously acknowledged this warning
+  // in future though, we could make this show if
+  //  - the current date is within X days of DST starting/ending
+  //  - and acknowledged is false or not within X days
+  var state = this.app.state;
+  var localTime = sundial.formatInTimezone(isoString, state.targetTimezone, 'YYYY-MM-DD[T]HH:mm:ss');
```
jebeck: All the gymnastics with `localTime` and then `localTimeTimestamp` are unnecessary. The `localTimeTimestamp` is just `Date.parse(isoString)`. (You're converting from UTC to local and then back again with the current logic.)

test/ui/testAppActions.js
```diff
@@ -942,4 +942,51 @@ describe('appActions', function() {
 
   });
 
+  describe('acknowledge', function() {
+    it('should record acknoledgment for user for specific key', function() {
```
jebeck: Sp: acknowledgment.

test/ui/testAppActions.js
```diff
@@ -942,4 +942,51 @@ describe('appActions', function() {
...
+      var userid = 'foobar';
+      var key = 'daylightSavings';
+      app.state.targetId = userid;
+      appActions.acknowledge(key);
+      expect(app.state.acknowledgments[key][userid]).to.be.ok;
+    });
+  });
+
+  describe('shouldDisplayDaylightSavingsMessage', function() {
+    it('should return false when currentTime is less than time chnage time and user not seen message', function() {
```
jebeck: Sp: change.

test/ui/testAppActions.js
```diff
@@ -942,4 +942,51 @@ describe('appActions', function() {
...
+      var currentTimeIso = '2015-10-30T14:34:00';
```
jebeck: This should be a real ISO String - '2015-10-30T14:34:00.000Z';

test/ui/testAppActions.js
```diff
@@ -942,4 +942,51 @@ describe('appActions', function() {
...
+      app.state.acknowledgments.daylightSavings = {};
+      expect(appActions.shouldDisplayDaylightSavingsMessage(currentTimeIso)).to.be.false;
+    });
+
+    it('should return false when currentTime is less than time change time and user has seen message', function() {
+      var currentTimeIso = '2015-10-01T14:34:00';
+      app.state.targetTimezone = 'US/Pacific';
+      app.state.targetId = 'foobar';
+      app.state.acknowledgments = {};
+      app.state.acknowledgments.daylightSavings = { foobar: 134577844};
```
jebeck: So the stored timestamp is a Unix timestamp in seconds, not the milliseconds. Which are we storing? Please store milliseconds and update the test to be consistent - I get we aren't doing anything with this yet, but it's confusing...

jebeck:
Merging and building for devel Chrome store! Thanks for hard work @GordyD :D

PR #180, opened 2015-10-30
Invert structure of acknowledgements so that we store a hash of ackno…

GordyD:
…wledgements for user ids - enables us to show warning message for every account someone uploads for

This branch shows a warning any time someone tries to upload for someone new whom they are allowed to upload for.

N.B The warning message currently blocks all controls. We may want to enable users to still use the user selector for choosing different person so they don't have to dismiss the message if they accidentally clicked on the wrong user. Not sure if this is thinking it over too much or not though....


jebeck:
I think this is the route we want to go so merging into other branch to look at the diff in one place.

PR #182, opened 2015-10-31
Adds new day event to scheduled basals

gniezen:
Adds a new day event to scheduled basals, so that we can break up flat-rate basals.

### DIFF
--- a/lib/drivers/tandemTslimDriver.js
+++ b/lib/drivers/tandemTslimDriver.js
@@ -1335,50 +1335,61 @@ module.exports = function (config) {
     for (var b = 0; b < basalRecords.length; ++b) {
       var event = basalRecords[b];
 
-      if((event.header_id == PUMP_LOG_RECORDS.LID_NEW_DAY.value) && (event.commanded_basal_rate !== undefined)) {
-        // new day event; breaks up flat-rate basals
-        postbasal = cfg.builder.makeScheduledBasal()
-          .with_deviceTime(event.deviceTime)
-          .with_rate(event.commanded_basal_rate)
-          .set('index', event.index);
-        cfg.tzoUtil.fillInUTCInfo(postbasal, event.jsDate);
-      }else{
-        switch (event.change_type) {
-          case 'only_timed_segment':
-          case 'new_profile':
-          case 'temp_rate_end':
-          case 'pump_resumed':
-            // TODO: remove scary logging when confident that this situation (command != base)
-            // either does not arise or is handled
-            if (event.command_basal_rate !== event.base_basal_rate) {
-              console.log('COMMAND BASAL NOT EQUAL TO BASE!!');
-            }
-            postbasal = cfg.builder.makeScheduledBasal()
-              .with_scheduleName(_.find(data.profiles,{idp:event.idp}).name)
-              .with_deviceTime(event.deviceTime)
-              .with_rate(event.command_basal_rate)
-              .set('index', event.index);
-            cfg.tzoUtil.fillInUTCInfo(postbasal, event.jsDate);
-            break;
-          case 'temp_rate_start':
-            postbasal = cfg.builder.makeTempBasal()
-              .with_deviceTime(event.deviceTime)
-              .with_rate(event.command_basal_rate)
-              .set('index', event.index);
-            cfg.tzoUtil.fillInUTCInfo(postbasal, event.jsDate);
-            break;
-          case 'pump_suspended':
-            postbasal = cfg.builder.makeSuspendBasal()
-              .with_deviceTime(event.deviceTime)
-              .set('index', event.index);
-            cfg.tzoUtil.fillInUTCInfo(postbasal, event.jsDate);
-            break;
-        }
-        // there are some unhandled change types, so it's sometimes null
-        // TODO: be certain about this ^
+      switch (event.change_type) {
+        case 'only_timed_segment':
+        case 'new_profile':
+        case 'temp_rate_end':
+        case 'pump_resumed':
+          // TODO: remove scary logging when confident that this situation (command != base)
+          // either does not arise or is handled
+          if (event.command_basal_rate !== event.base_basal_rate) {
+            console.log('COMMAND BASAL NOT EQUAL TO BASE!!');
+          }
+          postbasal = cfg.builder.makeScheduledBasal()
+            .with_scheduleName(_.find(data.profiles,{idp:event.idp}).name)
+            .with_deviceTime(event.deviceTime)
+            .with_rate(event.command_basal_rate)
+            .set('index', event.index);
+          cfg.tzoUtil.fillInUTCInfo(postbasal, event.jsDate);
+          break;
+        case 'temp_rate_start':
+          postbasal = cfg.builder.makeTempBasal()
+            .with_deviceTime(event.deviceTime)
+            .with_rate(event.command_basal_rate)
+            .set('index', event.index);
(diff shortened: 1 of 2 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yxoaAoL4&#x2F;46-tandem-simulate-new-day-events-to-handle-flat-rate-basals"> [tandem] simulate new-day events to handle flat-rate basals</a>

lib/drivers/tandemTslimDriver.js
```diff
@@ -1328,45 +1328,55 @@ module.exports = function (config) {
 
   var buildBasalRecords = function (data, records) {
     var basalRecords = filterLogEntries(
-      [PUMP_LOG_RECORDS.LID_BASAL_RATE_CHANGE],
+      [PUMP_LOG_RECORDS.LID_BASAL_RATE_CHANGE,PUMP_LOG_RECORDS.LID_NEW_DAY],
```
jebeck: Not a fan of this approach - will explain why tonight. We should have a separate function `buildNewDayRecords` and simulate new day records separately.

lib/tandem/tandemSimulator.js
```diff
@@ -47,6 +47,7 @@ exports.make = function(config){
 
   var currBasal = null;
   var currTimestamp = null;
+  var currStatus = null;
```
jebeck: You're not actually using this... but you should. Eventually we will be reading suspends and resumes, and should track `currStatus` in those terms. (`null` for no info, `suspended` or `resumed`, at least to start)

lib/tandem/tandemSimulator.js
```diff
@@ -119,6 +120,27 @@ exports.make = function(config){
         events.push(currBasal);
       }
     },
+    newDay: function(event) {
+      ensureTimestamp(event);
+      if((currBasal !== null) && (currBasal.deliveryType !== 'suspend')) {
```
jebeck: Translate this into a proper `currStatus` (e.g., `suspended`) and set the `currStatus` with a fn like the `setCurrBasal` function.

lib/tandem/tandemSimulator.js
```diff
@@ -119,6 +120,27 @@ exports.make = function(config){
...
+
+        if (currBasal.time !== event.time) {
+          if (!currBasal.isAssigned('duration')) {
+            currBasal.with_duration(Date.parse(event.time) - Date.parse(currBasal.time));
+          }
+          currBasal = currBasal.done();
+
+          event.set('type', 'basal');
+          event.previous = _.omit(currBasal, 'previous');
+          event.with_scheduleName(currBasal.scheduleName);
```
jebeck: Are you trying to rely on the fact that `currBasal.scheduleName` will almost always be `null` or `undefined` here? Schedule name on new day is just like schedule name on basals - can only be filled in on very final basal, as that's the only time we can guarantee it will be accurate (before we're reading and tracking a full settings history). If you're counting on the fallback to `null` or `undefined` here I'd prefer to use a different strategy - this is hard to read and understand.

lib/tandem/tandemSimulator.js
gniezen: Done (for now) using suspend/resume basals. Will be changed to use suspend/resume events when we're reading them in next sprint.

lib/drivers/tandemTslimDriver.js
gniezen: Done.

lib/tandem/tandemSimulator.js
```diff
@@ -85,6 +90,14 @@ exports.make = function(config){
           events.push(currBasal);
         }
       }
+
+      if(event.deliveryType === 'suspend') {
+        setCurrStatus('suspended');
+      }else if(event.deliveryType === 'resume') {
```
jebeck: Can we try to stick to a consistent style on if/elses? The rest of the codebase has:
```
if (blah) {
}
else {
}
```
With separate blocks and space after the `if` and `else` before the parens.

lib/drivers/tandemTslimDriver.js
jebeck: Wait, why are we still including `PUMP_LOG_RECORDS.LID_NEW_DAY` in the filter since they're now being processed in `buildNewDayRecords`?

lib/drivers/tandemTslimDriver.js
```diff
@@ -1374,6 +1375,28 @@ module.exports = function (config) {
...
+      var utcTime = sundial.applyTimezone(event.deviceTime, cfg.timezone).toISOString();
+      if(event.commanded_basal_rate !== undefined) {
+        // new day event; breaks up flat-rate basals
+        var postrecord = cfg.builder.makeScheduledBasal()
+            .with_deviceTime(event.deviceTime)
+            .with_rate(event.commanded_basal_rate)
+            .set('index', event.index)
+            .with_conversionOffset(0)
+            .with_timezoneOffset(0)
+            .with_time(utcTime); //new day events should always occur at midnight, so don't shift with time changes
```
jebeck: New day events are midnight local time, _not_ midnight UTC. Use bootstrapping via the tzoUtil as per usual. If that causes some sort of issue with the flat-rate basal data you have, let's troubleshoot together.

lib/drivers/tandemTslimDriver.js
```diff
@@ -1345,10 +1346,10 @@ module.exports = function (config) {
             console.log('COMMAND BASAL NOT EQUAL TO BASE!!');
           }
           postbasal = cfg.builder.makeScheduledBasal()
-            .with_scheduleName(_.find(data.profiles,{idp:event.idp}).name)
             .with_deviceTime(event.deviceTime)
             .with_rate(event.command_basal_rate)
-            .set('index', event.index);
+            .set('index', event.index)
+            .set('idp', event.idp);
```
jebeck: We don't want to upload the profile index `idp` on all basals. We should either put it in the payload or clean it up in the simulator prior to uploading.

lib/drivers/tandemTslimDriver.js
```diff
@@ -1374,6 +1375,26 @@ module.exports = function (config) {
     return records;
   };
 
+  var buildNewDayRecords = function(data, records) {
+    var newDayRecords = filterLogEntries([PUMP_LOG_RECORDS.LID_NEW_DAY], data.log_records);
+
+    for (var b = 0; b < newDayRecords.length; ++b) {
+      var event = newDayRecords[b];
+      var rate = event.commanded_basal_rate;
+      if ((rate !== undefined) && (rate > 0)) {
```
jebeck: Possibly an even safer check might be `rate != null` since this covers both `null` and `undefined`.

lib/tandem/tandemSimulator.js
```diff
@@ -85,6 +90,16 @@ exports.make = function(config){
           events.push(currBasal);
         }
       }
+
+      if (event.deliveryType === 'suspend') {
+        setCurrStatus('suspended');
+      }
+      else if (event.deliveryType === 'resume') {
```
jebeck: Oh, I missed this before. There's no such thing as a `resume` `deliveryType`. We should just use a distinction between `suspended` and `null` for now, I guess.

jebeck:
I'm going to make the change re: `currStatus` `resume` and then merge. To answer a question you had before, yes, I am seeing annotations on all my midnight-segment basals.

PR #183, opened 2015-11-02
Error handling

gniezen:
Improving error handling for Tandem pump, including device disconnected)

### DIFF
--- a/lib/serialDevice.js
+++ b/lib/serialDevice.js
@@ -32,6 +32,13 @@ var debug = require('./bows')('SerialDevice');
 
 var moduleCounter = 1;
 
+var SEND_ERRORS = {
+  'disconnected' : 'Disconnected. Reconnect the device.',
+  'pending' : 'A send was already pending.',
+  'timeout' : 'The send timed out.',
+  'system_error' : 'A system error occurred. Reconnect the device.'
+};
+
 module.exports = function(config) {
   config = config || {};
   var connection = null;
@@ -373,7 +380,7 @@ module.exports = function(config) {
       }
       if (info.error) {
         debug('Serial send returned ' + info.error);
-        callback(new Error(info.error),null);
+        callback(new Error(SEND_ERRORS[info.error]),null);
       }
       callback(null, info);
     };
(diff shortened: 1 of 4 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4TbQOZKz&#x2F;24-2-tandem-error-handling-including-device-disconnected"> &lt;2&gt; [tandem] error handling (including device disconnected)</a>

lib/serialDevice.js
```diff
@@ -371,13 +373,19 @@ module.exports = function(config) {
...
       }
-      callback(info.error, info);
+      callback(null, info);
     };
     if (bytes == null) {
       debug('just tried to send nothing!');
     } else {
-      chrome.serial.send(connection.connectionId, bytes, sendcheck);
+      if(connection == null) {
+        debug('Device not connected.');
```
jebeck: I see this logged when I'm busy playing YouTube in the foreground with the uploader in the background, device still very much connected.

jebeck:
The 5 milli spin still doesn't seem to work for me (at least not while playing YouTube video fullscreen).

Would be happy to merge, as everything else looks reasonable, except I also don't don't get device disconnect error when I yank out the USB partway through. Did I misunderstand that you were handling that better now?

jebeck:
After the merge, on first test attempt at producing error by pulling out cable while fetching log records, I hung Chrome, killed all my USB ports, and ultimately had to hard reboot my computer with the power button.

On second attempt, pulling out cable at almost exact same moment got this as expected:
screenshot 2015-11-03 11 45 37

@gniezen I believe you had something else you wanted to push up? Feel free to do that now (after pulling down my merge from tandem-mvp with conflict resolution) and then merge if you like, but I also think further investigation is needed to see if the nasty computer frozen situation can be avoided.

gniezen:
I'm now clearing the `sendTimer` interval timer if an error is thrown while a packet is being sent, so I'm hoping this will fix the issue you were seeing where the app hangs. My guess is that it wasn't an issue on my machine because it just had a couple of thousand records to fetch, but with your pump (with tens of thousands of records?) it probably caused an overflow in the USB stack.

Refactoring the send/receive functions so that it only sends a packet after successfully receiving a packet will make this a non-issue, so I'm hoping to get that done in this spring as well.

PR #185, opened 2015-11-06
Serial send fix

gniezen:
When the Uploader window is not in focus, the thread priority changes and packets in the serial port buffer are not send immediately. As we're sending and receiving packets in parallel instead of sequentially, this means we have to wait a bit longer before trying to send the packet again. This PR implements a backoff algorithm when a serial port send is pending.

### DIFF
--- a/lib/drivers/tandemTslimDriver.js
+++ b/lib/drivers/tandemTslimDriver.js
@@ -600,6 +600,7 @@ module.exports = function (config) {
   };
 
   var TIMEOUT_ERROR = 'Timeout while attempting to talk to pump';
+  var SEND_INC = 5;  // amount to increment interval by if serial send pending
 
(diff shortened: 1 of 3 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;l4a5Bnm0&#x2F;40-review-jc-s-changes-to-binary-search-in-tandem-driver-and-handle-merge-conflicts-with-current-tandem-mvp-branch"> review JC&#39;s changes to binary search in Tandem driver and handle merge conflicts with current tandem-mvp branch</a>

jebeck:
I put my uploader in the background for a while and had a convo in Slack, did some e-mail, it seems to have slowed nearly to a stop while in the background, but then it picked up without problems when it was refocused. That feels like a win to me, so merging!

PR #186, opened 2015-11-06
Final cleanup for devel

gniezen:
**WIP**: Do not merge.


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;74byvknk&#x2F;25-1-tandem-final-driver-clean-up-remove-absurd-logging-and-be-more-choosy-about-what-to-log-use-bows"> &lt;1&gt; [tandem] final driver clean-up; remove absurd logging and be more choosy about what to log; use bows</a>

test/tandem/testTandemSimulator.js
```diff
@@ -0,0 +1,608 @@
...
+        type: 'bolus',
+        subType: 'normal'
+      };
+
+      it('passes through', function(){
+        simulator.bolus(val);
+        expect(simulator.getEvents()).deep.equals([val]);
+      });
+
+      it('does not pass through a zero-volume bolus that does not have an expectedNormal', function() {
```
jebeck: Could probably have tests of this sort for squares and duals too, no?

test/tandem/testTandemSimulator.js
```diff
@@ -0,0 +1,608 @@
...
+    };
+
+    it('passes through with a bolus', function() {
+      simulator.wizard(val);
+      expect(simulator.getEvents()).deep.equals([val]);
+    });
+  });
+
+  describe('deviceEvent', function() {
+  /* TODO: revisit these when all device events are implemented
```
jebeck: I'm not a huge fan of pulling in commented-out code, especially with the fixtures still OmniPod-style. (Ah, but _all_ the fixtures are OmniPod-style in this file...want to change the `deviceId` globally?)

I'd much prefer checking in pending tests (couple ways to do this with mocha - see the docs). That way we _see_ the pending tests when we run them locally.

test/tandem/testTandemSimulator.js
```diff
@@ -0,0 +1,608 @@
...
+        }
+      };
+      it('passes through', function() {
+        simulator.changeDeviceTime(change);
+        expect(simulator.getEvents()).deep.equals([change]);
+      });
+    });
+  });
+
+  /* TODO: revisit these when settings are implemented
```
jebeck: Actually, a test here showing that settings currently *don't* pass through seems appropriate to me.

test/testCommonSimulations.js
```diff
@@ -0,0 +1,114 @@
...
+            }
+          ]
+        }
+      };
+      var finalBasal = common.finalScheduledBasal(basal,settings,'test');
+      expect(finalBasal.annotations[0].code).to.equal('test/basal/fabricated-from-schedule');
+      expect(finalBasal.duration).to.equal(23001000); // 864e5 - millisInDay
+    });
+
+    it('has an off-schedule rate', function() {
```
jebeck: Bit terse on the assertion descriptions in this file...wouldn't really knowing just reading test output what "has an off-schedule rate" means or even the above "fabricates from only one schedule segment" (fabricates _what_?)

Maybe expand a bit for clarity?

jebeck:
Tests look pretty good - some comments in there. No 'event interplay' simulator tests yet, and I think there's _one_ candidate: we do have event interplay happening to determine how the 'new day' basals are utilized. So tests showing what happens when a new day event follows a basal that results in `currStatus` of `suspended` versus when it doesn't would be good. (See 'event interplay' simulator tests elsewhere for examples.)

test/tandem/testTandemSimulator.js
gniezen: I'm not sure it's possible to have an extended bolus where `bolex_insulin_requested` is not defined, as the pump requires you to put in at least 0.4u?

PR #188, opened 2015-11-11
fix debug mode \o/

jebeck:
JavaScript/webpack trivia question: why did this reverse the breakage of debug mode even when `isChromeApp === true`?

(Somehow the stuff inside the `if` block was *locally* (that is, only in this module) messing up the global `__DEBUG__` variable, even when the `if` block wasn't triggered. Huh?)

### DIFF
--- a/lib/driverManager.js
+++ b/lib/driverManager.js
@@ -22,11 +22,6 @@ var statusManager = require('./statusManager.js');
 var debug = require('./bows')('DriverManager');
 
 var isChromeApp = (typeof chrome !== 'undefined');
-if (!isChromeApp) {
-  /* jshint ignore:start */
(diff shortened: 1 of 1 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;DmY1pvu8&#x2F;64-uploader-tech-debt-triage-perhaps-attempt-to-fix-non-operational-uploader-debug-mode"> [uploader] [tech debt] triage &amp; perhaps attempt to fix non-operational uploader DEBUG mode</a>

gniezen:
Bizarre. Tested it with BCN driver and it works for me too. Merging.

PR #193, opened 2015-11-14
fix Tandem read only from most recent pump shut down when non-DEBUG mode

jebeck:
What it says on the tin. I hadn't adjusted the check for a shut down since we changed the way we read basal change types. /o\

Review please @gniezen and let's publish a new build to the devel Chrome store.


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GVVdQ8Ic&#x2F;24-1-uploader-device-final-bugfixes-for-initial-tandem-prod-release"> &lt;1&gt; [uploader] [device] final bugfixes for initial Tandem prod release</a>

gniezen:
LGTM. Merging.

PR #199, opened 2015-11-17
Adding Tandem alarm records

gniezen:
Builds alarm records from Tandem's `alarm_actived` and `alert_activated` events.

### DIFF
--- a/lib/drivers/tandemTslimDriver.js
+++ b/lib/drivers/tandemTslimDriver.js
@@ -573,7 +573,25 @@ module.exports = function (config) {
       name: 'USB Enumerated Event',
       format: 'f............',
       fields: ['negotiated_current_mA']
+    },
+    LID_ALARM_ACTIVATED: {
+      value: 0x05,
+      name: 'Alarm Activated Event',
+      format: 'i............',
+      fields: ['alarm_id']
     }
+  };
+
+  var ALARM_TYPES = {
+    ALARM_OCCLUSION: { value: 2, name: 'Occlusion detected'},
+    ALARM_AUTO_OFF: { value: 7, name: 'Auto Off'},
+    ALARM_OUT_OF_LIQUID: { value: 8, name: 'Out of Liquid'},
+    ALARM_TEMPERATURE_OUT_OF_RANGE: { value: 10, name: 'Temperature Out of Range'},
+    ALARM_EXTREMELY_LOW_LIPO: { value: 12, name: 'Extremely low LiPo percent charge'},
+    ALARM_STUCK_WAKE_BUTTON: { value: 22, name: 'Stuck Wake Button'},
+    ALARM_PRESSURE_OUT_OF_RANGE: { value: 24, name: 'Atmosphere Pressure Out of Range'},
+    ALARM_CARTRIDGE_REMOVED: { value: 25, name: 'Cartridge Removed'},
+    ALARM_SECOND_OCCLUSION: { value: 26, name: 'Second Occlusion detected in a row'}
   };
 
   var IDP_TDEP = {
@@ -649,6 +667,10 @@ module.exports = function (config) {
 
   var getLogRecordName = function (idx) {
     return _getName(PUMP_LOG_RECORDS, idx);
+  };
+
+  var getAlarmName = function (idx) {
+    return _getName(ALARM_TYPES, idx);
   };
 
   var getLogRecordById = function (idx) {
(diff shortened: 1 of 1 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;u5lF5b1k&#x2F;25-2-uploader-tandem-alarms"> &lt;2&gt; [uploader] [tandem] alarms</a>

lib/drivers/tandemTslimDriver.js
```diff
@@ -1650,6 +1685,76 @@ module.exports = function (config) {
...
+        switch (alarmValue) {
+          case ALARM_TYPES.ALARM_OCCLUSION.value:
+            alarmRecord = alarmRecord.with_alarmType('occlusion');
+            break;
+          case ALARM_TYPES.ALARM_SECOND_OCCLUSION.value:
+            alarmRecord = alarmRecord.with_alarmType('occlusion');
+            alarmRecord = alarmRecord.with_payload({alarm_name: alarmText});
+            break;
+          case ALARM_TYPES.ALARM_AUTO_OFF.value:
+            alarmRecord = alarmRecord.with_alarmType('auto_off');
```
jebeck: So what we have to determine is whether when an alarm like this is triggered, if the suspension of insulin delivery is reflected elsewhere (e.g., in a basal rate change with `pump_suspended`) or if we have to do it ourselves. It's your happy job @gniezen to trigger these on your pump since you're not actually dependent on insulin :P 

The ones that fall into this category are: occlusions, auto_off, no_insulin, no_power, and the cartridge removed alarm definitely.

lib/drivers/tandemTslimDriver.js
```diff
@@ -1650,6 +1685,76 @@ module.exports = function (config) {
...
+        var alarmValue = alarmEntry.alarm_id;
+        var alarmText = getAlarmName(alarmValue,ALARM_TYPES);
+
+        switch (alarmValue) {
+          case ALARM_TYPES.ALARM_OCCLUSION.value:
+            alarmRecord = alarmRecord.with_alarmType('occlusion');
+            break;
+          case ALARM_TYPES.ALARM_SECOND_OCCLUSION.value:
+            alarmRecord = alarmRecord.with_alarmType('occlusion');
+            alarmRecord = alarmRecord.with_payload({alarm_name: alarmText});
```
jebeck: Why only the `alarmText` in the payload here when elsewhere you also include the `alarmValue`?

lib/drivers/tandemTslimDriver.js
```diff
@@ -1650,6 +1685,76 @@ module.exports = function (config) {
     return records;
   };
 
+  var buildAlarmRecords = function (data,records) {
```
jebeck: I think you forgot to call the `buildAlarmRecords` function down where the other builders are called. (Or at least on first test I got no `subType: 'alarm'` records uploaded, and I don't see it.

lib/drivers/tandemTslimDriver.js
```diff
@@ -1650,6 +1685,76 @@ module.exports = function (config) {
...
+    var alarmRecords = filterLogEntries([PUMP_LOG_RECORDS.LID_ALARM_ACTIVATED,PUMP_LOG_RECORDS.LID_ALERT_ACTIVATED],data.log_records);
+    alarmRecords.forEach(function (alarmEntry) {
+      var alarmRecord = cfg.builder.makeDeviceEventAlarm()
+        .with_deviceTime(alarmEntry.deviceTime)
+        .set('index', alarmEntry.log_index);
+      cfg.tzoUtil.fillInUTCInfo(alarmRecord, alarmEntry.jsDate);
```
jebeck: All the alarms (now that I added a call to the `buildAlarmRecords` fn locally) have an `uncertain-timestamp` annotation. This means the bootstrapping module is confused, so check through everything here.

lib/drivers/tandemTslimDriver.js
gniezen: Yep, fixed it.

gniezen:
@jebeck I can confirm that there is a corresponding suspend in my data for every `auto_off` alarm. 

There was one occlusion during a bolus, which did show it as interrupted, but the pump was not suspended. I'm now trying to trigger an occlusion during a basal to see what happens. Will test the others as well.

jebeck:
Merging into tandem-v2 as discussed. Further changes may be needed pending inquiries re: alarms that might need corresponding suspends.

PR #206, opened 2015-11-24
fix hide devices

jebeck:
A decision was made - given ongoing issues with Tandem on Windows - to remove Tandem from the device menu on Windows machines. When I went to go do this, I discovered that the implementation of `hideUnavailableDevices` was pretty problematic - located in appState instead of appActions where it logically belongs, being called during the React render cycle (surprised this wasn't causing an Invariant Violation, TBH), and mutating React state directly instead of using `setState`. This branch fixes these shortcomings.

This PR also includes the change to hide Tandem on Windows machines.

Ping @gniezen for review.

### DIFF
--- a/test/ui/testAppActions.js
+++ b/test/ui/testAppActions.js
@@ -556,12 +556,12 @@ describe('appActions', function() {
 
     describe('[windows]', function() {
       beforeEach(function() {
-        app._os = 'win';
+        app.setState({_os: 'win'});
       });
 
       it('excludes only Tandem', function() {
         expect(app.state.uploads.length).to.equal(10);
-        appActions.hideUnavailableDevices();
+        appActions._hideUnavailableDevices();
         expect(app.state.uploads.length).to.equal(9);
         expect(_.findWhere(app.state.uploads, {key: 'tandem'})).to.not.be.ok;
       });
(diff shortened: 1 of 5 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GN4TNKFl&#x2F;53-uploader-fix-hideunavailabledevices-to-avoid-mutating-state-directly"> [uploader] fix hideUnavailableDevices to avoid mutating state directly</a>

jebeck:
Just looking back, this feature came in through #156 and it's totally my bad for not noticing/recognizing these shortcomings before. @gniezen you should take a look at our commentary there to make certain I haven't introduced any regressions; I did look, and I don't believe I have.

jebeck:
I built a test .zip for Nick, and he loaded it on the Windows machine in the office, and everything looked right - no Tandem, but Abbott devices present.

test/ui/testAppActions.js
```diff
@@ -548,6 +548,41 @@ describe('appActions', function() {
...
+      app.setState(state);
+    });
+
+    describe('[windows]', function() {
+      beforeEach(function() {
+        app.setState({_os: 'win'});
+      });
+
+      it('excludes only Tandem', function() {
+        expect(app.state.uploads.length).to.equal(10);
```
gniezen: I'm not crazy about hard-coding this, as it means every new device added will break this test. How about just testing that uploads.length decreases by the expected number after the hide function is called?

test/ui/testAppActions.js
```diff
@@ -548,6 +548,41 @@ describe('appActions', function() {
...
+      });
+    });
+
+    describe('[mac]', function() {
+      beforeEach(function() {
+        app.setState({_os: 'mac'});
+      });
+
+      it('excludes all Abbott devices', function() {
+        expect(app.state.uploads.length).to.equal(10);
```
gniezen: Ditto.

gniezen:
Pending two minor comments, LGTM and can be merged. Tested on OS X El Capitan, Windows 7 and 8.1.

test/ui/testAppActions.js
jebeck: I actually love it for exactly that reason. Every time we add a device we're forced to think about whether it should be included/excluded on platforms. For one thing, should help prevent the oversight of a platform like Tandem + Windows. I'll merge as-is for now, happy to revisit.

PR #207, opened 2015-11-26
Fixing Tandem driver issues on Windows

gniezen:
Addresses the following issues:

- Found an issue where it is assumed that a synchronization byte will always be at the beginning of the buffer, which is not always true. Fixed by discarding bytes correctly.
- On a "serial timeout" or "serial pending" error, we now wait SEND_INC_MAX ms before sending again (replaces backoff algorithm)
- Found an issue where sometimes if you send a request to the Tandem device, it just responds with a burst of 1s and 0s. This was fixed by retrying after a timeout, in which case it replies with the expected packet.
- On a "system_error" error, we now disconnect and reconnect automatically and continue where we left off.
- While recovering from a previous error, out-of-order packets were not re-requested. This was fixed by always re-requesting out-of-order packets.
- During a recovery sequence, we were still requesting other packets, which made things slower. We now only request the packet we are trying to actually recover.
- Binary search was using the wrong requester function.
- We now increase the progress bar while searching for the newest event. This is useful for devices that haven't been used in a while.

### DIFF
--- a/lib/drivers/tandemDriver.js
+++ b/lib/drivers/tandemDriver.js
@@ -598,8 +598,9 @@ module.exports = function (config) {
   };
 
   var pending = false;    //if serial send is pending, we have to wait before we send again
-  var SEND_FREQ = 5;  // time between sending serial packets
+  var INTERVAL_FREQ = 5;  // time between sending serial packets
   var SEND_WAIT = 800; // time to wait when serial send is pending
+  var RETRY_TIMEOUT = 1000; // time to wait before retrying to send packet
 
   var BASE_TIME = Date.UTC(2008, 0, 1, 0, 0, 0).valueOf(); /* new Date(2008, 0, 1, 0, 0, 0).valueOf(); */
   var addTimestamp = function (o, rawTime) {
@@ -782,7 +783,7 @@ module.exports = function (config) {
           callback(err,null);
         }
       });
-    },1000);
+    },RETRY_TIMEOUT);
 
     var listenTimer = setInterval(function () {
       if (cfg.deviceComms.hasAvailablePacket()) {
@@ -876,7 +877,7 @@ module.exports = function (config) {
                       }
                     });
                   }
-                },1000);
+                },RETRY_TIMEOUT);
               });
             }
             else {
@@ -913,7 +914,7 @@ module.exports = function (config) {
         };
         processPacket(cfg.deviceComms.nextPacket());
       }
-    }, 5);
+    }, INTERVAL_FREQ);
 
     var sendTimer = setInterval(function () {
       if (send_seq % 1000 === 0) {
@@ -930,7 +931,7 @@ module.exports = function (config) {
           send_seq++;
         }
       }
-    }, SEND_FREQ); // if we spin too quickly on this, packets don't get sent when window doesn't have focus
+    }, INTERVAL_FREQ); // if we spin too quickly on this, packets don't get sent when window doesn't have focus
   };
 
   var newestEventRequester = function (start, end, progress, callback) {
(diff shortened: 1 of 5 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Rc4iSBOz&#x2F;57-uploader-tandem-windows-issues"> [uploader] Tandem Windows issues</a>

lib/drivers/tandemDriver.js
```diff
@@ -838,72 +859,78 @@ module.exports = function (config) {
...
               }
+              callback(null, pkt);
             }
           }
         };
         processPacket(cfg.deviceComms.nextPacket());
-        delay.forEach(processPacket);
       }
-    }, 1);
+    }, 5);
```
jh-bate: what does this number correspond too?

lib/drivers/tandemDriver.js
```diff
@@ -937,14 +968,25 @@ module.exports = function (config) {
...
+                    debug('Retrying to recover..',receive_seq);
+                    tandemCommand(COMMANDS.LOG_ENTRY_SEQ_REQ, [receive_seq], function (err) {
+                      if(err) {
+                        callback(err,null);
+                      }
+                    });
                   }
-                });
-              } // drop out-of-order packets on the floor; they will be re-requested.
+                },1000);
```
jh-bate: I assume the timeout interval, would it be better to be a constant like the others?

test/ui/testAppActions.js
```diff
@@ -559,11 +559,10 @@ describe('appActions', function() {
         app.setState({_os: 'win'});
       });
 
-      it('excludes only Tandem', function() {
+      it('excludes nothing', function() {
         expect(app.state.uploads.length).to.equal(10);
         appActions._hideUnavailableDevices();
-        expect(app.state.uploads.length).to.equal(9);
-        expect(_.findWhere(app.state.uploads, {key: 'tandem'})).to.not.be.ok;
+        expect(app.state.uploads.length).to.equal(10);
```
jh-bate: It could be worth explicitly testing that tandem is included rather than assuming it is included because there are 10 items.

jh-bate:
In general LGTM @gniezen. A couple of cleanup suggestions but thats about it.

One question that I couldn't tell from the code - could we now get in the situation where we are trying to be clever recovering from errors that it just keeps trying and trying??

lib/drivers/tandemDriver.js
gniezen: We have a `sendTimer` interval timer and a `listenTimer` interval timer that are running at the same time at a set frequency. I'm now renaming this frequency as `INTERVAL_FREQ` and applying it to the relevant timers to improve readability.

lib/drivers/tandemDriver.js
gniezen: Good idea, I'm renaming it `RETRY_TIMEOUT`.

test/ui/testAppActions.js
gniezen: I was previously explicitly testing for each device (including BGMs), but @jebeck changed it to this strategy as I think the list of devices that may or may not be in the list can become unwieldy?

gniezen:
Thanks @jh-bate! I've addressed your comments and will build a new devel uploader off this branch for Nick and Howard to test.  

With regards to retrying when something fails - we only do this once in each case. What I do think we may need to do is to add the `abortTimer`s back in to throw an error, as we currently do nothing if it fails again after retrying (as in, we don't get stuck in a loop, but we also don't return a useful error).

PR #209, opened 2015-12-01
WIP Sprint 24 version check

jh-bate:
**DO NOT MERGE**

@jebeck here are the changes I was looking at. As discussed they will change with your changes to how the uploader is structured.

### DIFF
--- a/lib/state/appActions.js
+++ b/lib/state/appActions.js
@@ -225,8 +225,6 @@ appActions.load = function(cb) {
       return cb();
     }
 
-    appActions._assertUploaderVersion();
-
     async.series([
       api.user.account.bind(null),
       api.user.profile.bind(null),
@@ -253,7 +251,10 @@ appActions.load = function(cb) {
 appActions.login = function(credentials, options, cb) {
   var self = this;
 
-  appActions._assertUploaderVersion();
+  var err = appActions._assertUploaderVersion();
+  if (err) {
+    return cb(err);
+  }
 
   async.series([
     api.user.login.bind(null, credentials, options),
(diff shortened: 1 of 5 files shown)


lib/state/appActions.js
```diff
@@ -247,6 +250,12 @@ appActions.load = function(cb) {
 
 appActions.login = function(credentials, options, cb) {
   var self = this;
+
+  var err = appActions._assertUploaderVersion();
```
jh-bate: so I am trying to assert that the version of the uploader they are using is allowed to upload to the platform. If it isn't a required version then they are prompted to update before going any further

lib/state/appActions.js
```diff
@@ -579,6 +588,11 @@ appActions.upload = function(uploadIndex, options, cb) {
   var self = this;
   cb = cb || _.noop;
 
+  var err = appActions._assertUploaderVersion();
```
jh-bate: Might be un-necessary as hopefully they have been prompted to update before if there uploader is out of date

lib/state/appActions.js
```diff
@@ -627,6 +643,18 @@ appActions.upload = function(uploadIndex, options, cb) {
     appActions.addMoreInfoToError(error, appActions.errorStages['STAGE_UPLOAD']);
     throw error;
   }
+
+
+
+};
+
+appActions._assertUploaderVersion = function() {
```
jh-bate: the basic assert that returns an error if they need to update their uploader

jebeck:
This will be included in #219 so closing for clarity.

PR #210, opened 2015-12-03
Tandem Settings

gniezen:
*WIP*: Do not merge.

Builds the most recent Tandem settings, but not the settings history.

### DIFF
--- a/lib/drivers/tandemDriver.js
+++ b/lib/drivers/tandemDriver.js
@@ -646,6 +646,7 @@ module.exports = function (config) {
     o.jsDate = dt;
     o.deviceTime = sundial.formatDeviceTime(dt);
   };
+  var uploadTime = {};
 
   // This is a particularly weak checksum algorithm but that's what Insulet and Tandem use...
   var weakChecksum = function (bytes, offset, count) {
@@ -785,6 +786,11 @@ module.exports = function (config) {
         if (response.postprocess) {
           response.postprocess(packet.payload);
         }
+      }
+      if(packet.descriptor === RESPONSES.IDP_TE.value) {
+        // record device time from timestamp at upload for device settings
+        packet.timestamp = struct.extractBEInt(bytes,packet.packet_len - 6);
+        addTimestamp(uploadTime,packet.timestamp);
       }
     }
     packet.valid = true;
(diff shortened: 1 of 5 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;PajgquED&#x2F;28-1-uploader-tandem-settings"> &lt;1&gt; [uploader] [tandem] settings</a>

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;y70IsGQY&#x2F;29-uploader-tandem-settings-history"> ??? [uploader] [tandem] settings history</a>

gniezen:
@jebeck to review. This PR now only uploads the most recent settings. It should only be merged once we support Tandem-style settings in Blip, or alternatively disabled the in the simulator so that they are not uploaded.

jebeck:
@gniezen That is not the plan. The plan is to start uploading them and simply let tideline filter out what it can't render. That way, we don't have to make two changes (to blip and the uploader) once we have the rendering, we can just deploy the new blip.

gniezen:
Great, even better. Didn't realize you were going to make changes to
Jellyfish already.

jebeck:
There are no required changes to jellyfish to upload Tandem-style `pumpSettings`, unless I'm missing something...?

gniezen:
Oh yeah, it's Tideline that rejects it when there's no bgTarget high/range,
not Jellyfish. No issues then.

lib/objectBuilder.js
```diff
@@ -330,6 +330,21 @@ module.exports = function () {
     return rec;
   }
 
+  function makePumpSettingsAllObjects() {
```
jebeck: Feels to me that this could just as easily be named `makeTandemPumpSettings` and it'll be accurate for quite a while if not forever. If you prefer that as well, go ahead and rename.

test/tandem/testTandemSimulator.js
```diff
@@ -307,11 +307,6 @@ describe('tandemSimulator.js', function() {
       expect(simulator.getEvents()).deep.equals([settings]);
     });
 
-    //TODO: remove this test when we handle settings
```
jebeck: Do we have a pass-through test now though? Maybe on the other branch?

lib/drivers/tandemDriver.js
```diff
@@ -1288,15 +1288,13 @@ module.exports = function (config) {
...
       .with_basalSchedules(basalSchedules)
-      // TODO: remove filtering on activeName so all profiles are included
-      // and will need to add & use with_carbRatios, with_insulinSensitivities, with_bgTargets
-      .with_carbRatio(carbSchedules[activeName])
-      .with_insulinSensitivity(sensitivitySchedules[activeName])
-      .with_bgTarget(targetSchedules[activeName])
+      .with_carbRatios(carbSchedules)
+      .with_insulinSensitivities(sensitivitySchedules)
+      .with_bgTargets(targetSchedules)
       .with_time(sundial.applyTimezone(data.log_records[data.log_records.length - 1].jsDate, cfg.timezone).toISOString())
```
jebeck: Let's talk about what timestamp we're using here.

test/tandem/testTandemSimulator.js
gniezen: Pass-through test now added to this branch.

lib/drivers/tandemDriver.js
gniezen: I'm now using the "device time in settings packet at time of upload" timestamp, and added an explanation in the comments.

PR #211, opened 2015-12-04
Tandem Settings History

gniezen:
**WIP**: Do not merge.

We may not be able to get this working until we upload all records during the first upload, and subsequently use diff uploads. The reason is that if a schedule's name change, we have to look to the previous "new schedule" or "rename schedule" record to determine what the name was. If we only have the most recent 90 days' worth of data, we don't have enough information to do this.

We need to take into account:
- [ ] a profile/schedule being added or deleted
- [ ] a profile/schedule being renamed
- [ ] a segment in a profile being added or deleted
- [x] parameters in a segment being changed


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;y70IsGQY&#x2F;29-uploader-tandem-settings-history"> ??? [uploader] [tandem] settings history</a>

gniezen:
Wrong merge target :( Closing.

PR #213, opened 2015-12-04
Tandem V2 tests

gniezen:
**WIP**: Do not merge.

Testing:
- [x] temp basal was terminated early and it's the final basal (time_left)
- [x] temp basal in new-day event
- [x] temp basal crossing multiple segments

### DIFF
--- a/lib/tandem/tandemSimulator.js
+++ b/lib/tandem/tandemSimulator.js
@@ -114,16 +114,9 @@ exports.make = function(config){
         if (currBasal.deliveryType !== 'scheduled') {
 
           if (currBasal.deliveryType === 'temp') {
-            if(currTempBasal && currTempBasal.duration && currTempBasal.time_left) {
-              if(currTempBasal.time_left !== 0) {
-                currBasal.duration = currTempBasal.duration - currTempBasal.time_left;
-                annotate.annotateEvent(currBasal, 'tandem/basal/fabricated-from-time-left');
-              }
-            }else{
               if(!currBasal.isAssigned('duration')) {
                 currBasal.duration = 0;
                 annotate.annotateEvent(currBasal, 'basal/unknown-duration');
-              }
             }
             currBasal = currBasal.done();
           }
@@ -159,7 +152,6 @@ exports.make = function(config){
         setCurrTempBasal(event);
       }
       else if(event.subType === 'stop') {
-        //add time_left in case temp basal was terminated early and it's the final basal
         if(currTempBasal != null) {
           var tempBasal = currTempBasal;
           tempBasal.subType = 'stop';
@@ -167,15 +159,23 @@ exports.make = function(config){
           setCurrTempBasal(tempBasal);
         }
 
-        if(currBasal !== null) {
-          if(currBasal.deliveryType === 'temp' && _.find(currBasal.annotations,{code:'tandem/basal/fabricated-from-new-day'})) {
-          // if it's a temp basal fabricated from new-day event, check if temp basal was cancelled
-            if(currTempBasal.time_left !== 0) {
-              var duration = currBasal.duration - currTempBasal.time_left;
-              if(duration > 0) {
+        if (currBasal !== null) {
+          if (currBasal.deliveryType === 'temp' && currTempBasal && currTempBasal.time_left) {
+          // check if temp basal was cancelled
+            if (currTempBasal.time_left !== 0 ) {
+              var duration;
+              if (currBasal.isAssigned('duration')) {
+                // a new-day event during a cancelled temp basal
+                duration = currBasal.duration - currTempBasal.time_left;
+              }
+              else{
+                duration = currTempBasal.duration - currTempBasal.time_left;
+              }
+              if (duration >= 0) {
                 currBasal.duration = duration;
                 annotate.annotateEvent(currBasal, 'tandem/basal/fabricated-from-time-left');
-              }else{
+              }
+              else{
                 currBasal.duration = 0;
                 annotate.annotateEvent(currBasal, 'basal/unknown-duration');
               }
--- a/lib/tandem/tandemSimulator.js
+++ b/lib/tandem/tandemSimulator.js
@@ -87,7 +87,11 @@ exports.make = function(config){
         if(currBasal.deliveryType === 'temp') {
           if (currTempBasal != null) {
             currBasal.percent = currTempBasal.percent;
-            currBasal.payload['duration'] =  currTempBasal.duration;
+            if(currBasal.isAssigned('payload')) {
+              currBasal.payload['duration'] =  currTempBasal.duration;
+            }else{
+              currBasal.payload = {duration : currTempBasal.duration};
+            }
           }
         }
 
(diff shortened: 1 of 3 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Hoi0eNEW&#x2F;69-2-uploader-unit-tests-for-tandem-v2"> &lt;2&gt; [uploader] Unit tests for Tandem v2</a>

gniezen:
@jebeck Tests are ready for review. While writing the tests, I discovered issues with the current temp basal logic (if a temp basal is cancelled early and there's a newDay or finalBasal event) and fixed it. Given that the tests were written before I made the changes, I have reasonable confidence that it's now working correctly (TDD? ;) but we should of course test with real-world data as well.

lib/tandem/tandemSimulator.js
```diff
@@ -152,13 +152,36 @@ exports.make = function(config){
...
           setCurrTempBasal(tempBasal);
         }
+
+        if (currBasal !== null) {
+          if (currBasal.deliveryType === 'temp' && currTempBasal && currTempBasal.time_left) {
+          // check if temp basal was cancelled
+            if (currTempBasal.time_left !== 0 ) {
+              var duration;
+              if (currBasal.isAssigned('duration')) {
+                // a new-day event during a cancelled temp basal
```
jebeck: I don't quite follow the leap of logic in this comment. Why is this necessarily a _new-day_ event?

test/tandem/testTandemSimulator.js
```diff
@@ -368,81 +367,117 @@ describe('tandemSimulator.js', function() {
...
+      simulator.basal(suppressed1);
+      simulator.tempBasal(tempBasalStart);
+      simulator.basal(tempBasal);
+      simulator.basal(suppressed2);
+      simulator.tempBasal(tempBasalStop);
+      simulator.basal(basal3);
+      expect(simulator.getEvents()).deep.equals([
+        suppressed1.done(),
+        expectedTempBasal,
+        suppressed2.done()
```
jebeck: I'm having trouble with this one. I'm expecting to see another expected temp basal since it's crossing a schedule.

test/tandem/testTandemSimulator.js
```diff
@@ -472,6 +507,65 @@ describe('tandemSimulator.js', function() {
...
+        .with_rate(0.5);
+      var newDay = builder.makeScheduledBasal()
+        .with_time('2014-09-26T00:00:00.000Z')
+        .with_deviceTime('2014-09-26T00:00:00')
+        .with_timezoneOffset(0)
+        .with_conversionOffset(0)
+        .with_rate(1);
+      newDay.set('type', 'new-day');
+
+      var basal2 = builder.makeScheduledBasal()
```
jebeck: Maybe just add a comment here about how it's disjoint but you're just using it to force all the ones you're interested in to be simulated.

gniezen:
@jebeck I'm now only using `time_left` to calculate a duration for the final basal (if the temp basal was cancelled, but there hasn't been a new rate change event yet - unlikely, but possible). I've also added a test for uploading during a temp basal (where duration will be 0).

The test for a temp basal crossing segments should now also be fixed.

jebeck:
Why will duration be 0 when uploading during a final temp basal? We get a `duration` from the temp basal start, right?

lib/tandem/tandemSimulator.js
```diff
@@ -87,7 +87,11 @@ exports.make = function(config){
         if(currBasal.deliveryType === 'temp') {
           if (currTempBasal != null) {
             currBasal.percent = currTempBasal.percent;
-            currBasal.payload =  { duration: currTempBasal.duration };
+            if(currBasal.isAssigned('payload')) {
+              currBasal.payload['duration'] =  currTempBasal.duration;
```
jebeck: Oh I remember you asked me about the style here. One thing I _would_ change is to use dot notation: `currBasal.payload.duration` instead of `currBasal.payload['duration']`. At one point we had that option set in linting, even.

test/tandem/testTandemSimulator.js
```diff
@@ -387,81 +386,124 @@ describe('tandemSimulator.js', function() {
...
+                                        .set('percent',0.65)
+                                        .with_duration(900000)
+                                        .with_previous(suppressed2.done())
+                                        .done();
+
+      simulator.basal(suppressed1);
+      simulator.tempBasal(tempBasalStart);
+      simulator.basal(tempBasal);
+      simulator.basal(suppressed2);
+      simulator.basal(tempBasal2);
```
jebeck: I still don't think this test matches what happens on the Tandem. The `suppressed2` is not something that gets simulated. After all, the `tempBasal2` is what results from the `timed_segment` basal rate change (see driver line 1608).

jebeck: To check my own understanding here I threw in some logging while uploading to look at a temp crossing schedule boundary:
```
basal called in simulator 2015-12-01T13:00:01 ["timed_segment"]
tempBasal called in simulator undefined start
basal called in simulator 2015-12-02T03:44:15 ["temp_rate_start"]
basal called in simulator 2015-12-02T04:04:15 ["timed_segment"]
basal called in simulator 2015-12-02T06:44:15 ["temp_rate_end"]
tempBasal called in simulator undefined stop
basal called in simulator 2015-12-02T08:04:15 ["timed_segment"]
```

We're not building anything from a change_type `temp_rate_end` so we do indeed have the equivalents of a `suppressed1`, `tempBasalStart`, `tempBasal`, `tempBasal2`, `tempBasalStop`, and `basal3`, in that order. So remove the `suppressed2` plz?

test/tandem/testTandemSimulator.js
```diff
@@ -536,6 +637,63 @@ describe('tandemSimulator.js', function() {
...
+          };
+      var temp = builder.makeTempBasal()
+        .with_time('2014-09-25T18:05:00.000Z')
+        .with_deviceTime('2014-09-25T18:05:00')
+        .with_timezoneOffset(0)
+        .with_conversionOffset(0)
+        .with_rate(1.3);
+
+      var expectedTempBasal = _.cloneDeep(temp);
+      expectedTempBasal.duration = 0;
```
jebeck: So just fix this as we discussed, since we know the duration (remember to remove annotation too).

test/tandem/testTandemSimulator.js
```diff
@@ -591,6 +749,71 @@ describe('tandemSimulator.js', function() {
...
+      expectedNewDay = expectedNewDay.done();
+
+      simulator.tempBasal(tempBasalStart);
+      simulator.basal(temp);
+      simulator.newDay(newDay);
+      simulator.tempBasal(tempBasalStop);
+      simulator.basal(basal2);
+      expect(simulator.getEvents()).deep.equals([expectedTempBasal,expectedNewDay]);
+
+      // check with different order of events
```
jebeck: Have you _seen_ both orders? The second order seems rather non-sensical to me. Also, are we/can we/should we maybe build the regular set of timestamps (through BtUTC) on temp basal start and stop events? I sort of get that they aren't necessary, but would be useful for logging/debugging in hairy situations, potentially.

gniezen: I thought we were already building BtUTC timestamps, but it turns out they weren't being done correctly. Agreed that it's helpful and fixing it now.

gniezen:
Addressed all comments. Merging.

PR #215, opened 2015-12-08
Tandem v2 cleanup

gniezen:
A couple of leftover tasks before we can push to prod.

Should be reviewed and merged **after** tandem-v2-tests branch.


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;qnZDHF6y&#x2F;71-2-uploader-tandem-v2-cleanup"> &lt;2&gt; [uploader] Tandem v2 cleanup</a>

gniezen:
@jebeck This one is to review and test after tandem-v2-tests have been merged into tandem-v2.

lib/tandem/tandemSimulator.js
```diff
@@ -79,6 +79,22 @@ exports.make = function(config){
...
     alarm: function(event) {
+      if (currStatus !== null && currStatus.status === 'suspended') {
+        var type = event.alarmType;
+        if(type === 'occlusion' || type === 'auto_off' || type === 'no_insulin' || type === 'no_power' ||
+            (type === 'other' && event.payload.alarm_name === 'Cartridge Removed') ) {
+
+          var status = currStatus;
+          status.reason = {suspended: 'automatic'};
+          if(type === 'other') {
+            status.payload = {cause: type, description: event.payload.alarm_name};
```
jebeck: Sadly, probably need to do the check for `isAssigned('payload')` and all that here too. The logIndices for the `suspended` are being steamrolled.

jebeck:
The fabrication of the suspended basals from occlusions isn't quite right yet. Saw this annotation:
screenshot 2015-12-09 18 49 51

This is the `basal/mismatched-series` which is added by jellyfish when the basal sequencing is interrupted because there's one w/no previous. Explained [here](http://developer.tidepool.io/data-model-v1/v1/basal/). I think you're not providing a previous (or not the right one) on the fabricated suspended basal. See also other comments above.

lib/tandem/tandemSimulator.js
```diff
@@ -79,6 +79,22 @@ exports.make = function(config){
 
   return {
     alarm: function(event) {
+      if (currStatus !== null && currStatus.status === 'suspended') {
+        var type = event.alarmType;
+        if(type === 'occlusion' || type === 'auto_off' || type === 'no_insulin' || type === 'no_power' ||
```
jebeck: The only one I could check this with in my data was occlusion, but I saw suspended.reason = 'automatic' as expected :)

lib/drivers/tandemDriver.js
```diff
@@ -1801,14 +1802,31 @@ module.exports = function (config) {
       if(alarmEntry.alarm_id != null) {
         var alarmValue = alarmEntry.alarm_id;
         var alarmText = getAlarmName(alarmValue,ALARM_TYPES);
+        var postbasal = null;
 
         switch (alarmValue) {
           case ALARM_TYPES.ALARM_OCCLUSION.value:
             alarmRecord = alarmRecord.with_alarmType('occlusion');
```
gniezen: As per our discussion: Since they should be infrequent, I'm going to annotate these suspended basals as `tandem/basal/fabricated-from-occlusion-alarm`.

lib/tandem/tandemSimulator.js
gniezen: It will be good to check the `cause` in the payload as well.

lib/tandem/tandemSimulator.js
gniezen: It looks like Jellyfish doesn't store any extra information with the cause in 'suspended' and 'resumed' in the payload, so I'm only going to fill in the suspend reasons for the other events (occlusion, no_power, no_insulin, auto_off). 

It also looks like 'cartridge removed' alarm events arrive before the suspended device event, so would then need a `setCurrAlarm` and all that goes with it (processing alarm clear events etc.). If we're then only storing {cause:'other'}, I feel like it's not worth the effort?

gniezen:
@jebeck Hmm, that mismatched-series is a strange one, because fabricated suspended basals are handled just like all the others, so their `previous` get added in the simulator. I don't see the annotation with mine - what is different with that one?

lib/tandem/tandemSimulator.js
jebeck: Yeah, not worth it.

lib/drivers/tandemDriver.js
jebeck: Can confirm I'm seeing this in my data :)

jebeck:
Oooooooh, the plot thickens. The `basal/mismatched-series` occurs because - as near as I can tell thus far from putting in a bunch of logging (in the simulator) and examining what's happening - we _do_ get a basal rate change with a `pump_suspended` change_type when the occlusion happens during a basal. The "during a basal" is what's different, in other words. So, um, this is the part where you get to say "I told you so! This is why I was trying so hard to produce an occlusion during a basal!" :sheep: 

So basically we have _two_ suspend basals at the same timestamp - we're seeing the one sourced from the Tandem basal rate change event first and giving _it_ the proper `previous`, but then it gets assigned a zero duration and/or gets skipped because it's at the same timestamp (see lines 115-116 of the simulator). The second one then doesn't get a `previous`, hence the annotation.

So we may (a) need to live with this quite extreme edge case and the resulting annotation (it being so hard to produce occlusions during basals) or (b) figure out a way to distinguish between bolus and basal occlusions. I'm leaning to (a).

gniezen:
Turns out it was a simple fix after I was able to reproduce it using a test. We should've been returning early when a duplicate is found, otherwise the duplicate overwrites the current basal in the simulator.

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VMXiTaII&#x2F;16-2-tandem-v2-cleanup"> &lt;2&gt; Tandem v2 cleanup</a>

lib/drivers/tandemDriver.js
```diff
@@ -1009,7 +1000,14 @@ module.exports = function (config) {
       PUMP_LOG_RECORDS.LID_BOLUS_REQUESTED_MSG1.value,
       PUMP_LOG_RECORDS.LID_BOLUS_REQUESTED_MSG2.value,
       PUMP_LOG_RECORDS.LID_BOLUS_REQUESTED_MSG3.value,
-      PUMP_LOG_RECORDS.LID_IDP_TD_SEG.value
+      PUMP_LOG_RECORDS.LID_IDP_TD_SEG.value,
```
jebeck: Since we're not reading settings history, it seems odd to me to include `LID_IDP_TD_SEG`, but it's (very) possible we discussed this, and I've forgot. Remind me?

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;jJKb0LSY&#x2F;13-1-uploader-documenting-tandem-specific-notes-on-checklist"> &lt;1&gt; [uploader] documenting Tandem-specific notes on checklist</a>

lib/drivers/docs/tandem.md
```diff
@@ -33,7 +33,10 @@
     - `[x]` basal rate interval with a start time, duration "guessed" from settings, rate delivered, and an annotation re: the "guessed" duration OR
     - `[ ]` basal rate interval with a start time and rate, no (= zero) duration
 
-Device-specific? (Add any device-specific notes/additions here.)
+##### Device-specific? (Add any device-specific notes/additions here.)
+
+- we cannot add the name of the basal schedule on each scheduled basal rate interval at the moment, as that would require a full settings history
+- we cannot represent suppressed scheduled basals during suspended basals at the moment, as we require a full settings history to determine them for each segment of the basal schedule that the suspended basal intersects
```
jebeck: Maybe add a note explaining that Tandem doesn't _have_ manual temp basals, and that's why they aren't implemented?

lib/drivers/docs/tandem.md
```diff
@@ -73,46 +76,50 @@ Device-specific? (Add any device-specific notes/additions here.)
...
+    - `[x]` low insulin
+    - `[x]` no insulin
+        - `[ ]` needed to infer a suspend (stoppage of all insulin delivery)
+    - `[x]` low power
+    - `[x]` no power
         - `[?]` needed to infer a suspend (stoppage of all insulin delivery)
+    - `[x]` occlusion
+        - `[x]` needed to infer a suspend (stoppage of all insulin delivery)
     - `[?]` no delivery
         - `[?]` needed to infer a suspend (stoppage of all insulin delivery)
```
jebeck: Maybe include a note about the `?`s here for `no delivery`? If my memory serves, there just isn't any alarm type that straightforwardly maps to `no delivery` for Tandem (which is fine but good to note).

lib/drivers/docs/tandem.md
```diff
@@ -73,46 +76,50 @@ Device-specific? (Add any device-specific notes/additions here.)
...
+- For status events reasons and causes are provided from the alarms if available, i.e., for occlusion, auto-off, no-insulin and no-power alarms. For cartridge-change alarms, the cause will show as other.
 
 #### SMBG
 
   - `[x]` blood glucose value
   - `[x]` subType (`linked` or `manual`)
-  - `*[?]` units of value (read from device, not hard-coded)
+  - `[ ]` units of value (read from device, not hard-coded)
   - `*[?]` out-of-range values (LO or HI)
   - `*[?]` out-of-range value thresholds (e.g., often 20 for low and 600 for high on BGMs)
```
jebeck: Is the question mark still accurate for these? This info is just not in the spec? Maybe we should ask Tandem.

lib/drivers/docs/tandem.md
```diff
@@ -174,14 +181,16 @@ No Tidepool data model yet:
...
-  - `[ ]` insulin action time
+        - `[-]` enabled
+        - `[-]` increment
+  - `[-]` insulin action time
   - `[ ]` clock display preference (12h vs 24h format)
 
-Device-specific? (Add any device-specific notes/additions here.)
+##### Device-specific? (Add any device-specific notes/additions here.)
+
+- To build a full settings history we'll need to use the entire log record, not just the last 90 days. For example, a profile name change event only contains the new name, so you'll need to read the record where it was first created to get the original name. Until we do diff uploads (so it only happens once), or significantly improve the upload speed, reading the entire log record is not possible.
```
jebeck: Add a note maybe about how instead of 'bolus “wizard”/calculator enabled' Tandem has carbs enabled or not (which is not quite the same thing).

lib/drivers/docs/tandem.md
```diff
@@ -155,18 +162,18 @@ Settings history:
...
+    - `[-]` BG reminder
+    - `[-]` bolus reminder
+  - `[-]` alert settings (volume or vibration-only; whether enabled)
   - basal features:
     - `[ ]` temp basal type (`manual` or `percentage`)
-    - `[ ]` max basal (as a u/hr rate)
+    - `[-]` max basal (as a u/hr rate)
   - bolus features:
     - `[ ]` bolus "wizard"/calculator enabled
     - `[ ]` bolus increment for non-"quick"/manual boluses
```
jebeck: I think the bolus increment is available. Double check? (I can set it in the UI under Options -> My Pump -> Pump Settings; it's not part of the IDP though.)

jebeck:
Remove the `x` from "UTC timestamp (Hey, one can dream!)" in the checklist plz - that's for the device providing the UTC timestamp _directly_, not us boostrapping it 😛

jebeck:
Other than checklist comments/TODOs and the upload-during-extended-bolus issue that we still need to make a strategy decision on, this PR LGTM.

lib/drivers/docs/tandem.md
gniezen: Since you're entering the value manually, I suppose it doesn't make sense to have LO/HI or out-of-range? Out-of-range/threshold info does exist in the spec for CGM data - I've put in a note.

lib/drivers/docs/tandem.md
gniezen: The spec only seems to have a bolus increment for quick boluses?

lib/drivers/docs/tandem.md
gniezen: Done.

lib/drivers/docs/tandem.md
gniezen: Done.

gniezen:
OK, I've addressed the checklist comments, so merging into tandem-v2.

lib/drivers/docs/tandem.md
jebeck: Oh. I think it's a typo in the checklist - should be bolus increment _for_ "quick"/manual boluses! I'll assess (and maybe update all pump checklists) later.

PR #166, opened 2015-10-16
Gordyd/react0.14.0

GordyD:
Update React to version 0.14.0 and also use babel-loader in Webpack.

Fix any breaking changes and make updates to get rid of any warnings.

PR #219, opened 2016-01-05
migrate state management to redux

jebeck:
**WIP** Don't merge.

This will include #166 and #209, so I'm closing those.


docs/state/preredux/Example.md
```diff
@@ -0,0 +1,152 @@
...
+      "key": "bayercontournextlink",
+      "source": {
+        "type": "device",
+        "driverId": "BayerContourNextLink"
+      }
+    }
+  ],
+  "_os": "mac"
+}
+```
```
GordyD: DAYUM!

docs/state/preredux/README.md
```diff
@@ -0,0 +1,62 @@
...
+The Tidepool uploader currently manages state using a single global object to represent the entirety of the application's state. All possible changes to the global state are performed through a set of functions found in [`lib/state/appActions.js`](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/state/appActions.js).
+
+Simple attributes of the global state object:
+
+- `dropMenu` - a boolean encoding whether the dropdown menu available by clicking on the logged-in user's name in the upper-right corner is currently visible (`true`) or hidden (`false`)
+
+- `howToUpdateKBLink` - the URL for [the Tidepool knowledge base article explaining how to update the uploader](https://tidepool-project.helpscoutdocs.com/article/6-how-to-install-or-upgrade-the-tidepool-uploader-gen)
+
+- `page` - the current "page" of the app being rendered
+   + "page" in scare quotes because of the uploader's nature as a single-page app with no router
```
GordyD: what are scare quotes?!

lib/components/Login.js
```diff
@@ -89,35 +83,22 @@ var Login = React.createClass({
 
   handleLogin: function(e) {
     e.preventDefault();
-    var username = this.refs.username.getDOMNode().value;
-    var password = this.refs.password.getDOMNode().value;
-    var remember = this.refs.remember.getDOMNode().checked;
+    var username = this.refs.username.value;
```
GordyD: Did you update the version of React too? (This was a API change in 0.14.0 right?)

lib/containers/root/Root.js
```diff
@@ -0,0 +1,24 @@
...
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+/* global __DEBUG__ */
+
+if (__DEBUG__ === true) {
```
GordyD: Nice

lib/redux/actions/async.js
```diff
@@ -0,0 +1,560 @@
...
+    dispatch(syncActions.uploadProgress(step, percentage));
+  };
+}
+
+function makeUploadCb(dispatch, getState, errCode, utc) {
+  return (err, recs) => {
+    const { devices, uploads, users, version } = getState();
+    const targetDevice = devices[uploads.uploadInProgress.pathToUpload[1]];
+    if (err) {
+      const serverDown = 'Origin is not allowed by Access-Control-Allow-Origin';
```
GordyD: From the error message, and its relation to CORS, this seems to me like a funny variable name. Something like serverError, might be better?

lib/redux/actions/async.js
```diff
@@ -0,0 +1,560 @@
...
+      }
+    });
+  };
+}
+
+/*
+ * COMPLEX ACTION CREATORS
+ */
+
+function getUploadsByUser(targetsByUser) {
```
GordyD: Could this be located somewhere else e.g. separate file, or a nested function if singular use?

GordyD: Not sure if either are preferential just a point of discussion.

lib/redux/actions/index.js
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import * as asyncActions from './async';
+import * as syncActions from './sync';
+
+export default {
+  asyncActions: asyncActions,
```
GordyD: I do still find this kind of weird, with the repetition of the word 'action' e.g.

```
import action from './lib/redux/actions'

actions.asyncActions.doSomething()
```

If you do just import the async actions at one point you can `import { async as asyncActions }` from './lib/redux/actions'`

```
import { async as asyncActions }` from './lib/redux/actions`

asyncActions.doSomething()
```

Again preference here, no big deal either way.

package.json
```diff
@@ -7,41 +7,67 @@
...
+    "lint": "./node_modules/.bin/eslint lib test"
   },
   "dependencies": {
-    "async": "0.9.0",
-    "babyparse": "0.4.1",
-    "blueimp-md5": "1.1.0",
-    "commander": "2.7.1",
-    "lodash": "3.6.0",
-    "node-uuid": "1.4.3",
-    "react": "0.12.0",
```
GordyD: Yea ok, React update!!

test/browser/redux/actions/sync.test.js
```diff
@@ -0,0 +1,1025 @@
...
+    });
+
+    describe('logoutSuccess', () => {
+      it('should be an FSA', () => {
+        let action = syncActions.logoutSuccess();
+
+        expect(isFSA(action)).to.be.true;
+      });
+
+      it('should create an action to announce the success of logout', () => {
```
GordyD: Great use of the word "announce" ;)

GordyD:
I was wondering where the async tests where, but then I realised that the file was too big to show in the merge. A lot of work has gone in here. It seems to a very high quality as far as I can tell.

Everything seems consistent minus the few actual question points I raised around naming.

Tests read well and all pass on my local.

:+1

lib/serialDevice.js
```diff
@@ -266,11 +266,14 @@ module.exports = function(config) {
...
-          }
+          chrome.runtime.getPlatformInfo(function(platformInfo) {
+            if ((platformInfo.os === 'mac') && (deviceInfo.driverId === 'Dexcom') &&
+              (deviceInfo.path.match(deviceInfo.portPattern) === null)) {
+              return cb(new Error('You may also need to power off and on your Dexcom receiver. ' + deviceDebugInfo));
+            }
 
-          cb(new Error('Could not connect to a matching device port: ' + deviceDebugInfo));
+            cb(new Error('Could not connect to a matching device port: ' + deviceDebugInfo));
+          });
```
gniezen: @jebeck I think the "matching device: callback should be outside of `getPlatformInfo`?

Also, can't we read the OS state here and is that why we have to call the chrome runtime?

jebeck: Yeah, this is the thing I thought you might find :)

Re: the location of the callback, I moved it in because otherwise it could get called too early, no? Since the `getPlatformInfo` is itself apparently asynchronous? It seems to still work in the desired circumstances - I get the 'Could not connect to matching device port' error details when clicking upload on Tandem with no Tandem plugged in, for example.

And indeed, we can't read the os state here the way we used to because before the app state was bound to a global app object, and honestly that was a hack (essentially to get around the clean separation of concerns). It's not a best practice, and I didn't want to continue it with redux, especially not here since this additional check is temporary, right? Once Chrome 48 is widely adopted we'll have better ways to distinguish the Dexcom error types?

gniezen: OK, yes, you're right - `getPlatformInfo` is asynchronous and we still get the desired result.

Basically we're waiting for Apple to fix their El Capitan USB stack, then we can remove the Dexcom-specific error. Or at least until most of our user base have the Tidepool Mac USB driver installed.

docs/state/preredux/README.md
jebeck: Hahaha, former linguist in me comes out. Wikipedia says it best: [scare quotes](https://en.wikipedia.org/wiki/Scare_quotes). Here the point is that calling it a "page" is not very literal since we don't have a router in the uploader.

lib/components/Login.js
jebeck: Yup, in fact this branch builds on *your* branch to do the React 14 update in the uploader. Remember that? It was *ages* ago. You have two commits at the top of this PR ;)

lib/containers/root/Root.js
jebeck: We're actually going to break this out further, because the redux dev tools are a *huge* performance suck and the logger especially is very very noisy if you working on devices (updates to the progress bar fire an action for *every* update). So we're probably going to have a collection of configs: `__DEBUG__`, `__REDUX_LOG__`, `__REDUX_DEV_UI__` etc.

lib/redux/actions/async.js
jebeck: The relation to CORS seems to actually have something to do with how Chrome apps work - i.e., I don't think they are *actually* CORS errors, they just surface that way in the app. But point taken, I'll rename.

lib/redux/actions/async.js
jebeck: It's used twice (in two thunks), which is why I factored it out. Maybe I'll do a utils file with similar functions I pulled out (`makeUploadCb` etc.).

lib/redux/actions/index.js
jebeck: I'm happy to match however you're doing this in blip - no strong opinion.

lib/redux/actions/async.js
```diff
@@ -0,0 +1,523 @@
...
+
+let services = {};
+let versionInfo = {};
+let daysForCareLink = null;
+
+/*
+ * ASYNCHRONOUS ACTION CREATORS
+ */
+
+export function doAppInit(config, servicesToInit) {
```
GordyD: Ah I see, so the reason you do not need to pass in the api into each async action is because you require this action, which initialises api, be performed prior to any other async action calls?

lib/redux/actions/async.js
```diff
@@ -0,0 +1,523 @@
...
+    dispatch(syncActions.setVersion(config.version));
+    services = servicesToInit;
+    versionInfo.semver = config.version;
+    versionInfo.name = config.namedVersion;
+    daysForCareLink = config.DEFAULT_CARELINK_DAYS;
+    const { api, carelink, device, localStore, log } = services;
+
+    dispatch(syncActions.initRequest());
+
+    async.series([
```
GordyD: With asynchronous handlers like `async` or `Promises` I tend to find it a bit clearer if you don't use inline functions unless it is a 1-2 liner. You can then put then clear function names instead. A good example is the further down in this file:

```
async.series([
        api.user.account,
        api.user.profile,
        api.user.getUploadGroups
      ], (err, results) => { ... });
```

What do you think about doing this for improved readability here? This statement could be shortened too:

```
aysnc.series([
  initialiseLocalStore,
  initialiseDevice,
  initialiseCareLink,
  initialiseApi,
  setupApiHosts
], (err, results) => { 
  ... 
});
```

lib/redux/reducers/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import { combineReducers } from 'redux';
+import * as reducers from './reducers';
```
GordyD: Not sure I understand the naming of this file here...

Possibly could be renamed to rootReducers if these reducers just operate on the root level of the state?

lib/redux/reducers/reducers.js
```diff
@@ -0,0 +1,173 @@
...
+    case actionTypes.READ_FILE_FAILURE:
+    case actionTypes.UPLOAD_FAILURE:
+    case actionTypes.UPLOAD_SUCCESS:
+      return false;
+    default:
+      return state;
+  }
+}
+
+export const working = combineReducers({
```
GordyD: Cool!

lib/redux/reducers/index.js
jebeck: Renaming to just `misc.js` (as many/all of the reducers in `users.js` and `uploads.js` are also at the root level).

lib/redux/actions/async.js
jebeck: Huh, the code I ported to redux was styled the way you suggest. I disprefer that style, but I guess I'm the minority since it was written your way before too. I'll change it.

PR #220, opened 2016-01-06
only reference webpack-injected __DEBUG__ from files not needed for CLI node scripts

jebeck:
Review please @gniezen? And maybe give the CLI tools a try again yourself to double-check me?

### DIFF
--- a/lib/core/carelink.js
+++ b/lib/core/carelink.js
@@ -15,6 +15,8 @@
  * == BSD2 LICENSE ==
  */
 
+/* global __DEBUG */
+
 var _ = require('lodash');
 
 var driverManager = require('../driverManager');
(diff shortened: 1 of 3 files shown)


gniezen:
Successfully uploaded with both the CareLink and OmniPod CLI tools. Also checked that DEBUG_ERROR environment variable still works for Uploader. I'm assuming it's still safe for the Tandem driver to use the `__DEBUG__` global?

LGTM!

jebeck:
Yup, the Tandem driver has no CLI tool that `require`s it, so no problems there.

Can we include this in one of the small bundles so it will get the benefit of whatever smoke testing happens with those?

gniezen:
Merging to include in OmniPod bundle of small tasks.

PR #221, opened 2016-01-07
Refactor Bayer Contour ASTM driver. Add support for Contour NEXT LINK 2.4

pazaan:
I agree to the terms of Tidepool Project’s Volunteer/Contributor License Agreement v1.0
as it exists at http://tidepool-org.github.io/TidepoolVCLA.pdf on 7th January 2016.


lib/drivers/bayerContourNext.js
```diff
@@ -44,111 +45,117 @@ module.exports = function (config) {
...
+    STX : 0x02
   };
 
+  var astmFrameBuffer = {
+    reset: function(){
+      this.bytes = new Uint8Array(0);
+      this.valid = false;
+      this.frame_len = 0;
+      this.payload = null;
+      return this;
```
gniezen: Is it necessary to return here, as we don't seem to be using the returned value?

lib/drivers/bayerContourNext.js
```diff
@@ -44,111 +45,117 @@ module.exports = function (config) {
    * TODO This info should be collected in the initial phase cause we can't assert if
    *      it is the same for all contour devices
    * */
-  var REPORT_BYTES = {
-    reportID: 0x41, // A
-    //checksum: 0x00, // Checksum is not been used by bayer contour device
-    hostID  : 0x42, // B
-    deviceID: 0x43  // C
-  };
+  var MAGIC_HEADER = 'ABC';
```
gniezen: Much better! :)

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +272,134 @@ module.exports = function (config) {
...
-    parseDataRecord(record,function(err,r){
-      var robj = {};
-      if (err) {
-          debug('Failure trying to read record', record);
-          debug(err);
-          return callback(err, null);
+  var getOneRecord = function (data, callback) {
+    var cmd = buildAckPacket();
+    bcnCommandResponse(cmd, function (err, record) {
+      // TODO - add error handling
```
gniezen: What error handling still needs to be done here?

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +272,134 @@ module.exports = function (config) {
...
-
-    hidDevice.send(commandpacket.packet, function () {
-        receivePacket(5000, commandpacket.parser, function(err, packet) {
-            var rawtext = '';
-            while(hidDevice.hasAvailablePacket()){
-                var pkt = hidDevice.nextPacket();
-                pkt.parsed_payload = commandpacket.parser(pkt);
-                rawtext += pkt.parsed_payload;
-            }
+  var debugFrame = function (astmFrame) {
```
gniezen: I guess if we're not using this function anymore it's better to remove it.

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +272,134 @@ module.exports = function (config) {
...
+    hidDevice.send(commandpacket, function () {
+      getASTMFrame(5000, 3, function(err) {
+        if (err !== null || !hidDevice.hasAvailablePacket()) {
+          // TODO - we should do retries here, depending on the error
+          callback(err, null);
+        } else {
+          // We're using the hidDevice as a message/frame buffer, not a USB packet buffer
+          // to save on buffer copying. It could be refactored differently easily enough.
+          var frame = hidDevice.nextPacket();
+          // TODO - we should ideally check the validity of the ASTM frame (sequence number, checksum etc) here as well for retries
```
gniezen: I thought we are verifying the checksum, or is there a separate checksum for the ASTM frame itself?

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +272,134 @@ module.exports = function (config) {
...
     });
   };
 
-  var receivePacket = function (timeout, parser, callback) {
+  var getASTMFrame = function (timeout, retries, callback) {
     var abortTimer = setTimeout(function () {
       clearInterval(listenTimer);
       debug('TIMEOUT');
-      callback('TIMEOUT', null);
+      callback('TIMEOUT');
```
gniezen: To have this surface properly in the UI, should be changed to `callback(new Error('Timeout error.'),null);`

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +272,134 @@ module.exports = function (config) {
...
-            });
-          }
-          else if(rec.TYPE == ASCII_CONTROL.TER){
+        var packet = new Uint8Array(raw);
+
+        var packetHead = struct.unpack(packet, 0, '3Z2b', ['HEADER', 'SIZE', 'BYTE1']);
+
+        if(packetHead['HEADER'] != MAGIC_HEADER){
+          debug('Invalid packet from Contour device');
+          callback('INVALID_USB_PACKET');
```
gniezen: Rather `callback(new Error('Invalid USB packet'), null);`?

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +272,134 @@ module.exports = function (config) {
...
             clearTimeout(abortTimer);
             clearInterval(listenTimer);
-            callback(null, '');
-          }
-        });
-      }, 20); //we need to be ready for the next packet quickly
+            callback(null);
+        }
+      });
+    }, 20); // The Contour timeout is 15 seconds, but let's be slow enough to be kind to the CPU.
```
gniezen: Maybe make this comment a bit clearer that the 20 is in milliseconds, not seconds?

lib/drivers/bayerContourNext.js
```diff
@@ -416,7 +441,7 @@ module.exports = function (config) {
           }
         dataToPost.push(smbg);
       }
-    }else{
+    } else {
       debug('Device has no records to upload');
       throw(new Error('Device has no records to upload'));
```
gniezen: can we return this in a callback so that it will surface properly in the UI?

lib/drivers/bayerContourNext.js
```diff
@@ -471,19 +495,21 @@ module.exports = function (config) {
...
-        var rec = data.rawrecords.shift();
         progress(100.0 * recnum / data.nrecs);
         setTimeout(function() {
-          getOneRecord(rec, data, cb);
+          getOneRecord(data, cb);
         }, 20);
       }
 
-      async.timesSeries(data.nrecs, getOneRecordWithProgress, function(err, result) {
+      // FIXME - We're adding +1 to skip the Patient Record (P)
```
gniezen: I don't quite follow what is going on here?

lib/drivers/bayerContourNext.js
pazaan: We are using it, further down.
var astmFrameBuffer = {
...
}.reset();
It's like a constructor of sorts to avoid code duplication.

lib/drivers/bayerContourNext.js
pazaan: According to the Contour ASTM protocol doc, if a sequence number is out of order, or the checksum doesn't match at this point, we should send a NAK. At the moment, we're always just sending an ACK, and not caring about the sequence number. This doesn't seem like too much of a hassle with USB devices, but could be a problem later if we want to use this code for serial devices (where interference and retransmits are more of a factor).

lib/drivers/bayerContourNext.js
pazaan: We're validating the checksum only after we've done the communication with the USB device (after we've pushed the results onto our record array). I'll have a look and see how much work is involved in doing it "properly" ;)

lib/drivers/bayerContourNext.js
pazaan: The first record that the Contour device gives back is a Patient record 'P'. The rest are data record types 'R'. I'll update the data structure to make it more obvious.

lib/drivers/bayerContourNext.js
pazaan: Will update. Do you also want me to modify all the other callback(err, null) in the existing code?

gniezen: If they're in the `callback(err, null);` format they are correct.

lib/drivers/bayerContourNext.js
gniezen: :+1:

lib/drivers/bayerContourNext.js
pazaan: There's a try/catch around the calling function now, so this exception will be surfaced in the UI.

lib/drivers/bayerContourNext.js
```diff
@@ -271,110 +287,139 @@ module.exports = function (config) {
...
-            });
-          }
-          else if(rec.TYPE == ASCII_CONTROL.TER){
+        var packet = new Uint8Array(raw);
+
+        var packetHead = struct.unpack(packet, 0, '3Z2b', ['HEADER', 'SIZE', 'BYTE1']);
+
+        if(packetHead['HEADER'] !== MAGIC_HEADER){
+          debug('Invalid packet from Contour device');
+          callback(new Error('Invalid USB packet received.'), null);
```
gniezen: I was just testing the error handling by unplugging the device during an upload. You need to return the callback, otherwise it will still try to process and upload the invalid data. Also, remember to clear the timeout timer and interval timers.

lib/drivers/bayerContourNext.js
```diff
@@ -470,48 +514,63 @@ module.exports = function (config) {
...
+      var dataRecords = [];
+      var error = false;
+
+      async.whilst(
+        // Get records from the meter until we get the Message Terminator Record (L)
+        // The spec says that unless we get this, any preceding data should not be used.
+        function () { return (recordType !== 'L' && !error); },
+        function (callback) {
+          getOneRecord(data, function (err, result) {
+            recordType = result.recordType;
```
gniezen: If `getOneRecord` returns an error, this code will still try to process `result` first before handling the error. Also, remember to do a `return callback(err,null)`, with just a `callback(err,null)` it will still try to process and upload the invalid data.

gniezen:
Hi there! I'm sorry I didn't get a chance to merge this before going on holiday - this PR will now need a little bit of updating to be compatible with the redux work that got merged into master recently. The main change is that there is now a single device definition in [`redux/reducers/devices.js`](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/redux/reducers/devices.js) where previously the definitions were all over the place.

gniezen:
Thanks! I've successfully uploaded (and verified BG readings) of:

- Bayer Contour Next USB
- Bayer Contour Next
- Bayer Contour USB
- Bayer Contour Next Link

One that still need to be fixed after the merge: `lib/core/deviceInfo.js` and `lib/state/appState.js` has both been deprecated and were removed from master. Don't put them back! ;)

gniezen:
LGTM! Will try to merge as soon.

PR #225, opened 2016-01-12
Fix for with_normal bug

gniezen:
One-line change: `bolus.bolus_insulin_requested` should be `bolus.insulin_requested`.

### DIFF
--- a/lib/drivers/tandemDriver.js
+++ b/lib/drivers/tandemDriver.js
@@ -1482,7 +1482,7 @@ module.exports = function (config) {
 
       // bolus records
       if (bolus.bolex_size !== undefined || bolus.bolus_option === 'extended') {
-        if (bolus.bolus_size !== undefined || bolus.bolus_insulin_requested !== undefined) {
+        if (bolus.bolus_size !== undefined || bolus.insulin_requested !== undefined) {
           record = cfg.builder.makeDualBolus();
         }
         else {


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1KvSBkOq&#x2F;25-uploader-bug-tandem-with-normal-error"> [uploader][bug] Tandem with_normal error</a>

gniezen:
Ping @jebeck for review. Explanation and instructions to reproduce bug is on attached card.

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1KvSBkOq&#x2F;25-uploader-bug-tandem-with-normal-error"> [uploader][bug] Tandem with_normal error</a>

lib/drivers/tandemDriver.js
```diff
@@ -1482,7 +1482,7 @@ module.exports = function (config) {
 
       // bolus records
       if (bolus.bolex_size !== undefined || bolus.bolus_option === 'extended') {
-        if (bolus.bolus_size !== undefined || bolus.bolus_insulin_requested !== undefined) {
+        if (bolus.bolus_size !== undefined || bolus.insulin_requested !== undefined) {
```
jebeck: Hmm, this looked reasonable this morning. Now I don't know if I'm second-guessing or onto something. Why is the operator here an `||`? Does `bolus_size` only refer to the `normal` portion of a bolus (so we're looking for either an actual delivered `normal` bolus or a requested one that may have been cancelled before anything was delivered)? That's the only way I think I can make sense out of it.

gniezen: Yes, `bolus_size` only refers to normal portion, and that's how I made sense of it as well.

jebeck:
OK, LGTM, thanks.

PR #226, opened 2016-01-13
Add test script to conform to test mechanism for other node projects

darinkrauss:
@jebeck @jh-bate Add simple test script to run npm test, similar to user-api-client testing (which also uses mocha to run tests)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,3 @@
+#! /bin/bash -eu
+
+npm test


jh-bate:
LGTM

jebeck:
Same comment as on blip - this is adding another thing to maintain, and I don't understand what's wrong w/`npm test`.

PR #230, opened 2016-01-18
Fix for signed byte

gniezen:
@jebeck Found a bug in the new `extractSignedByte` function I put into `struct.js`. We're currently only using it for processing a `filteredRateByte` in the Dexcom G5, but as we don't know what it does yet we're not storing it. As such, it doesn't break anything (yet), but will be nice to push as part of the Dexcom changes if you have time to review.

### DIFF
--- a/lib/struct.js
+++ b/lib/struct.js
@@ -114,13 +114,10 @@ module.exports = function() {
     var extractByte = function(b, st) {
         return b[st];
     };
-    var extractSByte = function(b, st) {
+    var extractSignedByte = function(b, st) {
       // it's a signed 8-bit integer (called sbyte in c sharp)
-      var num = parseInt(b, 16);
-      if (num > 127) {
-        num = num - 256;
-      }
-      return num;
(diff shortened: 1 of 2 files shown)


lib/struct.js
```diff
@@ -398,7 +395,6 @@ module.exports = function() {
         extractShort: extractShort,
         extractFloat: extractFloat,
         extractByte: extractByte,
-        extractSByte: extractSByte,
```
jebeck: Looks like you're not exporting it anymore? Intentional?

Also there are tests for all the stuff in struct, I believe. Shouldn't you add a couple?

jebeck:
Fine to merge once the issues listed above (export and tests) are addressed.

lib/struct.js
gniezen: Good eye! Yes, intentional - the other signed versions aren't exported either.

Added tests to `testStruct.js`, which all pass - I should've done this the first time round!

PR #231, opened 2016-01-20
Changing error message for no records available

gniezen:
One-liner to update error message when there are no records on the Dexcom receiver. Pinging @jebeck for review and merge. As v0.242.0 is currently a RC, I'm ok with this going out in the next release.


jebeck:
LGTM

PR #232, opened 2016-02-02
Update message seen when usb cable bad

cheddar:
When I tried to upload my data, I had a cable that apparently only provides power and does not transfer signal.  After a bunch of troubles, I found the entry on http://support.tidepool.org/article/13-upload-your-dexcom-g4 about trying another USB cable and that fixed the problem.  Having that as an option in the error message would've likely been a nicer experience.

I would've preferred to add the comment at the end of the message, but it appears like the error handling is forcing the link to the device driver to be at the end of the message.  I'm not necessarily in love with the wording or anything here, if you think it's suitable great.  If not, feel free to do something else that is more meaningful and just close this PR.


jebeck:
Hey @cheddar thanks for pointing this out. The surfacing of errors in the UI is being pretty heavily revamped in [jebeck/redux](https://github.com/tidepool-org/chrome-uploader/tree/jebeck/redux), which should be going into review and then through testing very soon. I'll check in with @gniezen and try to figure out a way to work in a "maybe your cable is bad" message.

cheddar:
Fwiw, as I thought about it some more, just linking to the Support portion of the tidepool.org website might actually be preferable?

jebeck:
Hmm, yes, that's a good idea @cheddar! I bet we could easily configure a link to the relevant section on a per-driver basis :)

PR #233, opened 2016-02-11
unbork the node CLI tools for uploading

jebeck:
I borked the CLI tools again due to ES6 deps on the CLIs' `require` path /o\

Probably no one should merge this until I figure out how to add a couple of tests to make sure I don't keep borking the tools. Should be pretty easy.


jebeck:
@gniezen when you get back can you look at/test this and merge if it looks good? I borked the CLI tools again... /o\

gniezen:
LGTM! The only thing I noticed was that the version message in the IBF loader CLI tool should say Insulet, not CareLink. I've already made the change and will now merge.

PR #234, opened 2016-02-12
replace async.timesSeries with for loop + async.until

jebeck:
The regression that caused only one data point to upload for Bayer meters was due to a change in the implementation of `async.timesSeries`, a dependency that was updated as part of the redux work.

This PR replaces the use of `async.timesSeries` with a for loop and `async.until`.


jebeck:
As as @GordyD pointed out, the root cause here is the fact that `async.timesSeries` can no longer handle the _string_ "8" as the first arg. The simpler #235 replaces this PR.

PR #235, opened 2016-02-12
just parse the string into an integer and all is well

jebeck:
Necessary because `timesSeries` method in new version of `async` won't coerce a _string_ as a first arg to an int!


GordyD:
Cool!

GordyD:
LGTM

PR #236, opened 2016-02-17
set uploadTargetUser properly 

jebeck:
There was a bug in the first production release (v0.251.0) of the redux migration. It involved not setting an `uploadTargetUser` in the case of non-PWD `loggedInUser` with upload access to exactly one other account.

This fixes the problem and updates the tests.

In the course of comparing against prod (i.e., v0.242.0) to understand the problem, I discovered another inconsistency (showing vs. not a user dropdown with only one user in the same case). That will be fixed here too.

Ping @GordyD @krystophv

### DIFF
--- a/lib/redux/reducers/users.js
+++ b/lib/redux/reducers/users.js
@@ -222,11 +222,14 @@ export function uploadTargetUser(state = null, action) {
     case actionTypes.LOGIN_SUCCESS:
     case actionTypes.SET_USER_INFO_FROM_TOKEN:
       const { user, profile, memberships } = action.payload;
+      const uploadMemberships = _.filter(memberships, (mship) => {
+        return !_.isEmpty(_.get(mship, ['profile', 'patient']));
+      });
       if (!_.isEmpty(profile.patient)) {
         return user.userid;
       }
-      else if (memberships.length === 1) {
-        return memberships[0].userid;
+      else if (uploadMemberships.length === 1) {
+        return uploadMemberships[0].userid;
       }
       else {
         return null;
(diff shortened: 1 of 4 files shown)


lib/redux/reducers/users.js
```diff
@@ -222,11 +222,14 @@ export function uploadTargetUser(state = null, action) {
     case actionTypes.LOGIN_SUCCESS:
     case actionTypes.SET_USER_INFO_FROM_TOKEN:
       const { user, profile, memberships } = action.payload;
+      const uploadMemberships = _.filter(memberships, (mship) => {
+        return !_.isEmpty(_.get(mship, ['profile', 'patient']));
+      });
```
jebeck: The basic misunderstanding was that the memberships _always_ includes the `loggedInUser`, even if they aren't a PWD! The mistake slipped in when I refactored the `users` reducer to reduce (har har) the nesting...

GordyD: Cool. Didn't know you could use an array of terms **or** a string to represent path to variable you want to `_.get()` e.g. 'profile.patient'.

GordyD:
Have tested Chrome Uploader that points to Devel with an account that can upload for 1 other patient and is not a PWD.

I am able to choose devices by specifying patient (only 1 choice), choose timezone and then device. The button is no longer greyed out and I can save.

lib/redux/reducers/users.js
jebeck: Yep. I'm probably failing to be consistent about which one I use... /o\

PR #238, opened 2016-02-18
pass the correct built object to the annotation utility

jebeck:
This is in response to a bug reported by Jaeb. What's a bit odd about it is the fact that the reported error "Cannot read property `annotations` of undefined" and reported situation (meter only has one ketone reading stored) should only turn up when there's a _ketone_ reading with an annotation, and the only time that _ketone_ readings get annotated is when the value is `HI`, which I would have assumed is very very rare in the wild.

Also, since we don't have a way to produce a `HI` on a test device, we may not be able to confirm this fix directly.

### DIFF
--- a/lib/drivers/abbottPrecisionXtra.js
+++ b/lib/drivers/abbottPrecisionXtra.js
@@ -479,7 +479,7 @@ module.exports = function (config) {
           .done();
         if (datum.annotations) {
           _.each(datum.annotations, function(ann) {
-            annotate.annotateEvent(smbg, ann);
+            annotate.annotateEvent(bloodKetone, ann);
           });
         }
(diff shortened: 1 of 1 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;62LiJIl6&#x2F;14-uploader-triaging-fixing-abbott-precision-xtra-issue-reported-by-jaeb"> [uploader] triaging&#x2F;fixing Abbott Precision Xtra issue reported by Jaeb</a>

jh-bate:
LGTM

gniezen:
As we can't confirm this fix directly, I'm going to merge, to be released as part of BtUTC fix update.

PR #240, opened 2016-02-29
hot module reload

krystophv:
Enable hot module reloading for JS/CSS in the uploader.


package.json
```diff
@@ -69,6 +70,8 @@
     "redux-mock-store": "0.0.6",
     "salinity": "0.0.8",
     "style-loader": "0.13.0",
-    "webpack": "1.12.12"
+    "webpack": "1.12.12",
+    "webpack-dev-server": "^1.14.1",
```
jebeck: Please lock down these dependencies - we don't use `^`. (You can do `npm install --save-exact` in future to avoid having to remember to change in the package.json.)

jebeck:
As mentioned above, please lock down the dependencies in the package.json and add an explanation of that weird warning (about the checked in key file) in the README. Otherwise LGTM, go ahead and merge @krystophv

krystophv:
@jebeck - Could I get a quick opinion about the summary of the warning Chrome throws on the unpacked extension? Is it clear enough?

jebeck:
Explanation of the warning looks good @krystophv, but want to take another look at point 10 on the README? Probably should change to explain hot-reloading. I'd put something like "React components and CSS will hot load after changes, but changes to device drivers and other code outside of the React components will not - use 'Reload' from chrome://extensions to reload after such changes."

krystophv:
@jebeck does https://github.com/tidepool-org/chrome-uploader/commit/044b3eddcc435c345ba7f7f37b1722db20e3a638 look good?

krystophv:
Yeah, webpack won't die on bundle errors. There should be a way to get it to not issue the hot-update either (and thus potentially not need the App refresh - see https://github.com/gaearon/react-hot-boilerplate/issues/1 ) but I couldn't get that implementation working properly, likely due to the write to disk plugin.

PR #241, opened 2016-02-29
initial proof of concept for CSS modules

krystophv:
PROOF OF CONCEPT: example of migration of a component to CSS Modules

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -38,7 +38,8 @@ var config = {
     loaders: [
       { test: /\.js$/, exclude: /(node_modules)/, loader: 'babel-loader' },
       { test: /\.jsx$/, exclude: /(node_modules)/, loader: 'babel-loader' },
-      { test: /\.less$/, loader: 'style!css!less' },
+      { test: /\.module\.less$/, loader: 'style?sourceMap!css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]!less?sourceMap' },
+      { test: /^((?!module).)*\.less$/, loader: 'style!css!less' },
       { test: /\.json$/, loader: 'json' }
     ]
   },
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -38,7 +38,8 @@ var config = {
     loaders: [
       { test: /\.js$/, exclude: /(node_modules)/, loader: 'babel-loader' },
       { test: /\.jsx$/, exclude: /(node_modules)/, loader: 'babel-loader' },
-      { test: /\.less$/, loader: 'style!css!less' },
+      { test: /\.module\.less$/, loader: 'style?sourceMap!css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]!less?sourceMap' },
```
jebeck: Ah I see so adding `.module` would be a temporary naming convention while porting?

jebeck:
What's your feeling on class names if we port? One of the things I like about the local scoping of CSS modules is the ability to get rid of the long and cumbersome SuitCSS-style class names, but wondering if that's a feeling you share @krystophv

webpack.config.js
krystophv: Yeah. I just needed some way to make sure that they only got processed by one loader chain or the other. I had to drop the !module regex on the standard less loader in order to prevent it from processing stuff already loaded by the module loader here.

krystophv:
I'm not a fan of the SuitCSS/BEM class naming convention. It just never sat particularly well with me. If we're going with modules, I'd much prefer to have shorter, nicer class names. I just left them the way they were to get the proof of concept while modifying as little as possible so the diff ends up showing precisely what changes to get CSS Modules going and nothing else.

jebeck:
Ok, awesome re: naming conventions - sounds like we are on the same page.

jebeck:
@krystophv go ahead and close this and delete the branch (at least from upstream) to clean up whenever you feel like you don't need it for reference anymore.

krystophv:
Closing as this effort is moving to a new branch to port the whole app to CSS modules. Relevant commit will be cherry-picked into the new branch.

PR #242, opened 2016-02-29
hot reload doAppInit fix

jebeck:
Take a look at this @krystophv and let me know if (a) it works for you locally and (b) it makes sense to you.

@GordyD and I considered two approaches, with the other possible approach being to look at the `working.initializingApp` flag in `App.js` itself and only call the `doAppInit` in the constructor if the flag is `true` but since we already have test instrumentation around the async actions (thunks), it seemed better to put the logic in the `doAppInit` thunk itself (although the test did turn out to be hacky 😕).

### DIFF
--- a/lib/redux/actions/async.js
+++ b/lib/redux/actions/async.js
@@ -41,6 +41,12 @@ let daysForCareLink = null;
 
 export function doAppInit(opts, servicesToInit) {
   return (dispatch, getState) => {
+    // when we are developing with hot reload, we get into trouble if we try to initialize the app
+    // when it's already been initialized, so we check the working.initializingApp flag first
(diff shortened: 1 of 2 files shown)


krystophv:
This (a) works for me locally and (b) makes sense. 

I think it's actually probably a good idea for the `doAppInit` to check to make sure that the app hasn't already been initialized before running, since the app should only ever be initialized once. The setTimeout hack hurts me at some level (much :unamused: since I can't think of a better solution and I hate it when that happens, but at least it's not a setTimeout zero, right?), but I think it'll probably be fine since we're testing a condition that should only happen during development. 

LGTM. I'll merge it.

PR #243, opened 2016-03-01
Krystophv/unflex

krystophv:
An effort to remove the global div/span flex styles and apply the flex layout options only to the relevant pieces of the application.


krystophv:
@jebeck could you take a look at this one. I'd like to get this in before migration of everything to CSS modules.

styles/components/App.less
```diff
@@ -15,12 +15,16 @@
 
 .App {
   height: 100%;
+  display: flex;
+  flex-direction: column;
```
jebeck: Seems like a mixin for the combo of these two props would be good, but maybe not now but rather as part of the CSS modules port/single resp refactor?

styles/components/VersionCheck.less
```diff
@@ -56,7 +56,7 @@
       // (effectively block)
       // TODO: fix that at the source
       span {
-        display: inline;
+        // display: inline;
```
jebeck: Remove the commented out line plz?

styles/core/dropdowns.less
```diff
@@ -3,7 +3,7 @@
 // because we're setting display: flex on every div
 // and it turns out that interferes with text-overflow :(
 .Select-placeholder {
-  display: block;
+  //display: block;
```
jebeck: Ditto.

styles/core/scaffolding.less
```diff
@@ -17,7 +17,7 @@
 // ====================================
 
 div, span {
-  .box-flex();
+  //.box-flex();
```
jebeck: And ditto. :wink:

styles/components/App.less
krystophv: Yeah, likely as part of a layout.css that we can compose from.

jebeck:
:+1: Also gave it a quick visual spot check locally.

krystophv:
https://trello.com/c/mg3NudYP/23-tech-debt-uploader-remove-display-flex-global-style-and-fix-what-breaks

PR #245, opened 2016-03-09
krystophv/hmr improvements

krystophv:
This is an attempt to get all of the application code to hot-reload (as opposed to just the CSS and React) by allowing the AppInit to complete almost as normal, just omitting the syncActions that force a state refresh and trip up the React cycle. Also reverting the `npm start` to what it was previously and adding a new `npm run start-hot` to differentiate and allow for non-HMR development.

This needs some tire kicking and relatively thorough testing to ensure that the re-initialization of all the extra pieces of the app doesn't muck anything up.


jebeck:
OK so having looked at this again, what I'd really like to understand before we move forward with this strategy is why only `initRequest` and `hideUnavailableDevices` seem to trigger the state transition during render effect. There are *lots* more things happening in the init flow that also should be triggering `setState`s. When you run as it is here, do you see the network requests lower in the init flow happening on each hot update?

krystophv:
This is _kinda_ moot with the impending Electron port.

PR #246, opened 2016-03-11
specify node version we're on to patch level; allow use of Travis containerized infrastructure

jebeck:
FYI @gniezen @krystophv

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
+sudo: false
+
 language: node_js
 node_js:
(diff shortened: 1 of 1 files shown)


gniezen:
@jebeck Cool! So can we verify if this PR is now running on the containerised architecture? Or did you notice that it ran a lot faster?

jebeck:
We can: 
screenshot 2016-03-11 08 09 13

(From clicking through to Travis and inspecting the build console.)

PR #249, opened 2016-03-14
CSS Modules and new design

krystophv:
Implementing CSS modules for the chrome uploader. All the component specific less files moved to CSS modules with updated class names. A few new single-responsibility css modules (typography, layout) to organize styles that are composable in multiple pieces of the application. Updating some of the styles (mostly colors) to match the new in-clinic design work.


krystophv:
https://trello.com/c/WxzA2pUm

jebeck:
Looking through running app locally first to hunt down regressions (I'm comparing against the clinic uploader flow document in Zeplin):

- sign-up icon missing and link color

- dropdown menu misaligned:
screenshot 2016-03-14 14 26 19

- link colors generally (e.g., 'Forgot your password?' - the hover color is being used for the main color; this also affects the "Error details" link in the upload page, among other places

- underline on driver links from Choose Devices is missing (I've asked Sara about this one in Zeplin):
screenshot 2016-03-14 13 15 35

- regression on upload page in padding and alignment of Go to Blip button
screenshot 2016-03-14 14 23 20

krystophv:
@jebeck: re: link colors - from the new design, most links (excepting the logged in user and feedback link) have had their interaction colors switched per the styleguide such that the normal state is now light and the hover state is now dark. is this what you're commenting on?

re: Go to Blip button - the new design has all buttons lined up against the right edge of their container. do you think that'd be reasonable to carry over to this button or go back to centered for the time being? (this might be a @skrugman question)

jebeck:
Biggest problem here so far is some real weirdness on initial launch and render - things render first in bizarre positions and then move into place. I've replicated with a real build locally (`npm run build` followed by loading the extension from `dist/`), so it's not just an issue in the webpack dev env.

jebeck:
Letters that drop below baseline cut off in CareLink inputs...make the inputs a bit bigger perhaps?
screenshot 2016-03-21 15 04 49

jebeck:
This should be on two lines (the part in parens on the second):
screenshot 2016-03-21 15 06 11

jebeck:
I think I'm even more confused than I was before re: the naming of .less files (with `module` in the name or not). Don't see a doc in the `docs/` dir. Are you still working on it? Would be very useful as soon as is possible...

jebeck:
Long names run into the T: 😜
screenshot 2016-03-21 15 23 11

jebeck:
The letters than drop below baseline issue also affects the login input...

krystophv:
image
Strange. I don't experience the input cutoff issue.

krystophv:
For future reference, the input clipping was caused by a line-height bug in Chrome 49 (I was running Dev 50.x) that causes clipping between line-height 17px and 21px for a font-size of 16px. 

The issue with loading I'm afraid may be an artifact of using CSS modules. I'm fairly certain that it's because the CSS is inlined in the javascript and dynamically inserted into the document at render time for a component. It seems that the DOM render is faster than the CSS re-flow so the components show up unstyled and then the the styles get applied. This might be something we can solve with the extract text Webpack plugin - I'll look into it.

The line break for details, long name logo collision, docs and line height issues should now be resolved.

jebeck:
Yes, we had similar issues (although less dramatic) with inlined CSS in blip (not CSS modules, of course, but just regular webpack bundling), and they were solved by the extract text - please give that a try @krystophv.

docs/misc/CSSModules.md
```diff
@@ -0,0 +1,17 @@
+## CSS modules
+
+### CSS what?
+For a quick breakdown on what CSS modules are, check out [Glen Maddern's post ](http://glenmaddern.com/articles/css-modules) which summarizes pretty well.
+
+### How they're used here
+CSS modules allow for module-scoped styling for components. This is done by `require`ing the styles file and allowing the [Webpack css loader](https://github.com/webpack/css-loader) to process the file as a module. Since there are some styles that need to remain global (mainly webfonts and styles for 3rd party components), we differentiate loaders by naming convention of `.module.less` vs `.less`.
```
jebeck: Nitpick, but in the Webpack loader link, capital CSS please?

And clarify "differentiate loaders" to something more like "differentiate files using the modules option with the webpack CSS loader from those being loaded normally" (feel free to adjust for accuracy).

docs/misc/CSSModules.md
```diff
@@ -0,0 +1,17 @@
...
+
+```css
+.awesomeClass {
+  composes: neatStyle from '../coolText.less'
+}
+```
+
+In general, all of the components will have a `.module.less` as the styles are `require`d from within the JavaScript of the component. The files under `/styles/core` are a slightly mixed bag as they aren't generally `require`d individually from a JavaScript file (excepting `main.less`). The files in `/core` marked as modules don't use any module-incompatible LESS syntax features and are intended to be used as single-responsibility modules solely through the `composes` keyword from other `.module.less` files. The plain `.less` files can be used through `composes`, but for one reason or another require the ability to be used in a global context in addition to a scoped context and therefore cannot be exclusively loaded by the Webpack css module loader.
```
jebeck: Another nit on capitalizing CSS right at the end. Otherwise, looks good! Helpful w/o being too wordy 👍

krystophv:
@jebeck I think that takes care of all of the outstanding issues, if you wanted to give it another look over.

jebeck:
OK, @krystophv aside from the error states, LGTM.

I realized that a small regression that had come in on master (when I rearranged some constants files for the redux implementation to more closely match blip's) was larger than I thought, and I fixed it in #258, which you should pull into this branch before you do the work to fix the surfacing of error states as described below. I also added some documentation to help you trigger the error states locally in [this docs PR](https://github.com/tidepool-org/docs/pull/4).

Here are some screenshots for reference re: what the 3 error states look like (this branch followed by v0.253.0 when they worked). Mainly it's the size & positioning of text and mostly opaque background for the overlays that needs to be fixed. You may want to check in with @skrugman re: font sizes since I know those changed overall in the redesign.

1. Offline error (which you should be able to reproduce by loading the uploader against a non-local API (prod or dev or whatever) and turning off your WiFi):
screenshot 2016-03-23 14 46 59
screenshot 2016-03-23 14 48 33

2. Out-of-date uploader:
screenshot 2016-03-23 16 36 37
screenshot 2016-03-23 16 10 02

3. Error when trying to check (against jellyfish's /info endpoint) for out-of-date uploader:
screenshot 2016-03-23 16 37 13
screenshot 2016-03-23 16 08 35

styles/components/VersionCheck.module.less
```diff
@@ -0,0 +1,83 @@
+@import '../core/variables.less';
+
+.purple-centered {
+  color: @purple-dark;
+  text-align: center;
+}
+
+.mostly-opaque-background {
```
jebeck: This name isn't camel-cased/consistent with the others. (It also appears to not be in effect in the error states, so perhaps this is the cause?)

styles/components/VersionCheck.module.less
```diff
@@ -0,0 +1,83 @@
+@import '../core/variables.less';
+
+.purple-centered {
```
jebeck: Ditto re: camel-casing...

styles/core/forms.less
```diff
@@ -13,25 +13,21 @@
...
 
-// Form controls
-// ====================================
+@import 'variables.less';
 
-@form-control-border-size:   2px;
-@form-control-height:        (30px + 2*@form-control-border-size);
-@form-control-border-color:  @purple-light;
 
 .form-control-border {
```
jebeck: The names here also haven't been updated to the camel-cased system. Intentional? If not, perhaps update?

krystophv: The forms.less file is still depended on by the dropdown.less in a traditional LESS way, so I didn't want to go changing all the classnames unnecessarily. The forms are also never directly required by a JS file, so the advantage of being able to reference the class name directly as a simple object property via the dot syntax vs using the array index string format isn't really there. I can add something in the doc concerning that.

jebeck: OK yeah, please add a brief section to the doc re: naming conventions for the CSS modules stuff vs. the old stuff.

jebeck:
Good addition to the docs re: naming, thanks!

:shipit: is merging now.

PR #251, opened 2016-03-15
Using Data Services API for basals and suspend/resume events

gniezen:
This uses the new Data Services API to send basals and suspend/resume events. Note that this requires this platform-client PR to work: tidepool-org/platform-client#61


gniezen:
Given the changes to other areas of the Uploader, I thought it'd be good for you to review this before merging to the main Animas branch @jebeck ?

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pkyUK1Uy&#x2F;2-ux-0-dev-5-tst-0-uploader-send-animas-basals-and-suspend-resume-events-with-crud-api"> &lt;ux:0 dev:5 tst:0&gt; [uploader] Send Animas basals and suspend&#x2F;resume events with CRUD API</a>

.config.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
   // the defaults for these need to be pointing to prod
   API_URL: process.env.API_URL || 'https://api.tidepool.org',
   UPLOAD_URL: process.env.UPLOAD_URL || 'https://uploads.tidepool.org',
+  DATASERVICES_URL: process.env.DATASERVICES_URL || 'https://dataservices.tidepool.org',
```
jebeck: I believe this needs to be updated, according to @darinkrauss's response in Slack, no?

lib/animas/animasSimulator.js
```diff
@@ -98,6 +98,12 @@ exports.make = function(config){
           annotate.annotateEvent(event,'animas/out-of-sequence');
         }
 
+        if((currBasal.rate === 0) && (currBasal.deliveryType !== 'suspend')) {
+          // As only the 30 most recent suspend/resume events are stored,
+          // we can't be sure if older basals with 0.0 rate are suspends or not
```
jebeck: Per your comment, would it be possible to tell if we're looking at an "older" basal, so we're not annotating the "newer" basals unnecessarily?

lib/core/api.js
```diff
@@ -252,12 +256,22 @@ function postBlockToPlatform(data, groupId, blockIndex, callback) {
...
   api.log('postBlockToPlatform #' + blockIndex + ': using id ', groupId);
 
-  tidepool.uploadDeviceDataForUser(groupId, data, function(err, result) {
-    if (err) {
-      return sad(err, err);
-    }
-    return happy(result);
-  });
+  if(dataservices) {
+    tidepool.uploadDataServicesDataForUser(groupId, data, function(err, result) {
```
jebeck: It took me a while to read this correctly. Do I understand that `data` might be regular data or dataservices data and that `dataservices` in the function args is a boolean indicating which call to make? I'd vastly prefer a factory to produce a `postBlockToPlatform` function for each API, then just use each as appropriate. That way is still DRY but also w/o the if/else logic and confusing (& long) list of args.

lib/core/api.js
```diff
@@ -279,11 +293,17 @@ api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb) {
...
+
+  var post_and_progress_dataservices = function (data, callback) {
+    progress(nblocks++ * 100.0 / blocks.length);
+    //off to the dataservices platfrom we go
+    return postBlockToPlatform(data, groupId, nblocks, true, callback);
   };
 
   var post_upload_meta = function (uploadMeta, callback) {
-    return postBlockToPlatform(uploadMeta, groupId, 0, callback);
+    return postBlockToPlatform(uploadMeta, groupId, 0, false, callback);
```
jebeck: And here this raises an additional question for @jh-bate and @darinkrauss. How are we going to enforce the minimum uploader version with the new "dataservices" API? Do we want to post a second upload metadata? I actually think this would be a good idea as it also relates to the problem I was posing in Slack earlier: how will we be able to trace data (for auditing purposes) back to either jellyfish or the new dataservices? If we do a separate upload metadata POST for each (and include information about where we're posting *to* in the metadata, in a new field of some sort), then we'll have that link easily via the `uploadId` (as long as we don't allow a situation to arise again where we end up not storing upload metadata even when the rest of the POSTs go through 😉).

lib/core/api.js
```diff
@@ -360,6 +380,53 @@ api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb) {
   });
 };
 
+/*
+ * process the data sending it to the dataservices platform in blocks and feed back progress to the calling function
+ */
+api.upload.toDataServices = function(data, sessionInfo, progress, groupId, cb) {
```
jebeck: ...and here can we also make a factory since this is nearly identical to `api.upload.toPlatform`?

main.js
```diff
@@ -79,21 +79,25 @@ function setServer(window, info) {
...
     Staging: {
       API_URL: 'https://stg-api.tidepool.org',
       UPLOAD_URL: 'https://stg-uploads.tidepool.org',
+      DATASERVICES_URL: 'https://stg-dataservices.tidepool.org',
       BLIP_URL: 'https://stg-blip.tidepool.org'
     },
     Production: {
       API_URL: 'https://api.tidepool.org',
       UPLOAD_URL: 'https://uploads.tidepool.org',
+      DATASERVICES_URL: 'https://dataservices.tidepool.org',
```
jebeck: Again, will need to be updated I believe...

lib/core/api.js
gniezen: The only real difference is whether we're calling `tidepool.uploadDataServiceDataForUser` or `tidepool.uploadDeviceDataForUser`. I've refactored it to remove the confusing if statements based on the boolean variable, and pass in the `uploadType` instead, that is then resolved to the correct function using `getUploadFunction()`. Does that make the code a bit clearer?

lib/core/api.js
gniezen: As part of the refactoring I've removed `api.upload.toDataServices`, passing in `uploadType` into `api.upload.toPlatform` instead to keep things DRY.

lib/animas/animasSimulator.js
gniezen: I think it's as easy as checking whether `currStatus === null`?

lib/core/api.js
```diff
@@ -263,9 +278,10 @@ function postBlockToPlatform(data, groupId, blockIndex, callback) {
 /*
  * process the data sending it to the platform in blocks and feed back progress to the calling function
  */
-api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb) {
+api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb, uploadType) {
```
jebeck: As discussed earlier, please just expand on the comment here to explain that the `uploadType` is the final arg (instead of the callback) so that calls to `api.upload.toPlatform` don't have to be modified in _every_ driver...

lib/drivers/animasDriver.js
```diff
@@ -1786,8 +1785,23 @@ module.exports = function (config) {
             progress(100);
             return cb(err, data);
           } else {
-            progress(100);
-            return cb(null, data);
+            progress(50);
+
+            cfg.api.upload.toPlatform(
+              data.post_dataservices,
```
jebeck: The two variables here aren't great, name-wise: `data.post_records` vs. `data.post_dataservices`... Perhaps you could come up with something clearer/more parallel to how you refer to the two APIs in api.js?

jebeck:
After comments from today addressed, good to merge into gniezen/animas. Then perhaps open a WIP PR for the Animas driver from that branch and add some instructions about checking out the gniezen/data-services branch of platform-client in order to use/test (via `npm link`)? Took me a bit to figure that out (app was hanging on loading because of problems initializing the dataservices API).

lib/drivers/animasDriver.js
gniezen: Sure! As I'm using 'uploads' and 'dataservices' for `uploadType`, I'll use `data.post_uploads` and `data.post_dataservices`.

lib/core/api.js
gniezen: Done.

gniezen:
@jebeck I'm going to hold off on merging this branch until the Data Services API is up and running on staging and I know what the production URL will be. If I merge it now, making fixes to the current Alpha Uploader (v0.254.1) will be challenging.

jebeck:
Oh true, like the mock server I had to set up. Damn, this branch is sure gonna hang out for a looooong while then...

gniezen:
As @darinkrauss has now re-implemented connecting to the data services API in `darinkrauss/platform-endpoints` branch, we can close this PR.

PR #254, opened 2016-03-18
fix misleading 'Using the default environment' message

jebeck:
The logic to determine whether to show you the "Using the default environment, which is now production" message would show you the message even if you had one or more - just not _all_ - variables set in your local environment. Now the message will only display if you have _none_ of the variables set, which is how it should be when building for production.

@gniezen please review and update your #251 accordingly.

CC @darinkrauss

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,7 +19,7 @@ if (process.env.DEBUG_ERROR === 'true') {
   console.log();
 }
 
-if ((!process.env.API_URL || !process.env.UPLOAD_URL || !process.env.BLIP_URL)) {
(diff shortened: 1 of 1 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;XdWKknvr&#x2F;24-uploader-fix-misleading-message-re-building-for-production"> [uploader] fix misleading message re: building for production</a>

darinkrauss:
LGTM :shipit:

gniezen:
LGTM too, merging.

PR #255, opened 2016-03-18
Animas zero-rate basal marked as suspended due to alarm

gniezen:
If we have a zero-rate basal that is not already marked as suspended, we look back two hours to see if we have any recent alarms that could have caused the zero-rate basal. If so, we mark it as suspended and remove the alarm (so that we don't mark any other zero-rate basals with the same alarm).

We then also generate a suspend event (at the time the basal rate changed to zero) and a resume event (when the basal rate changes again).

@jebeck please review when you get a chance and test with your data to see if it works as expected.

### DIFF
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -196,6 +196,7 @@ exports.make = function(config){
     bolus: function(event) {
       delete event.syncCounter; //wizard events already synced up
       delete event.requiredAmount;
+      delete event.jsDate;
       simpleSimulate(event);
       setCurrBolus(event);
     },
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -123,7 +123,6 @@ exports.make = function(config){
           };
 
           ensureTimestamp(status);
-          dataservicesEvents.push(status);
           setCurrStatus(status);
           setLastAlarm(null); //reset alarm as not to re-use for other zero-rate basals
           setCurrBasal(basal);
@@ -176,23 +175,14 @@ exports.make = function(config){
           else {
             // basal is non-zero, but status still suspended
             // is this a suspend event we created from an alarm?
-            // If so, we should create a corresponding resume event
+            // If so, we should submit a joined suspend/resume event
             if(currStatus.reason.suspended === 'automatic') {
-              var status = {
-                time: event.time,
-                deviceTime: event.deviceTime,
-                timezoneOffset: event.timezoneOffset,
-                conversionOffset: event.conversionOffset,
-                deviceId: event.deviceId,
-                type: 'deviceEvent',
-                subType: 'status',
-                status: 'resumed',
-                reason: {resumed: 'automatic'}
-              };
 
-              ensureTimestamp(status);
-              dataservicesEvents.push(status);
-              setCurrStatus(status);
+              currStatus.duration = currBasal.duration;
+              currStatus.reason.resumed = 'automatic';
+
+              dataservicesEvents.push(currStatus);
+              setCurrStatus(null);
             }
           }
         }
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -91,7 +91,7 @@ exports.make = function(config){
   }
 
   function markIfBasalSuspendedByAlarm(alarm,basal) {
-    if ((basal !== null) && (alarm != null) &&
+    if ((basal != null) && (alarm != null) &&
         (basal.rate === 0) && (basal.deliveryType !== 'suspend')) {
 
       // Animas does not generate suspend/resume events for alarms,
@@ -99,7 +99,6 @@ exports.make = function(config){
       // If so, we mark the basal as suspended and generate a new suspend event.
 
       var duration = Date.parse(basal.time) - Date.parse(alarm.time);
-      console.log(alarm, basal);
       var twoHours = (120 * sundial.MIN_TO_MSEC);
       if(duration < twoHours) {
 
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -181,6 +181,7 @@ exports.make = function(config){
 
               currStatus.duration = currBasal.duration;
               currStatus.reason.resumed = 'automatic';
+              annotate.annotateEvent(currStatus, 'animas/status/fabricated-from-alarm');
 
               dataservicesEvents.push(currStatus);
               setCurrStatus(null);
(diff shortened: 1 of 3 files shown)


lib/animas/animasSimulator.js
```diff
@@ -85,9 +90,53 @@ exports.make = function(config){
     events.push(e);
   }
 
+  function checkIfBasalSuspendedByAlarm(alarm,basal) {
```
jebeck: Since this function mutates the basal, can we call it something else? `checkIfBasalSuspendedByAlarm` made me think it was going to return `true` or `false`. Maybe `markIfBasalSuspendedByAlarm`?

lib/animas/animasSimulator.js
```diff
@@ -85,9 +90,53 @@ exports.make = function(config){
...
+    if ((basal !== null) && (alarm != null) &&
+        (basal.rate === 0) && (basal.deliveryType !== 'suspend')) {
+
+      // Animas does not generate suspend/resume events for alarms,
+      // so we check if an alarm occurred recently or at the same time.
+      // If so, we mark the basal as suspended and generate a new suspend event.
+
+      var duration = Date.parse(basal.time) - Date.parse(alarm.time);
+      console.log(alarm, basal);
```
jebeck: Remove this logging? Or switch to using bows and/or put it behind __DEBUG__ if you think you want to keep it indefinitely for debugging purposes?

lib/animas/animasSimulator.js
```diff
@@ -85,9 +90,53 @@ exports.make = function(config){
     events.push(e);
   }
 
+  function checkIfBasalSuspendedByAlarm(alarm,basal) {
+    if ((basal !== null) && (alarm != null) &&
```
jebeck: Any reason for the first check here using `!==` but the second using `!=`?

jebeck:
Random bug report, but this one confused me for a while. The `deviceTime` on boluses (either all boluses or maybe just the ones embedded in the `wizard`s) is wrong! It's the same as `time` without the `.000Z`.

lib/animas/animasSimulator.js
```diff
@@ -85,9 +90,53 @@ exports.make = function(config){
...
+        var type = alarm.alarmType;
+        if(type === 'occlusion' || type === 'auto_off' || type === 'no_insulin' || type === 'no_power') {
+
+          //this basal is suspended
+          basal.deliveryType = 'suspend';
+          annotate.annotateEvent(basal, 'animas/basal/marked-suspended-from-alarm');
+
+          var status = {
+            time: basal.time,
+            deviceTime: basal.deviceTime,
```
jebeck: This `deviceTime` is wrong too - same as `time` but w/o `.000Z`.

lib/animas/animasSimulator.js
```diff
@@ -119,13 +168,30 @@ exports.make = function(config){
...
+            //this basal is suspended
+            event.deliveryType = 'suspend';
+          }
+          else {
+            // basal is non-zero, but status still suspended
+            // is this a suspend event we created from an alarm?
+            // If so, we should submit a joined suspend/resume event
+            if(currStatus.reason.suspended === 'automatic') {
+
+              currStatus.duration = currBasal.duration;
```
jebeck: I'm not sure where the logic is going wrong here, but testing this out locally with my `no_power` alarm and subsequent suspend from yesterday, the suspended basal and the suspended deviceEvent are both getting a duration of 360000 - maybe somehow based on the scheduled basal? Not sure, but in this case my suspend only lasted from 22:56 to 23:02 so incorrect.

lib/animas/animasSimulator.js
```diff
@@ -119,13 +168,30 @@ exports.make = function(config){
...
-          event.deliveryType = 'suspend';
+        if(currStatus.status === 'suspended') {
+          if (event.rate === 0) {
+            //this basal is suspended
+            event.deliveryType = 'suspend';
+          }
+          else {
+            // basal is non-zero, but status still suspended
+            // is this a suspend event we created from an alarm?
+            // If so, we should submit a joined suspend/resume event
```
jebeck: I think I might be a little confused by this comment and all the branches flying around. It's not just these deviceEvent suspend/resumes that will need to be combined into a single event for upload to the dataservices API...so shouldn't it be done in common somewhere? Maybe you're already reworking this somewhere else?

jebeck:
Another random bug: all your basals still have `index` on them - please remove (and preserve only in the `payload.logIndices`).

jebeck:
Other than comments above, this did look good for my `no_power` alarm from yesterday :+1: All the annotations were added as expected, etc.

lib/animas/animasSimulator.js
gniezen: Nope, typo. Thanks for catching it!

lib/animas/animasSimulator.js
gniezen: @jebeck I think the issue is not with `deviceTime`, but with `time`. I wasn't sending the most recent record to `TZOUtil`, which meant that `time` and `deviceTime` would always be the same. I'm now sending the most recent event time (final suspend) to `TZOUtil` which should fix this issue. However, while investigating `deviceTime` I did find one place where it is incorrect: manual smbg events.

lib/animas/animasSimulator.js
gniezen: Yes, the rest of the joined suspend/resume event functionality is in #257 which I just opened.

lib/animas/animasSimulator.js
gniezen: 360000 milliseconds = 6 minutes = 22:56 - 23:02 ?

gniezen:
OK, I think I've addressed all the comments on this one. @jebeck, do you want to double-check my changes before I merge into `animas-dataservices`?

lib/animas/animasSimulator.js
jebeck: I don't know, when the `deviceTime` thing tripped me up, it did _not_ match the times in my pump history that I was scrolling through. (I had a bolus showing a `deviceTime` of between 4-5 a.m. when I was definitely not awake.) Unless you're saying that the issue with `time` was having some kind of a cascading effect? Let me know if/when/on what branch you want me to test again to see...

lib/animas/animasSimulator.js
jebeck: Balls I was clearly just reading it as 3600000 automatically = 1 hour (since that's such a common duration for basals). /o\

lib/drivers/animasDriver.js
```diff
@@ -1157,12 +1157,16 @@ module.exports = function (config) {
           t.setSeconds(t.getSeconds() + 30);
           t = t.toISOString();
 
+          var dt = new Date(bolusdatum.deviceTime);
+          dt.setSeconds(dt.getSeconds() + 30);
```
jebeck: What's this addition of 30 secs for? We shouldn't have to do anything to `deviceTime` for de-duping...

PR #257, opened 2016-03-22
Joining Animas suspend/resume events

gniezen:
We need to do this in Uploader, as we're uploading these to the Data Service API and cannot use the Jellyfish functionality.

### DIFF
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -167,27 +167,25 @@ exports.make = function(config){
       }
 
       if(currStatus !== null) {
-        var resumed = Date.parse(currStatus.time) + currStatus.duration;
-        if (resumed > Date.parse(event.time)) {
-          //status is suspended during this event
-
-          if (event.rate === 0 ) {
+        if (event.rate === 0 ) {
+          var resumed = Date.parse(currStatus.time) + currStatus.duration;
+          if (resumed > Date.parse(event.time)) {
             //this basal is suspended
             event.deliveryType = 'suspend';
           }
-          else {
-            // basal is non-zero, but status still suspended
-            // is this a suspend event we created from an alarm?
-            // If so, we should update the joined suspend/resume event
-            if(currStatus.reason.suspended === 'automatic') {
+        }
+        else {
+          // basal is non-zero, but status still suspended
+          // is this a suspend event we created from an alarm?
+          // If so, we should update the joined suspend/resume event
+          if(currStatus.reason.suspended === 'automatic') {
 
-              currStatus.duration = currBasal.duration;
-              currStatus.reason.resumed = 'automatic';
-              annotate.annotateEvent(currStatus, 'animas/status/fabricated-from-alarm');
+            currStatus.duration = currBasal.duration;
+            currStatus.reason.resumed = 'automatic';
+            annotate.annotateEvent(currStatus, 'animas/status/fabricated-from-alarm');
 
-              dataservicesEvents.push(currStatus);
-              setCurrStatus(null);
-            }
+            dataservicesEvents.push(currStatus);
+            setCurrStatus(null);
           }
         }
       }
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -138,13 +138,23 @@ exports.make = function(config){
       simpleSimulate(event);
     },
     resume: function(event) {
-      ensureTimestamp(event);
-      if (currStatus != null && currStatus.index < event.index) {
-        annotate.annotateEvent(event,'animas/out-of-sequence');
+      if(currStatus != null) {
+        ensureTimestamp(event);
+
+        var duration = Date.parse(event.time) - Date.parse(currStatus.time);
+
+        if(duration >= 0) {
+          currStatus.duration = duration;
+        }
+        else {
+          currStatus.duration = 0;
+          annotate.annotateEvent(event,'basal/unknown-duration');
+        }
+        currStatus.reason.resumed = event.reason.resumed;
+
+        dataservicesEvents.push(currStatus.done());
+        setCurrStatus(null);
       }
-      event = event.done();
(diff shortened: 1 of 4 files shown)


gniezen:
@jebeck OK, this one is ready for review. Data Services POST records match what I have on my pump.

lib/animas/animasSimulator.js
```diff
@@ -167,24 +167,25 @@ exports.make = function(config){
       }
 
       if(currStatus !== null) {
-        if(currStatus.status === 'suspended') {
```
jebeck: I'm having some trouble reading this logic, but it's probably just a *me* problem. So the reason you've deleted the check against `currStatus.status` is because the *only* statuses that exist now are `suspended`s, right?

lib/animas/animasSimulator.js
```diff
@@ -167,24 +167,25 @@ exports.make = function(config){
...
-          if (event.rate === 0) {
+        if (event.rate === 0 ) {
```
jebeck: I took me several reads of this if/else block to get it. I think a comment might be useful, perhaps to the effect of "categorize the basal we're processing as a suspend basal if it's rate is 0 and its start falls within the suspend event we have on deck". (Did I get that right?)

lib/animas/animasSimulator.js
gniezen: Correct.

jebeck:
Still bouncing around between reading through the logic carefully in the diff and looking at my data (when I'm trying to make sure I understand something).

In looking at my data, found that there is still an issue with `deviceTime` vs. `time`. If you require the JSON records as `recs` in n_ you can use this bit to see what I mean:
```
_.filter(recs, function(d) { return d.time.slice(0,-5) === d.deviceTime; })
```

Seems to affect `deviceEvent`s and `basal`s.

lib/drivers/animasDriver.js
```diff
@@ -1644,28 +1644,31 @@ module.exports = function (config) {
...
+            .with_reason({suspended: 'manual', resumed: 'manual'})
+            .with_duration(suspendresumedatum.resumeJsDate - suspendresumedatum.suspendJsDate)
+            .with_payload({
+              resumeDeviceTime : suspendresumedatum.resumeDeviceTime
+            })
+            .set('index', suspendresumedatum.index);
+          cfg.tzoUtil.fillInUTCInfo(suspendResume, suspendresumedatum.suspendJsDate);
+          postrecords.push(suspendResume.done());
+        }
+        else {
```
jebeck: So remind me, when you set the clock back after a suspend, is the resume date the same Jan 1, 2007 as it is when there's no resume (e.g., before uploading) or is it just a date before the suspend? If the latter, do we want to treat it slightly differently? Stick the out-of-order resume in the payload at least?

test/node/animas/testAnimasSimulator.js
```diff
@@ -243,82 +243,58 @@ describe('animasSimulator.js', function() {
...
         .with_timezoneOffset(0)
         .with_conversionOffset(0)
         .with_status('resumed')
-        .with_reason({resumed: 'manual'});
+        .with_duration(3600000)
+        .with_reason({suspended: 'manual', resumed: 'manual'})
+        .done();
 
-      it('a suspend passes through', function() {
+      it('a single suspend passes through', function() {
```
jebeck: Shouldn't a single suspend not just pass through but have an annotation re: incomplete tuple? This is a comment on the test fixture I guess.

lib/drivers/animasDriver.js
```diff
@@ -1644,28 +1644,31 @@ module.exports = function (config) {
...
 
-        suspend = cfg.builder.makeDeviceEventSuspend()
-          .with_deviceTime(suspendresumedatum.suspendDeviceTime)
-          .with_reason({suspended: 'manual'})
-          .set('index', suspendresumedatum.index);
-        cfg.tzoUtil.fillInUTCInfo(suspend, suspendresumedatum.suspendJsDate);
-        suspend = suspend.done();
-        postrecords.push(suspend);
-
         if(suspendresumedatum.resumeJsDate > suspendresumedatum.suspendJsDate) {
```
jebeck: Please make this `>=` to handle brief suspend/resumes in the same minute. I have one in my data that's coming through like this:
```
    {
        "time": "2016-03-21T18:01:00.000Z",
        "timezoneOffset": -420,
        "clockDriftOffset": 0,
        "conversionOffset": 0,
        "deviceTime": "2016-03-21T11:01:00",
        "deviceId": "IR1285-72-36121-15",
        "type": "deviceEvent",
        "subType": "status",
        "status": "suspended",
        "reason": {
            "suspended": "manual"
        },
        "payload": {
            "logIndices": [
                7
            ]
        },
        "annotations": [
            {
                "code": "status/incomplete-tuple"
            }
        ]
    },
```

But the UI shows a suspend and a resume at `2016-03-21T11:01:00`. So will have `duration: 0` which is a bit of a lie, so also maybe another annotation, so we can say "This suspend was too brief to capture the duration" or similar.

lib/animas/animasSimulator.js
gniezen: Correct, comment added.

lib/drivers/animasDriver.js
gniezen: It's the latter. I'm changing it to now always put the resume timestamp in the payload for suspends. I don't want to check specifically for 0 Jan 2007, as it may be different for the Vibe (and 0 Jan 2007 is also displayed in the UI, so it's not "wrong").

gniezen:
@jebeck This branch did not have the commit to use the most recent record for TZOUtil, and I've now merged it in. Can you check if the `deviceTime`/`time` issue is still there?

gniezen:
As all the other comments have been addressed, I'm going to merge this and then we can handle the `deviceTime`/`time` issue in #251.

PR #263, opened 2016-04-01
Adding support for Animas Vibe

gniezen:
- Checks pump model number to determine start year (2007 vs 2008)
- ignores CGM calibration records on Vibe (for now)

Known issue: Vibe screen timeout must be set to 60 seconds. Will be address by another PR.

### DIFF
--- a/lib/core/device.js
+++ b/lib/core/device.js
@@ -34,7 +34,7 @@ var insuletOmniPod = require('../drivers/insuletDriver');
 var oneTouchUltra2 = require('../drivers/oneTouchUltra2');
 var abbottFreeStyleLite = require('../drivers/abbottFreeStyleLite');
 var bayerContourNext = require('../drivers/bayerContourNext');
-var animasPing = require('../drivers/animasDriver');
+var animasDriver = require('../drivers/animasDriver');
 
 var device = {
   log: require('../bows')('Device')
@@ -53,7 +53,7 @@ device._deviceDrivers = {
   'BayerContourNextUsb': bayerContourNext,
   'BayerContourUsb': bayerContourNext,
   'BayerContourNextLink': bayerContourNext,
-  'AnimasPing': animasPing
+  'Animas': animasDriver
 };
 
 device._deviceComms = {
@@ -68,7 +68,7 @@ device._deviceComms = {
   'BayerContourNextUsb': hidDevice,
   'BayerContourUsb': hidDevice,
   'BayerContourNextLink': hidDevice,
-  'AnimasPing': serialDevice
+  'Animas': serialDevice
 };
 
 device._silentComms = {};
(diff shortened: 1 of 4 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;YvCxfQdb&#x2F;3-ux-0-dev-1-tst-3-uploader-animas-vibe-support-no-cgm"> &lt;ux:0, dev:1, tst:3&gt; [uploader] Animas Vibe support (no CGM)</a>

lib/core/device.js
```diff
@@ -34,7 +34,7 @@ var insuletOmniPod = require('../drivers/insuletDriver');
 var oneTouchUltra2 = require('../drivers/oneTouchUltra2');
 var abbottFreeStyleLite = require('../drivers/abbottFreeStyleLite');
 var bayerContourNext = require('../drivers/bayerContourNext');
-var animasPing = require('../drivers/animasDriver');
+var animasDriver = require('../drivers/animasDriver');
```
darinkrauss: So, the Vibe will use the same driver as the Ping?

lib/drivers/animasDriver.js
```diff
@@ -955,8 +958,11 @@ module.exports = function (config) {
   };
 
   var decodeDate = function(encoded) {
-    var year = (encoded.monthYear & 0x0f) + 2007; // OneTouch Ping starts in year 2007 as 0
-    // TODO: Animas Vibe starts in 2008
+    var startYear = 2007; // OneTouch Ping starts in year 2007 as 0
+    if(modelNumber === 'IR1295') {
```
darinkrauss: Since we know we end up with a lot of unusual situations in the wild, would it be useful to be exact in what we expect and fail otherwise?  Be defensive, that is. For example:

```
var startYear;
if (modelNumber === 'IR1285') {   // Ping
  startYear = 2007;
} else if (modelNumber === 'IR1295') {  // Vibe
  startYear = 2008;
} else {
  // Unexpected, so fail fast here...
}
```

darinkrauss:
@gniezen Looks good with a couple of comments. :shipit:

lib/core/device.js
gniezen: Yes, both will use the same driver, so I needed to make the driver name more generic.

lib/drivers/animasDriver.js
gniezen: Good point!

gniezen:
Addressed comments; merging.

jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yemDk9OV&#x2F;21-blip-add-blip-notes-asset-and-its-accompanying-footer-link"> [blip] add  blip notes asset and its accompanying footer link</a>

PR #264, opened 2016-04-01
Non IBF file upload state fix

krystophv:
In handling the `READ_FILE_ABORTED` action, the uploads reducer fails to reset the rest of the download states' `disabled` property that gets set in the `CHOOSING_FILE` handler. This uses the same sort of logic from the `CHOOSING_FILE` handler to undo what it does in order to restore upload functionality for the rest of the download options. 

I also added the accept property to the file upload element in order to encourage (though it's not a failsafe sort of thing) users to select an actual `.ibf` file by having other files either not displayed or disabled in the file selector.

### DIFF
--- a/lib/components/Upload.js
+++ b/lib/components/Upload.js
@@ -181,7 +181,7 @@ export default class Upload extends Component {
           disabled={disabled}
           ref="file"
           type="file"
-          accept={upload.source.extension}
+          accept={_.get(upload, 'source.extension')}
           onChange={this.onBlockModeInputChange}/>
       </div>
     );
--- a/lib/components/Upload.js
+++ b/lib/components/Upload.js
@@ -181,6 +181,7 @@ export default class Upload extends Component {
           disabled={disabled}
           ref="file"
           type="file"
+          accept={upload.source.extension}
           onChange={this.onBlockModeInputChange}/>
       </div>
     );
--- a/lib/redux/reducers/uploads.js
+++ b/lib/redux/reducers/uploads.js
@@ -122,7 +122,7 @@ export function uploadsByUser(state = {}, action) {
         });
       });
       if (uploadTargetUser && uploadTargetDevice) {
-        let devicesForCurrentUser = _.get(state, [uploadTargetUser], {});
+        let devicesForCurrentUser = _.get(state, uploadTargetUser, {});
         _.forOwn(devicesForCurrentUser, (upload, key) => {
           newState = update(
             newState,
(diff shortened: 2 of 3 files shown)


jebeck:
Only glanced, but add a regression test plz! (For the reducer certainly. Probably nothing needs to be changed for the async action creator unless a wrong file test case wasn't there in which case there prolly _should_ be one.)

krystophv:
@jebeck: now with test! :fireworks:

krystophv:
https://trello.com/c/rGD9tosH

krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;rGD9tosH&#x2F;5-ux-0-dev-3-tst-1-uploader-bug-ui-if-user-tries-to-upload-a-non-ibf-file-for-omnipod-all-other-upload-options-will-be-grayed-out-"> &lt;ux:0, dev:3, tst:1&gt; [uploader] [bug] [ui] if user tries to upload a non-.ibf file for OmniPod, all other upload options will be grayed out until after a successful OmniPod upload</a>

lib/components/Upload.js
```diff
@@ -181,6 +181,7 @@ export default class Upload extends Component {
           disabled={disabled}
           ref="file"
           type="file"
+          accept={upload.source.extension}
```
jebeck: Probably should use `_.get` for some extra security on the availability of the `extension` field. We wouldn't want to crash the whole app just because we can't find that property. Although then we need a fallback for the file extension...is there a wildcard you can (pointlessly) give to `accept`?

lib/redux/reducers/uploads.js
```diff
@@ -121,16 +122,33 @@ export function uploadsByUser(state = {}, action) {
...
-        return update(
-          state,
-          {[uploadTargetUser]: {[uploadTargetDevice]: {
-            choosingFile: {$set: false},
-            completed: {$set: true},
-            error: {$set: err},
-            failed: {$set: true}
-          }}}
-        );
+        let devicesForCurrentUser = _.get(state, [uploadTargetUser], {});
```
jebeck: Nitpick: I find it a bit weird to use the array syntax for `_.get` with just one value. And about 70% certain that the rest of the codebase here in the uploader doesn't use the array syntax in these situations, but not 100% certain...

jebeck:
Given your computer troubles @krystophv, I'm inclined to just get this fix in, and if we want to do something different later, we can.

Thoughts @krystophv @gniezen?

krystophv:
I didn't get to the stylistic changes, but if those are alright for now (the second one was actually just copy/paste code from another test and the first one doesn't error if that property is undefined - it just doesn't include the `accept` DOM property) then I'm alright getting it out.

krystophv:
@jebeck - got the stylistic changes for _.get() in

jebeck:
:shipit:

PR #265, opened 2016-04-04
WIP: Animas multi-record downloads

gniezen:
The Animas Vibe has the capability to pack multiple records inside a single data packet.


gniezen:
This was closed automatically because the merge target was merged.

PR #268, opened 2016-04-07
fix Insulet pumpSettings time

jebeck:
A property renaming from `dt` -> `deviceTime` slipped by us, and we were passing `undefined` to generate the timestamp for the `pumpSettings` object from sundial. Since the `pumpSettings` timestamp is by definition the most recent timestamp in Insulet data, it is used as the starting point for BtUTC, so this bug had hugely cascading effects o_O

See also [sundial #26](https://github.com/tidepool-org/sundial/pull/26) which this depends on (recommend using `npm link` locally to try it out).


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xYooCLdS&#x2F;19-uploader-omnipod-fix-incorrect-time-on-pumpsettings-object-which-also-is-the-kick-off-for-btutc"> [uploader] [OmniPod] fix incorrect `time` on pumpSettings object (which also is the kick-off for BtUTC)</a>

gniezen:
LGTM. Merging.

PR #269, opened 2016-04-07
Reworking Animas device comms to prevent screen timeouts

gniezen:
With the existing device comms strategy, when reading a specific record type, we terminate the connection when the record is empty (as all subsequent records will be empty), and reconnect to read the next record type. It's not possible to just send a reset command, as the pump will continue to try to send the next empty record. Diasend reads all the records for each type, even if they're empty, and we were hoping to eliminate this inefficiency.

Unfortunately, if the disconnect command is received on the Animas Vibe, the screen goes dark, making it impossible to reconnect to the pump.

One side-effect of new Animas device comms strategy: We are downloading ​_all_​ the records (including empty ones) off the pump each time (just like Diasend does), even if you only have a single bolus record on your pump. Luckily the pump has a small memory, so this only takes 2.5 minutes on my machine.

### DIFF
--- a/lib/drivers/animasDriver.js
+++ b/lib/drivers/animasDriver.js
@@ -46,6 +46,8 @@ module.exports = function (config) {
   var EOM_BYTE = 0xC1;
   var ADDRESS_CONNECT = 0xFF; //used to establish connection
 
+  var RETRIES = 10;
+
   var CMDS = {
     CONNECT: { value: 0x93, name: 'CONNECT'},
     DISCONNECT: { value: 0x53, name: 'DISCONNECT'},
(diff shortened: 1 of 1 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;YvCxfQdb&#x2F;3-ux-0-dev-1-tst-3-uploader-animas-vibe-support-no-cgm"> &lt;ux:0, dev:1, tst:3&gt; [uploader] Animas Vibe support (no CGM)</a>

gniezen:
Self-merging as it's not going to master.

PR #271, opened 2016-04-12
GitBook

jebeck:
This just affects dev tooling (docs), so will be self-merging if there are no objections in ~24 hrs.

CC @gniezen @krystophv


scripts/update-gh-pages.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+echo ""
+echo "### Building docs static website with GitBook... ###"
+echo ""
+
+./node_modules/.bin/gitbook build
+
+echo ""
+echo "### Syncing docs static webiste to web/ directory... ###"
```
gniezen: ✋, it's the spelling 👮!

PR #274, opened 2016-04-19
Fix for timeout error at with 50-60% progress bar

gniezen:
Problem: Some clinics have reported spurious timeout errors for Bayer meters, which can be resolved by retrying multiple times. It looks like these are produced for devices with lots of records
screen shot 2016-04-18 at 21 54 58

Possible cause: `abortTimer` was only cleared when the packet is successful, not when the packet is empty.  When the `listenTimer` interval timer fires multiple times, we have the unfired `abortTimer`s building up and all firing after their timeout of 5s.

Possible solution: Clearing `abortTimer` as soon as we receive a packet, but only clearing `listenTimer` when we receive a successful packet.

I haven't been able to reproduce the error myself, but it's possible to try and emulate it by changing the timeout value from 5s to 100ms. The current `master` branch will timeout, whereas this branch will succeed even with such a short timeout.

@jebeck or @darinkrauss to review?

### DIFF
--- a/lib/drivers/bayerContourNext.js
+++ b/lib/drivers/bayerContourNext.js
@@ -322,7 +322,7 @@ module.exports = function (config) {
                 if (err) {
                   debug('Failure trying to read record', record.frame);
                   debug(err);
-                  return whilstCb(err);
+                  return whilstCb(err,null);
                 } else {
                   robj.timestamp = parseInt(r[4]);
                   robj.annotations = getAnnotations(r[3], data);
--- a/lib/drivers/bayerContourNext.js
+++ b/lib/drivers/bayerContourNext.js
@@ -392,18 +392,26 @@ module.exports = function (config) {
   };
 
   var getASTMMessage = function (timeout, retries, callback) {
+    var e = new Error('Timeout error.');
+    e.name = 'TIMEOUT';
+
     var abortTimer = setTimeout(function () {
       clearInterval(listenTimer);
       debug('TIMEOUT');
-      var e = new Error('Timeout error.');
-      e.name = 'TIMEOUT';
       return callback(e, null);
     }, timeout);
 
     var listenTimer = setInterval(function () {
       hidDevice.receive(function(raw) {
         var packet = new Uint8Array(raw);
+
+        // we need to reset the timeout
         clearTimeout(abortTimer);
+        abortTimer = setTimeout(function () {
+          clearInterval(listenTimer);
+          debug('TIMEOUT');
+          return callback(e, null);
+        }, timeout);
 
         // Only process if we get data
         if ( packet.length === 0 ) {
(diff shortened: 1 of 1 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Ui9xjrx4&#x2F;33-uploader-bug-timeout-error-when-progress-bar-is-at-50-60"> [uploader] Bug: Timeout error when progress bar is at 50-60%</a>

lib/drivers/bayerContourNext.js
```diff
@@ -322,7 +322,7 @@ module.exports = function (config) {
                 if (err) {
                   debug('Failure trying to read record', record.frame);
                   debug(err);
-                  return whilstCb(err);
+                  return whilstCb(err,null);
```
darinkrauss: This isn't technically necessary, right?

lib/drivers/bayerContourNext.js
```diff
@@ -403,6 +403,7 @@ module.exports = function (config) {
     var listenTimer = setInterval(function () {
       hidDevice.receive(function(raw) {
         var packet = new Uint8Array(raw);
+        clearTimeout(abortTimer);
```
darinkrauss: Don't you need to reset the timeout here, rather than just clear it? What happens if the interval keeps firing, but never reaches a clearInterval? The interval would keep going forever, because the timeout was cancelled at the first interval.

darinkrauss:
A couple of comments.

lib/drivers/bayerContourNext.js
gniezen: No, not necessary. I just added it to be consistent with the other callbacks.

gniezen:
@darinkrauss Good catch regarding the timeout. I'm now resetting the timeout and only clearing it when a successful packet is received.

lib/drivers/bayerContourNext.js
```diff
@@ -392,18 +392,27 @@ module.exports = function (config) {
   };
 
   var getASTMMessage = function (timeout, retries, callback) {
+    var e = new Error('Timeout error.');
```
darinkrauss: Might want to call it `timeoutError` rather than `e`.

lib/drivers/bayerContourNext.js
```diff
@@ -392,18 +392,27 @@ module.exports = function (config) {
...
       return callback(e, null);
     }, timeout);
 
     var listenTimer = setInterval(function () {
       hidDevice.receive(function(raw) {
         var packet = new Uint8Array(raw);
 
+        // we need to reset the timeout
+        clearTimeout(abortTimer);
+        abortTimer = setTimeout(function () {
```
darinkrauss: Might want to shove this in a function to DRY this up. And you can pull the `Error` back into the function since it will only be there once.

darinkrauss:
A final couple of nits. Take them or leave them.  😄  :shipit:

gniezen:
I don't believe anymore that this is actually causing the spurious timeout issue, and the existing `abortTimer` is just fine. Solving the real issue is as simple as increasing the `listenTimer` interval, which has now been done in commit 474a0dd

PR #273, opened 2016-04-19
Handling timestamps in future when current interval start is null

gniezen:
We’ve got an OmniPod `Failed to look up UTC info!` error. There’s only one time change event (1 April to 15 April) and the first event is on 4 April.

@jebeck to review if the change is correct.

PR #275, opened 2016-04-19
make sure log index of 0 gets passed through for bootstrapping

jebeck:
I believe this is the root issue of the fix being attempted in #273 

How I got here: when I started looking at @gniezen's changes, the first thing I noticed was that they were in an `else` clause that should very rarely be triggered - for situations when we **don't** have the log index for sequencing. In OmniPod this only happens with some alarms that are not "history" records with normal log indices. The uploader was actually already logging the object that failed to have its UTC info looked up (although it wasn't super clear that that's *what* was being logged - another change I made here was to add some more text to that debug message to clarify). When I looked at the logged object causing the lookup failure it was NOT an alarm at all, and it had an index (of `0`). So then the hunt was to figure out why that index wasn't getting passed through, and shamefully truthiness was to blame /o\

I think the logic here is correct, but I'm definitely open to suggestions re: better way to "phrase" it for readability/understandability. We should also try to see if we can add a regression test if I don't get to that by the end of today.


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;TfVVfeqA&#x2F;34-uploader-bug-omnipod-failed-to-look-up-utc-info-error"> [uploader] Bug: OmniPod  `Failed to look up UTC info!` error</a>

jebeck:
Ping @gniezen for review if that wasn't obvious 😜

jebeck:
Added the regression test myself, so 101% ready for review.

gniezen:
LGTM. Merging.

PR #276, opened 2016-04-26
move timezone to upload screen and intro redux-persist 

krystophv:
- Moves the timezone selection dropdown to the upload screen
- introduces suggested device time indicator below timezone dropdown
- introduces redux-persist as the local storage mechanism for targetDevices and targetTimezones


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;qV6VFkyO&#x2F;11-ux-1-dev-2-tst-2-uploader-in-clinic-update-current-uploader-to-match-new-in-clinic-design-timezone-placement-and-suggested-devic"> &lt;ux:1, dev:2, tst:2&gt; [uploader] [in-clinic] update current uploader to match new in-clinic design timezone placement and suggested device time</a>

lib/components/TimezoneDropdown.js
```diff
@@ -45,6 +45,19 @@ var TimezoneDropdown = React.createClass({
     }
   },
 
+  componentDidMount: function() {
+    var self = this;
+    self.updateSuggestedInterval = setInterval(
```
jebeck: Interesting... The trendier cousin of `redux-thunk`, a lib called `redux-saga`, has the advantage of working well for timer-type tasks. I guess we should keep an eye on how often we need such things, although I think it'd take a lot to convince me that we should port to sagas from thunks in the uploader - it would be a fair amount of porting.

lib/redux/actions/utils.js
```diff
@@ -31,6 +31,24 @@ export function getDeviceTargetsByUser(targetsByUser) {
   });
 }
 
+export function getUsersWithTargets(targetDevices, targetTimezones){
```
jebeck: Unit tests for this? Only if it's gonna survive the next steps of persisting to back-end...

jebeck:
Closing this one since #280 includes it. (We can always re-open if necessary.)

PR #277, opened 2016-04-28
Sending Animas through new APIs; everything else through Jellyfish

gniezen:
By adding an optional `uploadType` parameter to `toPlatform` we can upload using either `jellyfish` or `dataservices`.

Ping @darinkrauss for review.


darinkrauss:
Looks good. 🚢

PR #278, opened 2016-04-28
New platform dataservices endpoints for Animas pump

darinkrauss:
@gniezen Back at ya!


gniezen:
LGTM! Merging, and going to make a note on the parent PR to `npm link` to the right `platform-client`.

PR #281, opened 2016-05-03
Bug fix: handling temp basal without a temp_rate_start event

gniezen:
When a temp basal starts around the same time as an extended basal, it's possible that there's no `temp_basal_start` basal rate change event generated. We have to use the `LID_TEMP_RATE_ACTIVATED` event to determine when the temp basal started, and annotate the temp basal with `tandem/basal/temp-without-rate-change`.

More detail: https://trello.com/c/ynVaskZx

@jebeck to test and review?

### DIFF
--- a/lib/drivers/tandemDriver.js
+++ b/lib/drivers/tandemDriver.js
@@ -1974,17 +1974,18 @@ module.exports = function (config) {
       }
 
       if (!_.isEmpty(data.log_records)) {
+        // ordering below is important for simulator, e.g.
+        // buildTempBasalRecords needs to be before buildBasalRecords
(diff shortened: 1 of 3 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ynVaskZx&#x2F;39-tandem-missing-temp-basal-event"> [Tandem] missing temp basal event</a>

gniezen:
I believe I've now been able to reproduce the various instances in which this bug occurs and documented the steps in the test strategy on https://trello.com/c/ynVaskZx. @jh-bate to review?

jh-bate:
LGTM - includes tests and code matches the existing style

PR #283, opened 2016-05-09
Using embedded bolus GUID for wizard event

gniezen:
Jellyfish used to squash embedded bolus events down into a hash ID. As the new platform API will not be doing this, we need to move where we're generating the GUID so that we can put that it in the wizard event instead of the bolus event itself.


gniezen:
Closing this, as we've decided to rather send embedded boluses in wizard events, and let the platform API create the linkage and separate bolus event.

PR #285, opened 2016-05-11
Using USB product ID and vendor ID (PID/VID) to find device (on Chrome 50+)

gniezen:
A regression in Google Chrome 45 has now partially been fixed in Chrome 50. 

How it used to work: We pass the USB PID/VID to `chrome.usb.getDevices`, and only get back the devices that match that combo

What broke and the hotfix: Windows doesn't return any serial devices, so we removed the PID/VID filter and returned all serial devices.

How it's fixed now in this PR: On Windows Chrome 50+, `chrome.serial.getDevices` now returns the PID/VID combo as well, so we can use this to filter ourselves. OS X does not return the PID/VID combo, so we use `chrome.usb.getDevices` on Mac.

### DIFF
--- a/lib/core/device.js
+++ b/lib/core/device.js
@@ -277,28 +277,25 @@ device.detectUsb = function(driverId, cb) {
     var chromeVersion = /Chrome\/([0-9]+)/.exec(navigator.userAgent)[1];
     self.log('Chrome version detected as', chromeVersion);
 
-    if (chromeVersion >= 50) {
-      chrome.serial.getDevices(function(results) {
+    chrome.serial.getDevices(function(results) {
+      if (chromeVersion >= 50) {
         // As of Chrome 50, product ID and vendor ID is returned
         // on Windows (only) for serial.getDevices
         results = _.filter(results, function(p) {
           return (p.vendorId === driverManifest.vendorId) &&
           (p.productId === driverManifest.productId);
         });
-
-        getDevice(results);
-      });
-    }
-    else {
-      // this is the hotfix for Windows Chrome 49 and earlier
-      chrome.serial.getDevices(function(results) {
+      }
+      else{
+        // With Chrome 49 and earlier, there may be
+        // multiple connected devices
         self.log('Connected device(s): ');
         for(var i in results) {
             self.log(results[i].path);
         }
-        getDevice(results);
-      });
-    }
+      }
+      getDevice(results);
+    });
   }
   else {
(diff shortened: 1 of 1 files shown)


lib/core/device.js
```diff
@@ -232,11 +232,8 @@ device.detectUsb = function(driverId, cb) {
     vendorId: driverManifest.vendorId,
     productId: driverManifest.productId
   };
-  chrome.serial.getDevices(function(results) {
```
HowardLook: Why remove logging of the connected devices. Doesn't that having that help us when things go wrong?

lib/core/device.js
```diff
@@ -274,7 +271,39 @@ device.detectUsb = function(driverId, cb) {
...
-  });
+  };
+
+  if (self._os === 'win') {
+    var chromeVersion = /Chrome\/([0-9]+)/.exec(navigator.userAgent)[1];
+    self.log('Chrome version detected as', chromeVersion);
+
+    chrome.serial.getDevices(function(results) {
+      if (chromeVersion >= 50) {
+        // As of Chrome 50, product ID and vendor ID is returned
```
HowardLook: In comment, can you link to the chrome/chromium issue that shows the bug or fix?

gniezen: No, that issue is still open, which why we need Windows- and Mac-specific code here. Once they fix issue 529037, it will simplify things a bit. I've now added the link to the Chromium bug report in the comments: https://bugs.chromium.org/p/chromium/issues/detail?id=599101

lib/core/device.js
gniezen: I haven't removed it, I just moved it down to line 293-297 😉

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ja99ARQw&#x2F;13-ux-0-dev-3-tst-3-uploader-verify-product-id-and-vendor-id-for-usb-serial-devices"> &lt;ux:0 dev:3 tst:3&gt; [uploader] verify product ID and vendor ID for USB serial devices</a>

lib/core/device.js
HowardLook: Ah, thanks. Makes sense.

lib/core/device.js
HowardLook: OK, thanks for the clarification.

HowardLook:
LGTM to merge. Thanks!

PR #286, opened 2016-05-11
Fix new data endpoints (temporarily via Styx)

darinkrauss:
@gniezen These are the currently planned endpoints for now. Later (when we have time) we'll move it a separate "data" address.


gniezen:
Should `.config.js` also be changed?

darinkrauss:
@gniezen Good point. Updated `.config.js` and committed. How does that look?

gniezen:
LGTM. Merging.

PR #287, opened 2016-05-18
Support for Animas device settings

gniezen:


gniezen:
Self-reviewed and merging, as it's not going to master.

PR #288, opened 2016-05-20
Tandem bug fix: Handling events around a pump shutdown 

gniezen:
If a pump shutdown occurs and the clock is set back to 2008 it can cause unexpected issues. For those we've been able to reproduce, this PR does two things:

- Drops a normal bolus if it's incomplete because the start event is missing due to a time change
- Drops a new-day event if it is immediately followed by a time-change event. (We use new-day events to fabricate scheduled basals in order to break up flat-rate basals. They can occur after you switch the pump back on, but before you change the date from 2008 to the correct time, which means we fabricate a scheduled basal in 2024.) 

Ping @jh-bate for review.

### DIFF
--- a/lib/drivers/tandemDriver.js
+++ b/lib/drivers/tandemDriver.js
@@ -1547,8 +1547,14 @@ module.exports = function (config) {
           .with_expectedExtended(bolus.bolex_insulin_requested);
       }
       record = record.set('index', bolus.index);
-      cfg.tzoUtil.fillInUTCInfo(record, bolus.jsDate);
-      records.push(record.done());
+
+      if(record.subType === 'normal' && bolus.insulin_delivered === undefined) {
+        debug('Only part of a bolus received, dropping:', bolus);
+      }
+      else{
+        cfg.tzoUtil.fillInUTCInfo(record, bolus.jsDate);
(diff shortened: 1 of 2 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;vSH6qZej&#x2F;15-bug-uploader-tandem-issues-when-battery-dies"> [bug][uploader] Tandem issues when battery dies</a>

jh-bate:
@gniezen code looks reasonable but do we have any tests ?

gniezen:
We don't have any tests (for any of the drivers) when pulling data off the pumps. All our testing happens in the second part: parsing data to fit our data model before sending off. This bugfix is in the first part unfortunately.

gniezen:
Hi @jh-bate ! I've added one more commit to this PR that fixes a small issue with Tandem that was reported today. I'll add the test strategy for this last commit to the Trello card. Can you have a quick look at it and then decide whether this PR is ready for merging?

jh-bate:
@gniezen LGTM

gniezen:
Merging, to be tested together with multi-record downloads merged from PR #289

PR #289, opened 2016-05-23
Support for Tandem multi-record downloads

gniezen:
See Trello card for details: https://trello.com/c/OIiGU5jG

@darinkrauss - are you ok to review? (This task is lowest priority for current sprint, so not urgent)

### DIFF
--- a/lib/drivers/tandemDriver.js
+++ b/lib/drivers/tandemDriver.js
@@ -979,6 +979,94 @@ module.exports = function (config) {
     }, INTERVAL_FREQ); // if we spin too quickly on this, packets don't get sent when window doesn't have focus
   };
 
+  var multiLogRequester = function (start, end, progress, callback) {
+    // TODO: implement and test multi-record download commands
+    if (__DEBUG__) {
+      debug('multiLogRequester', start, end);
+      var start_exec = Date.now();
+    }
+    var send_seq = start;
+    var receive_seq = start;
+    var recovering = false;
+    var percentage = 0;
+    var prevPercentage = 0;
+    var retryRecoverTimer;
+
+    tandemCommand(COMMANDS.LOG_ENTRY_SEQ_MULTI_REQ, [send_seq,end-send_seq], function (err) {
+      if(err) {
+        callback(err,null);
+      }
+      else {
+        var listenTimer = setInterval(function () {
+          if(pending) {
+            debug('pending');
+          }
+          while (cfg.deviceComms.hasAvailablePacket() && !pending) {
+            var processPacket = function (pkt) {
+              if (pkt.valid &&
+                pkt.descriptor === RESPONSES.LOG_ENTRY_TE.value &&
+                pkt.payload['header_log_seq_no'] >= receive_seq) {
+                if (receive_seq != pkt.payload['header_log_seq_no']) {
+                  if (!recovering) {
+                    recovering = true;
+                    debug('recovering ', receive_seq, '(received ',pkt.payload['header_log_seq_no'], ')');
+                    send_seq = receive_seq + 1;
+                  }
+
+                  tandemCommand(COMMANDS.LOG_ENTRY_SEQ_MULTI_REQ, [receive_seq,end-receive_seq], function (err) {
+                    if(err) {
+                      callback(err,null);
+                    }
+                    retryRecoverTimer = setTimeout(function() {
+                      if(recovering) {
+                        debug('Retrying to recover..',receive_seq);
+                        tandemCommand(COMMANDS.LOG_ENTRY_MULTI_SEQ_REQ, [receive_seq,end-receive_seq], function (err) {
+                          if(err) {
+                            callback(err,null);
+                          }
+                        });
+                      }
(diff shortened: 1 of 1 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OIiGU5jG&#x2F;16-ux-0-dev-5-tst-2-uploader-device-implement-tandem-multi-record-download-commands"> &lt;ux:0, dev:2, tst:2&gt; [uploader] [device] implement Tandem multi-record download commands</a>

lib/drivers/tandemDriver.js
```diff
@@ -821,6 +821,12 @@ module.exports = function (config) {
 
   var listenForPacket = function (command,args,callback) {
 
+    var abortTimer = setTimeout(function () {
+      clearInterval(listenTimer);
```
darinkrauss: Do we need to clear the `retryTimer` here, too?

lib/drivers/tandemDriver.js
```diff
@@ -821,6 +821,12 @@ module.exports = function (config) {
...
+      debug('TIMEOUT');
```
darinkrauss: Debug statement?

lib/drivers/tandemDriver.js
```diff
@@ -851,7 +858,16 @@ module.exports = function (config) {
     if (format) {
       payload_len = struct.structlen(format);
       payload = new Uint8Array(payload_len);
-      struct.pack(payload, 0, format, args);
+      switch(args.length) {
+        case 1 :
+          struct.pack(payload, 0, format, args);
+          break;
+        case 2:
+          struct.pack(payload, 0, format, args[0],args[1]);
```
darinkrauss: Space between arguments.

lib/drivers/tandemDriver.js
```diff
@@ -979,6 +994,79 @@ module.exports = function (config) {
     }, INTERVAL_FREQ); // if we spin too quickly on this, packets don't get sent when window doesn't have focus
   };
 
+  var multiLogRequester = function (start, end, progress, callback) {
+    if (__DEBUG__) {
+      debug('multiLogRequester', start, end);
+      var start_exec = Date.now();
```
darinkrauss: So, what is our standard for case, snake_case, camelCase? It looks like the rest of the code uses camelCase. We should be consistent.

lib/drivers/tandemDriver.js
```diff
@@ -979,6 +994,79 @@ module.exports = function (config) {
...
+    }
+    var send_seq = start;
+    var receive_seq = start;
+    var percentage = 0;
+    var prevPercentage = 0;
+
+    var retryTimeout = function() {
+      debug('Retrying from record ',receive_seq,' onwards.',end-receive_seq,' record(s) left..');
```
darinkrauss: Spacing between arguments (through the changes).

lib/drivers/tandemDriver.js
```diff
@@ -1131,14 +1219,33 @@ module.exports = function (config) {
...
       }
     }
 
     end_seq = data.end_seq;
     start_seq = data.start_seq;
-    tandemLogRequester(start_seq, end_seq, progress, iterate);
+    debug('Firmware version is #',data.firmware_version);
+    if(data.firmware_version >= COMMANDS.LOG_ENTRY_SEQ_MULTI_REQ.version) {
+      // woohoo, we can use multi-record downloads!
+      multiLogRequester(start_seq, end_seq+1, progress, iterate);
```
darinkrauss: Why is this `end_seq+1`?

darinkrauss:
@gniezen I'm not completely up on the whole device communication side of things so I can't really provide feedback on the comm stuff. It looks good otherwise, though I'd recommend you do another pass over all of the timers/intervals to make sure you are cancelling them all in the right locations.

lib/drivers/tandemDriver.js
gniezen: We have the exact same debug statement in several of the other drivers. For debug statements that are more "chatty", detailed logging is only available if `DEBUG_ERROR` environment variable is true, which enables debug statements wrapped in `if(__DEBUG__)` statements.

lib/drivers/tandemDriver.js
gniezen: When we send a multi-record request we send the number of records to retrieve instead of just the index number. I've moved where the one gets added to a place where it makes more sense, and also added a comment to explain.

lib/drivers/tandemDriver.js
gniezen: `retryTimer` should always fire before `abortTimer`, as `abortTimer` waits twice as long before firing.

gniezen:
@darinkrauss Thanks for the suggestion. I've made another pass of the Tandem driver and found quite a few places where interval timers weren't being cleared when an error occurs. Hopefully I've got them all now.

gniezen:
Just realized that @darinkrauss already reviewed and ok'ed this PR, so going to merge and build for testing.

PR #290, opened 2016-05-26
suppress authentication errors on profile update

krystophv:
this is specifically a fix for the work to store a user's timezone and device selection preferences in their profile to ensure that users with permissions to upload on behalf of another user are not presented with errors concerning permissions to also modify that user's profile data.

### DIFF
--- a/lib/redux/actions/async.js
+++ b/lib/redux/actions/async.js
@@ -427,6 +427,8 @@ export function setTargetTimezone(userId, timezoneName) {
         targetTimezone: timezoneName
       }
     }, (err, profile) => {
+      // suppress unauthorized error until custodial vs normal permissions are ironed out
+      // TODO: remove conditional when perms are finalized or accounts are converted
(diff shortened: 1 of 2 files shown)


krystophv:
@gniezen - can you review these changes? This is some pretty basic error suppression to move forward with storing the user's preferences in their patient profile while not re-working our permissions system.

krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;02LWvMli&#x2F;18-uploader-suppress-error-on-timezone-change-for-users-who-don-t-have-permission"> [uploader] Suppress error on timezone change for users who don&#39;t have permission</a>

gniezen:
LGTM - I also double-checked that the tests fail when the status code is changed in the code.

Shouldn't you maybe add some comments/TODOs so that this can be removed at a later stage when the permissions system is updated?

PR #291, opened 2016-06-03
Do not send byUser in upload record to new platform

darinkrauss:
@gniezen The platform sets this programmatically based upon the session token user (or will after I deploy the latest code).


gniezen:
LGTM! Merging.

PR #293, opened 2016-06-06
CGM Support for tandem t:slim G4

gniezen:
See https://trello.com/c/FPxMDjuR for details.

### DIFF
--- a/lib/drivers/docs/tandemCGM.md
+++ b/lib/drivers/docs/tandemCGM.md
@@ -1,0 +1,94 @@
+## Checklist for CGM Implementation
+
+(Key:
+
+ - `[x]` available in data protocol/documented in spec and implemented
+ - `[-]` available in data protocol/documented in spec but *not* yet implemented
+ - `[?]` unknown whether available in data protocol/documented in spec; *not* yet implemented
+ - `*[ ]` TODO: needs implementation!
+ - `[ ]` unavailable in data protocol and/or not documented in spec and not yet implemented)
+
+### Required if Present
+
+#### CBG
+
+  - `[ ]` cbg values
+  - `[ ]` units of cbg values (read from device, not hard-coded)
+  - `[ ]` out-of-range values (LO or HI)
+  - `[ ]` out-of-range value thresholds (e.g., often 40 for low and 400 for high on CGMs)
+
+Device-specific? (Add any device-specific notes/additions here.)
+
+#### Device Events
+  - `[ ]` calibrations
+    - `[ ]` calibration value
+    - `[ ]` units of calibration value (read from device, not hard-coded)
+  - `[ ]` time changes (presence of which is also in the [BtUTC section](#bootstrapping-to-utc) below)
+    - `[ ]` device display time `from` (before change) and `to` (result of change)
+    - `[ ]` agent of change (`automatic` or `manual`)
+    - `[ ]` timezone
+    - `[ ]` reason for change (read from device)
+
+Device-specific? (Add any device-specific notes/additions here.)
+
+#### Settings
+
+  - `[ ]` units preference for BG display
+  - `[ ]` units of data being uploaded (will be mutated to mmol/L storage units if not mmol/L)
+  - `[ ]` transmitter ID
+  - `[ ]` low alert settings
+    - `[ ]` enabled
+    - `[ ]` level/threshold
+    - `[ ]` snooze threshold
+  - `[ ]` high alert settings
+    - `[ ]` enabled
+    - `[ ]` level/threshold
+    - `[ ]` snooze threshold
+  - `[ ]` rate-of-change alerts
+    - `[ ]` fall rate alert
+        - `[ ]` enabled
+        - `[ ]` rate threshold for alerting
+    - `[ ]` rise rate alert
+        - `[ ]` enabled
+        - `[ ]` rate threshold for alerting
+  - `[ ]` out-of-range alerts
+    - `[ ]` enabled
+    - `[ ]` snooze time between alerts
+  - `[ ]` predictive alerts
+    - `[ ]` low prediction
+        - `[ ]` enabled
+        - `[ ]` time sensitivity (minutes to predicted low for alerting)
+    - `[ ]` high prediction
+        - `[ ]` enabled
+        - `[ ]` time sensitivity (minutes to predicted high for alerting)
+  - `[ ]` calibration alerts/reminders
+    - `[ ]` pre-reminder
+    - `[ ]` overdue alert
+
+Settings history:
+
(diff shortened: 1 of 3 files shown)


gniezen:
@jebeck This is the PR for review. It may be worthwhile checking the spec (specifically wrt calibration events) to make sure we interpret it the same. (Note that this work is low priority for this sprint, so don't worry about it too much)

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;FPxMDjuR&#x2F;11-ux-0-dev-2-tst-2-pump-tandem-t-slim-with-dexcom-g4-integration"> &lt;ux:0 dev:2 tst:2&gt; [pump] tandem t:slim with Dexcom G4 Integration</a>

lib/drivers/tandemDriver.js
```diff
@@ -1996,6 +2017,99 @@ module.exports = function (config) {
...
+        if (event.status === CGM_STATUS.CGM_CALIBRATION.value ||
+                 event.status === CGM_STATUS.CGM_LO_CALIBRATION.value ||
+                 event.status === CGM_STATUS.CGM_HI_CALIBRATION.value) {
+          // in response to calibration
+          record.with_payload({
+            calibration_response: true
+          });
+        }
+
+        record.with_value(value)
```
jebeck: I'm comparing with the spec, and I'm a little confused. It feels to me like you should be checking first if the `glucoseValueStatus === 0`, no? Are there no "magic numbers" for LID_CGM_DATA? The spec sort of reads like there's not, but parallelism with LID_CGM_CAL would suggest that there might be, and I'm not entirely sure we should trust one reading of the spec entirely, unless you have a way to test this.

lib/drivers/tandemDriver.js
```diff
@@ -1996,6 +2017,99 @@ module.exports = function (config) {
...
+      else if (event.header_id === PUMP_LOG_RECORDS.LID_CGM_DATA.value){
+
+        record = cfg.builder.makeCBG();
+
+        if (event.status === CGM_STATUS.CGM_CALIBRATION.value ||
+                 event.status === CGM_STATUS.CGM_LO_CALIBRATION.value ||
+                 event.status === CGM_STATUS.CGM_HI_CALIBRATION.value) {
+          // in response to calibration
+          record.with_payload({
+            calibration_response: true
```
jebeck: From a data model perspective, I'm wondering about this "value in response to calibration" business. Haven't seen discussion of it in the Dexcom specs, in my memory. Have you seen it there?

lib/drivers/docs/tandemCGM.md
```diff
@@ -0,0 +1,97 @@
...
+  - `[-]` rate-of-change alerts
+    - `[-]` fall rate alert
+        - `[-]` enabled
+        - `[-]` rate threshold for alerting
+    - `[-]` rise rate alert
+        - `[-]` enabled
+        - `[-]` rate threshold for alerting
+  - `[-]` out-of-range alerts
+    - `[-]` enabled
+    - `[-]` snooze time between alerts
```
jebeck: Hmm, not sure if checklist is wrong here, or if the Tandem spec is different from Dexcom's. The Tandem "Threshold" for the out of range alert is not a snooze time but rather a number of minutes time out of range before alerting. Looking at my G5 receiver, it appears the only thing settable is the threshold and not the snooze, so I suspect checklist is wrong.

...and now I've just looked at my data in prod, and not only is checklist wrong, but data is wrong! We'll have to make a data model change and rev `_schemaVersion` to correct, as `outOfRangeAlerts.snooze` should be `outOfRangeAlerts.threshold`... /o\

lib/drivers/docs/tandemCGM.md
```diff
@@ -0,0 +1,97 @@
...
+  - `[ ]` predictive alerts
+    - `[ ]` low prediction
+        - `[ ]` enabled
+        - `[ ]` time sensitivity (minutes to predicted low for alerting)
+    - `[ ]` high prediction
+        - `[ ]` enabled
+        - `[ ]` time sensitivity (minutes to predicted high for alerting)
+  - `[-]` calibration alerts/reminders
```
jebeck: I couldn't find this in the spec... Where did you see it?

lib/drivers/tandemDriver.js
gniezen: Good point, and we should also never trust that the spec is implemented the way it's specified. I'm now also checking that LID_CGM_DATA is either 40-400 or 0 (in which case it's changed to 39 or 401 depending if it's LO or HI)

lib/drivers/tandemDriver.js
gniezen: Nope, I was actually hoping you'd know more. Initially I thought they were the actual calibration value, but it looks more like it's the initial response to the calibration value. That is, it looks as if this initial response "over-corrects" after the calibration, with subsequent values being more in line with previous values. May be worthwhile asking Nate about?

lib/drivers/docs/tandemCGM.md
gniezen: Now I'm confused. I thought threshold and snooze time were the same thing, as in snooze time could be "time in between alerts": Sensor goes out of range ---snooze--> first alert --snooze--> second alert, or is that not how it works?

lib/drivers/docs/tandemCGM.md
gniezen: Looks like it's CGM alert ID 4 and/or 18  - see Table 12?

lib/drivers/docs/tandemCGM.md
jebeck: As discussed, the only thing settable for out of range is how many minutes must elapse (e.g., 15, 20, 30) in the "out of range" state before you get the alert.

lib/drivers/docs/tandemCGM.md
jebeck: Want to edit checklist draft since we discussed this and what you're referring to is not the setting (which doesn't exist) but the alert itself? As part of drafting CGM alert data models, I'll draft additions to the checklist.

lib/drivers/tandemDriver.js
jebeck: For posterity: in discussing we have decided to ask Nate.

gniezen:
Marking this as WIP for now, as it's dependent on #298 to be merged first.

lib/drivers/tandemDriver.js
gniezen: To follow up: Nate doesn't know either, so going to leave it in the payload as is for now.

gniezen:
I have just merged #298 - so now merging this too.

PR #280, opened 2016-05-02
Store device and timezone targets in profile

krystophv:
Store the user's targetDevices and targetTimezone in their profile instead of some sort of local storage.

PR #294, opened 2016-06-07
Verified clinic account experience

krystophv:
The interface for a verified clinic account to create and manage patient users effectively. 

Opening this PR for review and comparison purposes. To my knowledge, this is feature complete. This will need to be tested against the clinic backend work that should be on `stg` and currently use https://github.com/tidepool-org/platform-client/tree/krystophv/in-clinic-confirm for `platform-client`.

Disabling stuff during upload:
- [x] timezone selector
- [x] edit info
- [x] change devices
- [x] change person
- [x] logout


krystophv:
ping @jebeck for PR review.

jebeck:
Merge conflicts?

Also more of a UX question, but seems odd to land on the "Who are you uploader for?" with a fresh VCA who doesn't have access to anyone to upload. Was that a conscious decision made to keep things simple?

lib/pages/App.js
```diff
@@ -173,12 +203,41 @@ export default class App extends Component {
...
+      const { allUsers, page, targetUsersForUpload, uploadTargetUser } = this.props;
+      return (
+        <div>
+          <ClinicUserEdit
+            targetId={uploadTargetUser}
+            allUsers={allUsers}
+            loggedInUser={this.props.loggedInUser}
+            createUser={this.props.async.createCustodialAccount}
+            updateUser={this.props.async.clickEditUserNext}
+            cancelEdit={this.handleClickChangePerson.bind(this)}
```
jebeck: Move the binding into the constructor like we do for `handleClickChooseDevices` and `handleDismissDropdown`? For this and for `handleClickEditUser`, plz.

lib/pages/App.js
```diff
@@ -141,11 +159,18 @@ export default class App extends Component {
         );
     } else if (page === pages.MAIN) {
       const viewDataLink = _.get(this.props, ['blipUrls', 'viewDataLink'], '');
-      let timezoneDropdown = this.renderTimezoneDropdown();
+      let timezoneDropdown = null;
+      let viewDataLinkButton = this.renderUploadListDoneButton();
+      let clinicUserBlock = this.renderClinicUserBlock();
```
jebeck: This (line 164) is repeated below in line 192 (and perhaps more places). We can consolidate, can't we?

lib/components/ClinicUserEdit.js
```diff
@@ -0,0 +1,234 @@
...
+    }
+    return {
+      name: '',
+      year: '',
+      month: '',
+      day: ''
+    };
+  },
+
+  zeroPad: function(value){
```
jebeck: Isn't there an npm library for this kind of thing? 😉

lib/redux/utils/errors.js
```diff
@@ -74,6 +74,21 @@ export function getUpdateProfileErrorMessage(status) {
...
+export function getCreateCustodialAccountErrorMessage(status){
+  switch(status) {
+    case 400:
+      return 'Something looks funky, make sure this account info is correct.';
+    case 401:
+      return 'Your session timed out. You\'ll need to log back in.';
+    case 409:
+      return 'We can\'t create this account becuase that email address already has an account.';
+    case 500:
+      return 'Er sorry, we cant create this account right now. Try again in a few minutes. ';
```
jebeck: Typo, change to `can't`.

lib/redux/utils/errors.js
```diff
@@ -74,6 +74,21 @@ export function getUpdateProfileErrorMessage(status) {
...
+    default:
+      return 'Uh oh, we cant create this account. Try again in a few minutes and check your internet connection. ';
```
jebeck: Ditto.

lib/components/ClinicUserEdit.js
```diff
@@ -0,0 +1,234 @@
...
+    return (value.length === 1 ? '0': '') + value;
+  },
+
+  isDisabled: function() {
```
jebeck: This whole function is very dense. I'd be in favor of breaking it out for readability (and adding comments perhaps, for clarity). Was it a conscious decision to not provide user feedback re: things like date validation errors, or simply not addressed?

lib/components/ClinicUserEdit.js
```diff
@@ -0,0 +1,234 @@
...
+    var dateString = _.get(this,'refs.year.value')+'-'+_.get(this,'refs.month.value')+'-'+this.zeroPad(_.get(this,'refs.day.value', ''));
+    return (!_.get(this, 'refs.name.value')) || (!_.get(this,'refs.month.value')) ||
+      (!_.get(this,'refs.day.value')) || (!_.get(this,'refs.year.value')) ||
+      (!sundial.isValidDateForMask(dateString, 'YYYY-MM-DD')) ||
+      (sundial.dateDifference(new Date(), dateString, 'd') < 0);
```
jebeck: Is this last bit "Date cannot be in the future"?

lib/redux/reducers/users.js
```diff
@@ -87,6 +88,18 @@ export function updateProfileErrorDismissed(state = null, action) {
   }
 }
 
+export function createCustodialAccountErrorMessage(state = null, action) {
+  switch (action.type) {
+    case actionTypes.CREATE_CUSTODIAL_ACCOUNT_FAILURE:
+      const err = action.payload;
+      return err.message;
+    case actionTypes.CREATE_CUSTODIAL_ACCOUNT_REQUEST:
```
jebeck: This doesn't seem sufficient for clearing these errors. If I simulate an error on custodial account creation, I get the big red error rendered. I navigate away & back again, and it's still there. Not sure I would ever try again with that error showing. Was making the error dismissable through direct user action (rendering an 'x' in the corner) or dismissing it "under the hood" when user navigates away to another "page" discussed?

jebeck: Just discussed this with Sara. She agrees the error should be dismissable ('x' in the corner) and the form validation errors should be surfaced with the same UI, although those should also auto-dismiss when the form validates as good. LMK if you have any questions tomorrow @krystophv...can explain more fully if needed.

lib/components/ClinicUserEdit.js
```diff
@@ -0,0 +1,234 @@
...
+        profile.patient.mrn = mrn;
+      }
+
+      if(this.props.targetId){
+        this.props.updateUser(profile);
+      } else {
+        this.props.createUser(profile);
+      }
+    } else {
+      this.setState({errorMessage: 'Invalid birthday.'});
```
jebeck: This won't ever surface because you're also checking `sundial.isValidDateForMask` in the `isDisabled` for the submit button...

jebeck:
As an overall comment, cuz I just noticed, you could have written the new `Clinic*` components in ES6. Not sure if you felt you had to "match" older components. As long as we don't mix _within_ components, I think writing new ones in ES6 would be better in future, but LMK if you feel differently.

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+    allUsers: React.PropTypes.object.isRequired,
+    onUserChange: React.PropTypes.func.isRequired,
+    targetId: React.PropTypes.string,
+    targetUsersForUpload: React.PropTypes.array.isRequired,
+    onAddUserClick: React.PropTypes.func.isRequired,
+    setTargetUser: React.PropTypes.func.isRequired
+  },
+
+  handleChange: function(e) {
+    this.props.setTargetUser(e);
```
jebeck: Can't we just pass `this.props.setTargetUser` directly to `<Select />` as the `onChange`? Compare with `<UserDropdown />`...

Got on the track of looking at this because I found it pretty confusing at first, especially with `e` as the argument. I was like, "we're passing a _raw_ event to one of our action creators??" So at first I was gonna suggest just not using `e` when it's not a raw event, then I realized I think the whole wrapper function is unnecessary...

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+          {name} {mrn ? 'MRN:'+mrn : ''}
+        </div>
+        <div className={styles.optionLabelBirthday}>
+          {dateString}
+        </div>
+      </div>
+    );
+  },
+
+  renderError: function() {
```
jebeck: Why do we surface update profile errors here?

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+      <div className={classes} disabled={disabled} onClick={this.handleClickNext}>
+        Next
+      </div>
+    );
+  },
+
+  renderAddNew: function() {
+    var classes = cx({
+      [styles.addLink]: true,
+      [styles.valueSelected]: this.props.targetId
```
jebeck: Stylistically, I prefer wrapping the truthiness use of `this.props.targetId` here in a `Boolean`. How do you feel?

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+          <i className={styles.iconClose} onClick={this.props.dismissUpdateProfileError}></i>
+        </div>
+      );
+    }
+  },
+
+  renderSelector: function(){
+    var allUsers = this.props.allUsers;
+    var targets = this.props.targetUsersForUpload;
+    var sorted = _.sortBy(targets, function(targetId) {
```
jebeck: I assume this is the same as the sorting in `<UserDropdown />`. Be great to refactor into some kind of utility.

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+      var targetInfo = allUsers[targetId];
+      if (_.get(targetInfo, ['patient','isOtherPerson'])) {
+        return targetInfo.patient.fullName;
+      }
+      return targetInfo.fullName;
+    });
+
+    var selectorOpts = _.map(sorted, function(targetId) {
+      var targetInfo = allUsers[targetId];
+      var mrn = _.get(targetInfo, ['patient', 'mrn'], '') ?  ' ' +_.get(targetInfo, ['patient', 'mrn']) : '';
```
jebeck: Doing the same `_.get` here makes this a bit tough to read...

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+      if (_.get(targetInfo, ['patient','isOtherPerson'])) {
+        return {value: targetId, label: targetInfo.patient.fullName};
```
jebeck: Do "child" accounts not have MRNs?

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+  handleClickNext: function(e) {
+    e.preventDefault();
+    if(this.props.targetId){
+      this.props.onUserChange(this.props.targetId);
+    }
+  },
+
+  valueRenderer: function(option) {
+    var user = _.get(this.props.allUsers, option.value);
+    var name = _.get(user, 'fullName');
```
jebeck: Here you are implicitly assuming the `user` won't be a legacy fake "child" account...

lib/components/ClinicUserSelect.js
```diff
@@ -0,0 +1,162 @@
...
+        valueRenderer={this.valueRenderer} />
+    );
+  },
+
+  renderButton: function() {
+    var classes = cx({
+      [styles.button]: true,
+      disabled: !this.props.targetId
+    });
+    var disabled = !this.props.targetId ? 'disabled' : '';
```
jebeck: Wouldn't it be a bit cleaner to get rid of this `disabled` var and just do `disabled={!this.props.targetId}`? I believe React JSX transpiler now does the transformation from `true` -> "disabled" string automatically.

jebeck:
I just added a checklist in the PR comments of all the stuff to make sure is disabled during upload (both clinic version and non-clinic version, if applicable). I'll stop reviewing for now...I think there's a lot here and maybe some stuff to discuss.

lib/components/ClinicUserEdit.js
krystophv: It is dense. It started out simple but kept growing and I didn't break it out when I probably should have. It was a conscious decision, mostly because it's going to run this on every form onChange event (every keystroke, essentially) and I personally hate it when forms tell me my input is invalid when I'm in the middle of actually typing it it.

lib/components/ClinicUserEdit.js
krystophv: Yup

lib/components/ClinicUserEdit.js
krystophv: Defensive coding habits die hard I suppose.

lib/components/ClinicUserSelect.js
krystophv: I do believe you are correct. I can't remember what my train of thought was for the wrapping function.

lib/components/ClinicUserSelect.js
krystophv: We don't. The prop is never passed in. This is a remnant from having used `TimezoneDropdown` to template this file out real fast and this function was overlooked in replacing all the relevant code. I'll make it gone.

lib/components/ClinicUserSelect.js
krystophv: I generally like to use `!!` to coerce to `Bool`.

lib/components/ClinicUserSelect.js
krystophv: MRN is optional, so any account may not have an MRN. We do expect clinics to be interacting with the old fake child accounts until we come up with and implement the migration from the old child accounts to custodial account relationships where it's relevant. Even then, the user might take ownership of their account and it will be much like the current fake child accounts (if I understand that correctly).

lib/components/ClinicUserSelect.js
krystophv: I'm not sure I fully understand the implication of that.

lib/components/ClinicUserSelect.js
krystophv: Yup.

lib/components/ClinicUserSelect.js
jebeck: Well it's just in other places you were taking pains to account for legacy child accounts. If we need to do that (and it sounds like you believe we expect clinics to potentially still be interacting with the old accounts for a while), then we need to be consistent about it. At this point it's really ripe for factoring out something like blip's `personUtils` (I think that's what we call them) to pass in a user and get the appropriate `fullName` no matter what kind of account it is.

lib/components/ClinicUserSelect.js
jebeck: No, I'm not under the impression that taking ownership of an account will make new accounts like old fake child accounts at all. They have their own `userId`s from the beginning (unlike fake child accounts) and won't have the `isOtherPerson` business going on.

lib/components/ClinicUserEdit.js
jebeck: The problem is that *no* invalid form information ever gets surfaced to the user because an invalid form disabled the submit button, and it's on submit that you actually try to set state for errors re: invalid form state. The invalid date thing seems the only thing worth surfacing, perhaps we could try separating out that validation in onChange and setting state there, then run it by Sara?

lib/components/ClinicUserSelect.js
krystophv: It is, but now that the name extraction is factored into personUtils, the sort is a one-liner. Seems like overkill to refactor into a utility to have a one liner repeated a total of two times.

krystophv:
@jebeck I think I've addressed all the comment concerns now if you want to resume the review.

lib/components/ClinicUserEdit.js
krystophv: turns out lodash has one built in: https://github.com/lodash/lodash/blob/3.10.1/doc/README.md#_padleftstring-length0-chars-

jebeck:
Tiny thing: cursor when hovering over Logout during upload doesn't give you the "forbidden" feedback.

lib/core/personUtils.js
```diff
@@ -0,0 +1,34 @@
...
+*
+* You should have received a copy of the License along with this program; if
+* not, you can obtain one from Tidepool Project at tidepool.org.
+*/
+
+var _ = require('lodash');
+
+var personUtils = {};
+
+personUtils.patientFullName = function(person) {
```
jebeck: Did you copy this exactly from blip? Because the way we organize the user info in `allUsers` is not identical (apparently), and it doesn't work here... (I noticed because in the "home" side of the experience a "fake" child account shows up in the user dropdown with the parent's name instead of the patient name.)

A couple of unit tests would be good as well.

jebeck:
screenshot 2016-06-17 10 47 17

Small thing, probably shouldn't a required prop since user may not be selected.

jebeck:
The disabling of stuff in the clinic experience looks good 👍 

One style nitpick: do we have a disabled state for the Change Person, Edit Info, Change Devices links to provide a little more user feedback? Or at least turning the hand pointer into the forbidden perhaps?

lib/redux/actions/async.js
```diff
@@ -155,6 +155,13 @@ export function doLogin(creds, opts) {
         profile: results[1],
         memberships: results[2]
       }));
+
+      // detect if a VCA here and redirect to clinic user select screen
+
+      if(actionUtils.userHasRole(results[0].user, 'clinic')){
```
jebeck: Perhaps we should unify the `userHasRole` function under the new `personUtils`? I think it makes more sense to keep everything relating to info we're extracting from a user object in one place.

lib/redux/actions/async.js
```diff
@@ -463,6 +470,43 @@ export function clickDeviceSelectionDone() {
   };
 }
 
+export function clickEditUserNext(profile) {
+  return (dispatch, getState) => {
+    const { uploadTargetUser } = getState();
+    const { api } = services;
+    dispatch(syncActions.updateProfileRequest());
```
jebeck: Maybe move this inside the following `if`? After all, we're not making a request if the profile is empty...

lib/redux/actions/async.js
```diff
@@ -463,6 +470,43 @@ export function clickDeviceSelectionDone() {
...
+            dispatch(syncActions.setAllUsers(results[0],results[1],results[2]));
+            const { targetDevices, devices, allUsers, loggedInUser } = getState();
+            const targetedDevices = _.get(targetDevices, uploadTargetUser, []);
+            const supportedDeviceKeys = Object.keys(devices);
+            const atLeastOneDeviceSupportedOnSystem = _.some(targetedDevices, (key) => {
+              return _.includes(supportedDeviceKeys, key);
+            });
+            if (_.isEmpty(targetedDevices) || !atLeastOneDeviceSupportedOnSystem) {
+              return dispatch(syncActions.setPage(pages.SETTINGS));
+            } else {
```
jebeck: Hmm, this is interesting. This action creator should only be available to clinics anyway, right? But say we later expose it to home users as well, then we may end up with unexpected behavior (not updating the "page"). I guess what I'm saying is, wouldn't it be fine to get rid of the `if` on line 499 and just always redirect to `pages.MAIN` in the `else`?

lib/pages/App.js
```diff
@@ -173,12 +213,43 @@ export default class App extends Component {
...
             onDone={this.props.async.clickDeviceSelectionDone} />
         </div>
       );
+    } else if (page === pages.CLINIC_USER_SELECT) {
+      const { allUsers, page, targetUsersForUpload, uploadTargetUser } = this.props;
+      return (
+        <div>
+          <ClinicUserSelect
+            allUsers={allUsers}
+            onUserChange={this.props.async.setUploadTargetUserAndMaybeRedirect}
```
jebeck: I'm not sure reusing `setUploadTargetUserAndMaybeRedirect` is really the best approach here. We end up dispatching SET_UPLOAD_TARGET_USER twice and the redirect logic is more complex. Let's discuss.

lib/redux/actions/async.js
```diff
@@ -557,6 +625,17 @@ export function setUploadTargetUserAndMaybeRedirect(targetId) {
     });
     if (_.isEmpty(targetedDevices) || !atLeastOneDeviceSupportedOnSystem) {
       return dispatch(syncActions.setPage(pages.SETTINGS));
+    } else {
+      if (actionUtils.userHasRole(_.get(allUsers, loggedInUser), 'clinic')) {
```
jebeck: Similar comment as above for `clickEditUserNext` but also see comment about not reusing this action creator in the clinic case...I think we could simplify by having something separate, especially as I think also the `setBlipViewDataUrl` part is unnecessary? Am I right that there's no "Go to Blip" button in the clinic flow?

lib/redux/actions/async.js
```diff
@@ -538,13 +582,37 @@ export function retrieveTargetsFromStorage() {
...
+export function createCustodialAccount(profile) {
+  return (dispatch, getState) => {
+    const { api } = services;
+    dispatch(syncActions.createCustodialAccountRequest());
+    api.user.createCustodialAccount(profile, (err, account) => {
+      if (err) {
+        dispatch(syncActions.createCustodialAccountFailure(err));
+      }
+      else {
+        dispatch(syncActions.createCustodialAccountSuccess(account));
```
jebeck: So you're passing in the new `account` but there's no `payload` in the success action... If that's because we're refetching everything (next `async.series`), then perhaps don't include the param (for less confusion) and make a comment?

lib/redux/actions/sync.js
```diff
@@ -492,6 +507,33 @@ export function updateProfileFailure(err) {
...
+ */
+
+export function createCustodialAccountRequest() {
+  return {
+    type: actionTypes.CREATE_CUSTODIAL_ACCOUNT_REQUEST,
+    meta: {source: actionSources[actionTypes.CREATE_CUSTODIAL_ACCOUNT_REQUEST]}
+  };
+}
+
+export function createCustodialAccountSuccess() {
```
jebeck: Commented above on the lack of a payload here...I find that pretty strange in a success action. Would like to understand a bit more why the refetching is necessary instead of just updating the state using the response, which could be put into a `payload`?

lib/redux/actions/async.js
krystophv: I think you're right, now that I'm looking at it. I can't remember the reason I had that wrapped in the `if`. If I remember what my logic was, I can add it back in later. I'll take it out for now.

lib/redux/actions/async.js
krystophv: It is because we refresh everything, but I think it actually makes more sense for that success action to carry the payload of the newly created account (even if we're not currently listening for/using it anywhere yet), so I'll add the payload to that action.

lib/redux/actions/async.js
krystophv: This is actually what got me to put that `if` in above. It is necessary here because we don't want to redirect users in the non-clinic case since they'll already be on the `MAIN` page when they're using the user dropdown, but clinic users are on the `CLINIC_USER_SELECT` page and need to be redirected.

I think you may be correct that this could be simplified by using a dedicated action creator for the clinic experience.

lib/redux/actions/async.js
krystophv: And it does seem, upon reflection that the full refresh is probably just hitting the problem with a large hammer. I'll see about just updating `allUsers` appropriately.

krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;vmO6zxzn&#x2F;1-ux-2-dev-3-tst-2-uploader-in-clinic-integrate-new-back-end-for-in-clinic-account-creation"> &lt;ux:2, dev:3, tst:2&gt; [uploader] [in-clinic] integrate new back-end for in-clinic account creation</a>

jebeck:
I'd still like to see a couple of unit tests for the `personUtils` please.

lib/pages/App.js
jebeck: [This comment](https://github.com/tidepool-org/chrome-uploader/pull/294#discussion_r67573104) hasn't been addressed yet, it looks like. If we discussed and decided not to change anything, remind me, cuz I don't remember.

jebeck:
This isn't in the diff because it dates from #280 (I think), but I'm confused about why the test w/the description "new target user has not selected timezone" (testing for a redirects to SETTINGS, not MAIN) is passing...

jebeck:
Good to go as soon as the things we've discussed are addressed, as well as the explanation on that test that confuses me as to why its passing...

PR #295, opened 2016-06-07
CGM Support for Animas Vibe

gniezen:
See https://trello.com/c/cRSsoyes for details.


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;cRSsoyes&#x2F;12-ux-0-dev-3-2-tst-2-uploader-animas-vibe-cgm-data"> &lt;ux:0 dev:3+2 tst:2&gt; [uploader] Animas Vibe CGM data</a>

lib/drivers/animasDriver.js
```diff
@@ -1131,6 +1197,33 @@ module.exports = function (config) {
     });
   };
 
+  var readTimeAdjust = function(cb) {
+    if (modelNumber === 'IR1295') {
```
jh-bate: constant? used more than once

lib/drivers/animasDriver.js
```diff
@@ -422,6 +442,52 @@ module.exports = function (config) {
     };
   };
 
+  var requestCalibration = function() {
+    return{
+      recordType: RECORD_TYPES.CGM_CALIBRATION,
+      parser : function (payload) {
+        var dt = decodeDate(struct.unpack(payload,4,'bbbb',['monthYear','day','hour','minute']));
```
jh-bate: we decode the date in this same format twice in the new code - is it worth having a specialisation of the function that just assumes that format? maybe not but just wanted to put it out there

lib/drivers/animasDriver.js
```diff
@@ -1647,6 +1749,79 @@ module.exports = function (config) {
     return postrecords;
   }
 
+  function buildCBGRecords(data, postrecords) {
+    data.bgRecords.forEach(function (event) {
+      var record = cfg.builder.makeDeviceEventCalibration();
+      record.with_value(event.value)
+            .with_deviceTime(event.deviceTime)
+            .with_units('mg/dL')      // hard-coded in mg/dL
```
jh-bate: do we need to clarify as to why its hardcoded?

lib/drivers/animasDriver.js
```diff
@@ -1647,6 +1749,79 @@ module.exports = function (config) {
...
+            .set('index', event.index);
+      cfg.tzoUtil.fillInUTCInfo(record, event.jsDate);
+      record = record.done();
+      postrecords.push(record);
+    });
+
+    var timeChanges = 0;
+
+    data.cbgRecords.forEach(function (event) {
```
jh-bate: being paranoid but just wondering in these situations can the details in the event ever be null? e.g. `event.cbg_types` or `event.glucose`?

lib/drivers/animasDriver.js
```diff
@@ -1779,6 +1954,7 @@ module.exports = function (config) {
         readSerialandModelNumber,
         readBasalProgramNames,
         readBasalPrograms,
+        readTimeAdjust
```
jh-bate: comma missing?

jh-bate:
@gniezen some comments that may or may not be relevant but if you could let me know. Ta Jamie

lib/drivers/animasDriver.js
gniezen: Will do, as it's also easier to read PING and VIBE instead of 'IR1295' and 'IR1285'.

lib/drivers/animasDriver.js
gniezen: `event.cbg_types` will always be an array, possibly empty but never null. If `event.glucose` is null, it will get filtered out by the if statement in line 1768. The only other property is `event.rawTimeStamp`, for which I'll now add a null check.

lib/drivers/animasDriver.js
gniezen: Actually, we're always decoding it in the same format (9 times), so good catch! I'm going to move the format into the `decodeDate` function itself. (The only difference is where in the payload it starts, so I'll pass the payload and index as parameters)

lib/drivers/animasDriver.js
gniezen: No? Don't think trailing commas in an array is a good idea.

jh-bate: yeah was just looking at the compare and the previous one did so thought I would double check

jh-bate:
LGTM @gniezen

PR #296, opened 2016-06-13
Tandem multi-record issue

gniezen:
It's possible that's there a max limit on how many records can be read with a multi-record request, even though this is not mentioned in the spec. This PR reads 7000 records at a time.

@jebeck, will you please check if it resolves the issue at all, and then review?


jebeck:
No dice yet, on 3x tries 🙁

gniezen:
Closing this PR, as it did not resolve the issue.

PR #298, opened 2016-06-15
Tandem multi-record support

gniezen:
After reverting #289 because it failed on a pump with firmware 54676, I've now rebased the work using `git rebase -i HEAD~9` and a `rework` on the first commit, and then did a `git merge master`. 

This is still WIP, as we're waiting on feedback from Tandem on why it failed on this specific pump.


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OIiGU5jG&#x2F;8-ux-0-dev-2-tst-2-uploader-device-implement-tandem-multi-record-download-commands"> &lt;ux:0, dev:2, tst:2&gt; [uploader] [device] implement Tandem multi-record download commands</a>

gniezen:
Turns out that at least one firmware version after 47144 does not actually support multi-record requests, so we're falling back to regular event history requests when the multi-record request fails.

@jebeck - would you please review the new changes, check if the upload (fallback) succeeds on your pump and comment to that effect on the card (linked above)?

lib/drivers/tandemDriver.js
```diff
@@ -1221,8 +1230,13 @@ module.exports = function (config) {
 
     function iterate(err, result) {
       if (err) {
-        debug('error retrieving record ', result);
-        callback(err, null);
+        if(err.name === 'multi-fail') {
```
jebeck: Do you assign `name`s to errors elsewhere in this driver or in other drivers? Just wondering if it's a standard. It seems pretty much as arbitrary as matching the `error.message` to me, although certainly more concise.

jebeck:
LGTM 🚢

lib/drivers/tandemDriver.js
gniezen: Yes, we're also using `err.name` in the Bayer driver. My understanding is that each `Error` [object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) contains both a `name` and `message` property.

PR #299, opened 2016-06-16
Bug fix: Abbott FreeStyle meters with a large number of records

gniezen:
For more details: https://trello.com/c/WpikEtp1


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;WpikEtp1&#x2F;20-uploader-bug-abbott-devices-with-large-number-of-records-fail-to-upload"> [uploader][bug] Abbott devices with large number of records fail to upload</a>

gniezen:
Pinging @jebeck for review?

jebeck:
So the probing and the `getSerialNumber` were the two extraneous reads?

LGTM.

gniezen:
@jebeck Yep, there's basically just one command you can send to this device (`mem`) and it returns all data, including serial number. (This is different from the PXtra, where you have two different commands). If 400 records take 20s to download, sending the same command three times (probing, get serial number, get data) would take 3 x 20s = 60s.

PR #300, opened 2016-07-11
Dexcom device version check

gniezen:
See https://trello.com/c/7WRyBmpw for more details

### DIFF
--- a/lib/drivers/dexcomDriver.js
+++ b/lib/drivers/dexcomDriver.js
@@ -26,6 +26,7 @@ var struct = require('../struct.js')();
 
 var mungeUserSettings = require('../dexcom/userSettingsChanges');
 var TZOUtil = require('../TimezoneOffsetUtil');
+var errorText = require('../redux/constants/errors');
 
 var debug = require('../bows')('DexcomDriver');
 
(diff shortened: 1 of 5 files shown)


gniezen:
Pinging @jh-bate for review. This is medium time-critical as we have access to a G6 receiver this week only - luckily this is a small change.

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7WRyBmpw&#x2F;19-uploader-dexcom-version-check"> [uploader] Dexcom version check</a>

lib/drivers/dexcomDriver.js
```diff
@@ -285,12 +286,19 @@ module.exports = function (config) {
     var all = [];
     var ctr = 0;
 
-    var format = 'iisbs';
-    var recordformat = ['systemSeconds', 'displaySeconds', 'glucose', 'trendArrow', 'crc'];
+    var format;
+    var recordformat;
+    debug('EGV header revision:',header.revision);
```
jh-bate: do you still want this here?

lib/redux/constants/errors.js
```diff
@@ -30,5 +30,6 @@ module.exports = {
   E_READ_FILE: 'Error reading file ',
   E_SERIAL_CONNECTION: 'Hmm, we couldn\'t detect your device',
   E_SERVER_ERR: 'Sorry, the Tidepool servers appear to be down',
-  E_UPLOAD_IN_PROGRESS: 'Sorry, an upload is already in progress'
-};
\ No newline at end of file
+  E_UPLOAD_IN_PROGRESS: 'Sorry, an upload is already in progress',
+  E_DEXCOM_UNSUPPORTED: 'Sorry, we don\'t support this Dexcom device yet'
```
jh-bate: as we have been talking about language and errors does this need to be checked off do you think? Or should it say the version we don't support? Being picky but just wanted to check

jh-bate:
@gniezen two queries but once those are answered LGTM

lib/drivers/dexcomDriver.js
gniezen: Yes, this will be useful for debugging in case there are unexpected revision numbers that should pass (In the unlikely case that a G4 receiver has a revision number larger than 4.)

lib/redux/constants/errors.js
gniezen: This is the first time that I'm adding an error message that only appears if you click on Error Details to this list, instead of just in-line. `Something went wrong during device upload` is the higher-level "checked off" error message that gets displayed. I don't think we should get sign off for all error detail messages. I'm not even entirely sure I should be putting it in this file - maybe we need a separate file for error detail messages?

lib/drivers/dexcomDriver.js
gniezen: On second thought, this is too chatty as it will appear for every EGV record - will remove.

jh-bate:
LGMT @gniezen

PR #302, opened 2016-07-18
Animas bug fix: Final basal when rate not yet zero after suspend

gniezen:
Fixes an edge case where at the time of upload the basal rate hasn't changed to zero yet. (It takes up to three minutes for the basal rate to change after a command is issued on the pump). 

This PR calculates the duration of the final basal based on the time of the suspend event, so that it can be displayed.

### DIFF
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -153,7 +153,7 @@ exports.make = function(config){
           // categorize the basal we're processing as a suspend basal if its rate
           // is zero and its start falls within the suspend event we have on deck
           var resumed = Date.parse(currStatus.time) + currStatus.duration;
-          if (resumed > Date.parse(event.time)) {
+          if (resumed >= Date.parse(event.time)) {
             //this basal is suspended
             event.deliveryType = 'suspend';
             delete event.rate;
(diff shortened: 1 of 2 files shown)


gniezen:
@jebeck will you be able to review?

lib/animas/animasSimulator.js
```diff
@@ -153,7 +153,7 @@ exports.make = function(config){
           // categorize the basal we're processing as a suspend basal if its rate
           // is zero and its start falls within the suspend event we have on deck
           var resumed = Date.parse(currStatus.time) + currStatus.duration;
-          if (resumed > Date.parse(event.time)) {
+          if (resumed >= Date.parse(event.time)) {
```
jebeck: Why the change to `>=` here? I'm not quite following.

lib/drivers/animasDriver.js
```diff
@@ -2085,6 +2085,7 @@ module.exports = function (config) {
             debug('[Hand-off to simulator] Unhandled type!', datum.type);
         }
       }
+      simulator.finalBasal();
```
jebeck: So, thinking out loud here to check my understanding... We didn't have a `finalBasal` method in the simulator before now analogous to the method used for OmniPod & Tandem (looking up settings as of upload and guessing at duration of basal) because you have to suspend the pump in order to upload an Animas, correct? And in the "normal" case (not the case this PR is about), the pump *has* already gone to rate `0` by upload and so that suspend segment would be the final basal, and so we don't have any way to guess about the duration?

lib/animas/animasSimulator.js
gniezen: Not completely relevant to this PR, but as it was a one-character change I didn't want to create a whole separate PR. Basically, since we only have one-minute resolution, the suspend/resume event and the basal can stop at exactly the same time (but they can also be up to three minutes apart).

lib/drivers/animasDriver.js
gniezen: Exactly. 💯

lib/animas/animasSimulator.js
jebeck: Ah, OK. I'm glad it's not completely related because that's what was confusing me most - trying to see the connection with the other changes.

jebeck:
My questions have been answered, so :shipit:

gniezen:

PR #303, opened 2016-07-19
Animas bug fix: Some uploads stalling at 40%

gniezen:
During alpha testing some uploads are stalling at 40%. This is caused by an uncaught error where we are trying to parse a duplicate (empty) packet. This PR:

- makes sure we catch these errors and return them in the UI
- checks for duplicate packets across different requests, not only in each request


gniezen:
Self-merging, as it's not going into master yet.

PR #304, opened 2016-07-21
Remaining work on Animas

gniezen:
See https://trello.com/c/5Q1QHuWU for details.


gniezen:

lib/drivers/animasDriver.js
```diff
@@ -577,23 +581,6 @@ module.exports = function (config) {
     return packet;
   };
 
-  var parsePayload = function (packet) {
```
gniezen: Unused function, removed.

gniezen:
Pinging @jebeck for review.

lib/animas/animasSimulator.js
```diff
@@ -261,14 +265,23 @@ exports.make = function(config){
               }
             }
           }
+
+          // Filter out dates 2008 and earlier, as this is year 0 for Vibe.
+          // We are including 2008 as it's possible that the pump ran for a
```
darinkrauss: It says we are including 2008, but the code is `<= 2008` is dropped.  To me, I read this as 2008 is NOT included.  Or am I misreading the documentation.

lib/animas/animasSimulator.js
```diff
@@ -204,7 +206,7 @@ exports.make = function(config){
       if(currSMBG != null && currSMBG.value === event.value) {
         console.log('Duplicate SMBG value (',event.value,')',currSMBG.subType,currSMBG.time,'/',event.subType,event.time);
         var duration = Date.parse(event.time) - Date.parse(currSMBG.time);
-        if ((duration < (15 * sundial.MIN_TO_MSEC)) && (event.subType === 'manual')) {
+        if ((duration < (15 * sundial.MIN_TO_MSEC)) && (event.subType === 'manual') && (currSMBG.subType === 'linked')) {
```
darinkrauss: What does 'linked' mean here?

lib/drivers/animasDriver.js
```diff
@@ -1759,7 +1760,7 @@ module.exports = function (config) {
         postrecords.push(bgRecord);
       }
       else {
-        console.log('Discarding control solution test');
+        debug('Discarding control solution test');
```
darinkrauss: Should some other of these `debug` statements be wrapped in a `if (verbose) {}` clause?

lib/drivers/animasDriver.js
```diff
@@ -1775,6 +1776,19 @@ module.exports = function (config) {
...
       postrecords.push(record);
+
+      // Since Animas does not generate a separate SMBG event for Vibe, we use the
+      // calibration event from the pump to also generate a manual SMBG record
+      // in order to display it in Blip
+      var bgRecord = cfg.builder.makeSMBG()
+        .with_deviceTime(event.deviceTime)
+        .with_subType('manual')
+        .with_value(event.value)
+        .with_units('mg/dL')
```
darinkrauss: Is this always `mg/dL`?

darinkrauss:
A few minor comments, but otherwise LGTM. :shipit:

lib/animas/animasSimulator.js
gniezen: I meant that we're including 2008 in the filter - I understand how that can be misunderstood, will change. And yes, the rest of your explanation is correct, so I'll include it as part of the comments for future reference (parts of it verbatim if you don't mind ;)

lib/drivers/animasDriver.js
gniezen: I'm only including the `debug` statements that are very chatty in the `verbose` clause. The ones that are not wrapped don't appear that often, and could provide valuable debugging info when on prod.

lib/drivers/animasDriver.js
gniezen: According to the spec, yes. Would love to see if that is indeed the case on a `mmol/L` pump too, but we don't have any to verify.

lib/animas/animasSimulator.js
gniezen: It refers to a linked meter, i.e., value was transferred directly from the meter to the pump: http://developer.tidepool.io/data-model/device-data/types/smbg.html#subtype

For Animas pumps, this is only supported by Ping pump when used with Ping meter.

PR #305, opened 2016-07-22
Fabricate and embed suspend/resume events for alarms

gniezen:
See https://trello.com/c/3bOHXVaf for details.


gniezen:
Pinging @jebeck for review. We're planning on testing this PR on Thursday, so #304 takes priority over this one.

gniezen:
@darinkrauss Will you also test with this branch to see if the embedded status event is now created correctly for the alarm on your pump?

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;3bOHXVaf&#x2F;4-ux-0-dev-2-tst-2-uploader-fabricate-and-embed-suspend-resume-events-for-alarms"> &lt;ux:0 dev:2 tst:2&gt; [uploader] Fabricate and embed suspend&#x2F;resume events for alarms</a>

darinkrauss:
@gniezen Yes, I can confirm that latest version of this branch generates an embedded `deviceEvent/status` datum in a related `deviceEvent/alarm` datum.

test/node/animas/testAnimasSimulator.js
```diff
@@ -493,7 +493,7 @@ describe('animasSimulator.js', function() {
         type: 'deviceEvent',
         subType: 'status',
         status: 'suspended',
-        reason: {suspended: 'automatic', resumed: 'automatic'},
+        reason: {suspended: 'automatic', resumed: 'manual'},
```
gniezen: @jebeck Just want to make sure I'm using the nomenclature for `reason` correctly here. For the fabricated suspend/resume events I'm marking `suspended` as `automatic` (since the alarm initiated the suspend), but I mark `resumed` as `manual`, as the user needs to perform some action before the pump will resume (even though it does not necessarily mean clicking a "resume" button).

jebeck: Yeah, that seems correct to me.

jebeck:
As an overall note, since the number of basal records is quite limited on this pump and you're fabricating the status events inside the basal stream, we won't actually fabricate a `suspended` for alarms that fall outside the date range of the basal records, correct? Is this a shortcoming you'd already thought about?

jebeck:
We have enough information to build the suspend but not the resume, technically.

PR #306, opened 2016-07-25
Redux mock store updates

krystophv:
Updates to unit tests to conform to the new redux-mock-store API and move the `async` library calls out of action creators and into the API layer so that they can be stubbed out to work with the new redux-mock-store. Also took the opportunity to rework `clickEditUserNext` to pass the updates through `updateProfileSuccess` properly instead of a larger state tree refresh.

### DIFF
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -94,6 +94,14 @@ api.makeBlipUrl = function(tail) {
 
 api.user = {};
 
+api.user.initializationInfo = function(cb){
+  async.series([
+    api.user.account,
+    api.user.loggedInProfile,
+    api.user.getUploadGroups
+  ], cb);
+};
+
 api.user.login = function(user, options, cb) {
   api.log('POST /auth/login');
 
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -115,7 +115,7 @@ api.user.login = function(user, options, cb) {
 
 api.user.loginExtended = function(user, options, cb){
   async.series([
-    api.user.login(user, options),
+    api.user.login.bind(null, user, options),
     api.user.loggedInProfile,
     api.user.getUploadGroups
   ], cb);
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -161,12 +161,12 @@ api.user.updateProfile = function(userId, updates, cb){
     if(err){
       return cb(err);
     }
-    var patient = _.get(profile, 'patient', {});
-    var patient_updates = _.get(updates, 'patient', {});
-    var new_patient = _.assign(patient, patient_updates);
-    var new_profile = _.merge(profile, updates);
-    new_profile.patient = new_patient;
-    return api.user.addProfile(userId, new_profile, cb);
+    var newProfile = _.merge(profile, updates, function(original, update){
+      if (_.isArray(original)) {
+        return update;
+      }
+    });
+    return api.user.addProfile(userId, newProfile, cb);
   });
 };
 
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -33,6 +33,7 @@ var errorText = require('../redux/constants/errors');
 var log = isChromeApp ? bows('Api') : console.log;
 var builder = require('../objectBuilder')();
 var localStore = require('./localStore');
+var actionUtils = require('../redux/actions/utils');
 
 // for cli tools running in node
 if (typeof localStore === 'function') {
@@ -161,11 +162,7 @@ api.user.updateProfile = function(userId, updates, cb){
     if(err){
       return cb(err);
     }
-    var newProfile = _.merge(profile, updates, function(original, update){
-      if (_.isArray(original)) {
-        return update;
-      }
-    });
+    var newProfile = actionUtils.mergeProfileUpdates(profile, updates);
     return api.user.addProfile(userId, newProfile, cb);
   });
 };
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -113,6 +113,14 @@ api.user.login = function(user, options, cb) {
   });
 };
 
+api.user.loginExtended = function(user, options, cb){
+  async.series([
+    api.user.login(user, options),
+    api.user.loggedInProfile,
+    api.user.getUploadGroups
+  ], cb);
+};
+
 api.user.account = function(cb) {
   api.log('GET /auth/user');
   tidepool.getCurrentUser(cb);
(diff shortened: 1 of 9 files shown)


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;exbgszKx&#x2F;2-ux-0-dev-3-tst-1-uploader-improve-unit-testing-in-uploader"> &lt;ux:0 dev:3 tst:1&gt; [uploader] improve unit testing in uploader</a>

krystophv:
ping @jebeck for PR review

lib/redux/actions/async.js
```diff
@@ -428,19 +431,31 @@ export function doVersionCheck() {
...
       // TODO: remove conditional when perms are finalized or accounts are converted
-      if(err && _.get(err,'status') !== 401){
-        dispatch(syncActions.updateProfileFailure(err));
+      if (err){
+        if (_.get(err,'status') !== 401) {
+          dispatch(syncActions.updateProfileFailure(err));
+        } else {
+          let profile = allUsers[userId];
+          let patient = _.get(profile, 'patient', {});
+          let patient_updates = _.get(updates, 'patient', {});
```
jebeck: We don't use snake case anywhere but the drivers side of the uploader, I believe. Can you redo these variables with a style more consistent with the rest of the UI side of the code (camelCase)?

lib/redux/actions/async.js
```diff
@@ -428,19 +431,31 @@ export function doVersionCheck() {
...
       // suppress unauthorized error until custodial vs normal permissions are ironed out
       // TODO: remove conditional when perms are finalized or accounts are converted
-      if(err && _.get(err,'status') !== 401){
-        dispatch(syncActions.updateProfileFailure(err));
+      if (err){
+        if (_.get(err,'status') !== 401) {
+          dispatch(syncActions.updateProfileFailure(err));
+        } else {
+          let profile = allUsers[userId];
+          let patient = _.get(profile, 'patient', {});
```
jebeck: `patient` appears to be unused, no?

lib/redux/actions/async.js
```diff
@@ -428,19 +431,31 @@ export function doVersionCheck() {
...
+          let patient_updates = _.get(updates, 'patient', {});
+          let new_patient = _.assign(profile.patient, patient_updates);
```
jebeck: I don't believe you're using this `new_patient` either. This whole sequence is really hard to understand. Can you pare it down?

lib/redux/actions/async.js
```diff
@@ -449,21 +464,32 @@ export function setTargetTimezone(userId, timezoneName) {
...
+      api.user.updateProfile(uploadTargetUser, updates, (err, profile) => {
         // suppress unauthorized error until custodial vs normal permissions are ironed out
         // TODO: remove conditional when perms are finalized or accounts are converted
-        if(err && _.get(err,'status') !== 401){
-          dispatch(syncActions.updateProfileFailure(err));
+        if (err) {
+          if (_.get(err,'status') !== 401) {
+            dispatch(syncActions.updateProfileFailure(err));
+          } else {
+            let profile = allUsers[uploadTargetUser];
```
jebeck: Ditto above and would like to see it refactored if, as I suspect, it's common code. Would like to review again to make sure I understand what you're doing in the final version...

lib/redux/actions/async.js
```diff
@@ -473,33 +499,40 @@ export function clickDeviceSelectionDone() {
...
       api.user.updateProfile(uploadTargetUser, profile, (err, profile) => {
+        // suppress unauthorized error until custodial vs normal permissions are ironed out
+        // TODO: remove conditional when perms are finalized or accounts are converted
         if (err) {
-          dispatch(syncActions.updateProfileFailure(err));
+          if(_.get(err,'status') !== 401) {
+            return dispatch(syncActions.updateProfileFailure(err));
+          } else {
+            const { allUsers } = getState();
+            let profile = allUsers[uploadTargetUser];
```
jebeck: Ditto.

test/browser/redux/reducers/users.test.js
```diff
@@ -90,6 +90,19 @@ describe('users', () => {
...
+    it('should handle UPDATE_PROFILE_SUCCESS', () => {
+      const action = {
+        type: actionTypes.UPDATE_PROFILE_SUCCESS,
+        payload: { profile, userId: 'a1b2c3' }
+      };
+      expect(users.allUsers(undefined, action)).to.deep.equal({
+        a1b2c3: profile
+      });
+      let initialState = {};
+      // test to be sure not *mutating* state object but rather returning new!
```
jebeck: We have the object invariant test helper for this. Haven't used it much in the uploader, that's still a TODO, but just note that tests for mutation should *not* be written like this in general. Examples for how to use the object invariant test helper are in blip and in a POC here in misc.test.js

krystophv:
@jebeck - updated concerning parts. 

To try and explain the strange profile update merge logic, I was trying to accomplish using the `_.merge` behavior for everything except array values (such as the `profile.email` and `profile.patient.targetDevices`) where we want to replace the array instead of replacing/augmenting individual array indices (which is what `_.assign` was doing). It was in a strange order because both `_.assign` and `_.merge` modify the original arguments and I needed to avoid modifying some properties till other's had been merged.  I looked harder at the `_.merge` API in the version of lodash we're using and it provides for a customizer function which I've switched over to using and the code is much clearer as to it's behavior.

lib/core/api.js
```diff
@@ -161,12 +161,12 @@ api.user.updateProfile = function(userId, updates, cb){
     if(err){
       return cb(err);
     }
-    var patient = _.get(profile, 'patient', {});
-    var patient_updates = _.get(updates, 'patient', {});
-    var new_patient = _.assign(patient, patient_updates);
-    var new_profile = _.merge(profile, updates);
-    new_profile.patient = new_patient;
-    return api.user.addProfile(userId, new_profile, cb);
+    var newProfile = _.merge(profile, updates, function(original, update){
```
jebeck: This is a place where you've missed just using the factored out `actionUtils.mergeProfileUpdates`, right? Or is there a subtle difference I'm missing?

test/browser/redux/actions/utils.test.js
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+/*eslint-env mocha*/
+
+import _ from 'lodash';
+
+import * as utils from '../../../../lib/redux/actions/utils';
+
+describe('utils', () => {
+  describe('mergeProfileUpdates', () => {
```
jebeck: Thank you for the clear tests with a variety of examples. Makes it much much easier to understand what's going on :)

jebeck:
One tiny TODO assuming I haven't missed a subtlety in the refactoring, otherwise LGTM. Feel free to merge once the TODO is done.

PR #308, opened 2016-07-28
Animas: Cap long running suspended basals after 5 days

gniezen:
We were capping scheduled basals after 5 days, but this logic wasn't applied correctly to suspended or final basals. This PR breaks out the functionality into its own validation function so it can be applied to all basal types.

Pinging @jebeck for review - BA has confirmed that it works with tester pump.

### DIFF
--- a/lib/animas/animasSimulator.js
+++ b/lib/animas/animasSimulator.js
@@ -74,6 +74,21 @@ exports.make = function(config){
     return e;
   }
 
+  function validateDuration(basal) {
+    var fiveDays = (5 * 1440 * sundial.MIN_TO_MSEC);
+    if(basal.isAssigned('duration')) {
+      if(basal.duration > fiveDays) {
+        //flat-rate basal
+        basal.duration = fiveDays;
+        annotate.annotateEvent(basal, 'animas/basal/flat-rate');
+      }
+    } else {
+      basal.with_duration(0);
+      annotate.annotateEvent(basal, 'basal/unknown-duration');
+    }
+    return basal;
+  }
+
   function simpleSimulate(e) {
     ensureTimestamp(e);
     events.push(e);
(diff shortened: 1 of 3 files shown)


gniezen:
Linking Trello card for reference: https://trello.com/c/5Q1QHuWU

gniezen:

lib/animas/animasSimulator.js
```diff
@@ -74,6 +74,21 @@ exports.make = function(config){
     return e;
   }
 
+  function validateDuration(basal) {
```
jebeck: I'd prefer a function name like `truncateDuration` since this can mutate the basal object passed in.

lib/animas/animasSimulator.js
```diff
@@ -74,6 +74,21 @@ exports.make = function(config){
...
+    var fiveDays = (5 * 1440 * sundial.MIN_TO_MSEC);
+    if(basal.isAssigned('duration')) {
+      if(basal.duration > fiveDays) {
+        //flat-rate basal
+        basal.duration = fiveDays;
+        annotate.annotateEvent(basal, 'animas/basal/flat-rate');
```
jebeck: Do you feel like this is the right annotation even for the suspends? Feels different to me... a long-running suspend b/c of a pump vacation isn't really the same thing as a scheduled single rate, other than having the same data limitations.

jebeck:
Overall LGTM, just minor points re: naming and annotation code(s). 🚢

lib/animas/animasSimulator.js
gniezen: Fair point. I wasn't too happy with the name either but couldn't think of something better - `truncateDuration` works for me too.

lib/animas/animasSimulator.js
gniezen: Whether a basal rate is 0 U/hr or 2 U/hr, if it runs for a long time it's a flat rate, right? If you feel strongly that flat rate has a specific connotation in the diabetes world and will be misleading, I'm happy to change it. Keep in mind we'll have to make the corresponding change to tideline too, which in turn means we will be having more discussions around changing the other annotations, e.g., whether we should change the word "capture" to something that better fits Tidepool's voice... ;)

PR #309, opened 2016-07-29
Bug fix: Connecting to the wrong COM port for USB serial devices

gniezen:
See https://trello.com/c/NM8FBoH1 for details.

Pinging @darinkrauss for review as Jana is OoO today. I'm going to merge this into the Animas branch so long so that Nick can test today, but please do bring up any issues identified during review.

PR #310, opened 2016-07-29
Bug fix: Making changes in #309 work for Mac too

gniezen:
See https://trello.com/c/NM8FBoH1 for details.

Pinging @darinkrauss for review. I'm also mergin this into the Animas branch so long so that Nick can test today, but please do bring up any issues identified during review.

### DIFF
--- a/lib/core/device.js
+++ b/lib/core/device.js
@@ -254,38 +254,35 @@ device.detectUsb = function(driverId, cb) {
     return cb(null, devdata);
   };
 
-  if (self._os === 'win') {
-    var chromeVersion = /Chrome\/([0-9]+)/.exec(navigator.userAgent)[1];
-    self.log('Chrome version detected as', chromeVersion);
 
-    chrome.serial.getDevices(function(results) {
-      if (chromeVersion >= 50) {
-        // As of Chrome 50, product ID and vendor ID is returned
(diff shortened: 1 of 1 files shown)


lib/core/device.js
```diff
@@ -254,38 +254,35 @@ device.detectUsb = function(driverId, cb) {
...
-      else{
-        // With Chrome 49 and earlier, there may be
-        // multiple connected devices
-        self.log('Connected device(s): ');
-        for(var i in results) {
-            self.log(results[i].path);
+        } else {
+          return (p.vendorId === driverManifest.vendorId) &&
+          (p.productId === driverManifest.productId) &&
+          p.path.match('/dev/cu.+');
```
darinkrauss: What does /dev/cu.+ typically match?

darinkrauss:
One question, but otherwise LGTM. :shipit:

lib/core/device.js
gniezen: On Mac and Linux, each serial device shows up twice in `/dev`, one for `tty` (incoming) and one for `cu` (outgoing) communication, so you can do full duplex. For example, both `/dev/cu.usbserial` and `/dev/tty.usbserial` will show up. We always want to connect to the `/dev/cu*` one.

PR #313, opened 2016-08-02
Post VCA Additions

krystophv:
Using `redux-form` for validation error handling in in-clinic account creation and adding metrics for many in-clinic experience items.


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VWfUa0A2&#x2F;1-ux-0-dev-3-tst-1-uploader-in-clinic-post-release-feature-cleanup"> &lt;ux:0 dev:3 tst:1&gt; [uploader] [in-clinic] Post-release feature cleanup</a>

krystophv:
ping @jebeck for PR review

jebeck:
Has the current implementation of the form validation been through the final in-browser review process with SK? Just wondering about the language here:
screenshot 2016-08-03 16 31 05

Language seems a bit more formal than we usually have (doesn't blip have like "Hmm, this date doesn't look right"?)

And as a functionality check, we're _not_ trying to validate e-mails at all?

lib/components/ClinicUserEdit.js
```diff
@@ -19,11 +19,37 @@ var React = require('react');
...
+    if(!isValidDate(values.year + '-' + values.month + '-' + zeroPad(values.day))){
+      errors.year = 'Invalid date';
+    }
+  } else {
+    errors.year = 'Invalid date';
+  }
+  return errors;
+};
+
+var isValidDate = function(dateString){
```
jebeck: I slightly prefer non-anonymous functions (e.g., `function isValidDate(dateString) { ... }` or just adding the function name again in addition to assigning to the variable) because they are easier to debug in the console, but not a biggie. If you have strong feelings on a preference, would be interested to hear.

lib/components/ClinicUserEdit.js
```diff
@@ -233,40 +150,78 @@ var ClinicUserEdit = React.createClass({
...
-        <form className={styles.form}>
+        <form className={styles.form} onSubmit={handleSubmit(this.handleNext)}>
           <div className={styles.inputWrap}>
-            <label className={styles.inputLabel} htmlFor="name">Patient Full Name</label>
-            <input className={styles.input} ref="name" name='name' value={this.state.name} onChange={this.handleChange} onBlur={this.handleBlur} placeholder=""/>
+            <label className={styles.inputLabel} htmlFor="name">
+              Patient Full Name
+            </label>
+            <input className={styles.input} {...fullName}/>
+            {fullName.touched && fullName.error && <div className={styles.validationError}>{fullName.error}</div>}
```
jebeck: TBH, I haven't seen this syntax before. What are you doing with the second `&&`??

lib/redux/actions/sync.js
```diff
@@ -89,11 +132,15 @@ export function setForgotPasswordUrl(url) {
   };
 }
 
-export function setPage(page, actionSource = actionSources[actionTypes.SET_PAGE]) {
+export function setPage(page, actionSource = actionSources[actionTypes.SET_PAGE], metric) {
+  let meta = {source: actionSource};
```
jebeck: I think you could actually use `const` here (and in the AirBnb ESLint config world it would be preferable) - because the `_.assign` doesn't reassign anything _to_ metric, just assigns new properties within it, which is allowed by `const`.

lib/redux/constants/metrics.js
```diff
@@ -20,6 +20,23 @@ export const CLICK_GO_TO_BLIP = 'Clicked See Data in Blip';
...
+export const CLINIC_ADD_MRN = 'VCA Add MRN Saved';
+export const CLINIC_ADD_EMAIL = 'VCA Add Patient Email Saved';
+export const CLINIC_ADD_NEW_PATIENT = 'VCA Add New Patient Saved';
+export const CLINIC_ADD_INVALID_DATE = 'VCA Add Invalid Date';
+export const CLINIC_CHANGE_PERSON = 'VCA Change Person';
+export const CLINIC_EDIT_INFO = 'VCA Edit Info';
+export const CLINIC_DEVICE_STORED = 'VCA Device Stored';
+export const CLINIC_DEVICES_DONE = 'VCA Devices Done';
+export const CLINIC_TIMEZONE_SELECT = 'VCA Timezone Selected';
+export const CLINIC_CHANGE_DEVICES = 'VCA Change Devices';
```
jebeck: Some of these added metrics are not reflected in any tests and should be:
- [ ] CLINIC_SEARCH_SELECTED
- [ ] CLINIC_ADD_CANCEL
- [ ] CLINIC_CHANGE_PERSON
- [ ] CLINIC_EDIT_INFO
- [ ] CLINIC_DEVICES_DONE
- [ ] CLINIC_TIMEZONE_SELECT
- [ ] CLINIC_CHANGE_DEVICES

If there's some reason why (some or all of) these are particularly hard to instrument, let's discuss.

[Background: we've learned the hard way before that metrics are easy to lose if there aren't tests around validating their continued presence.]

lib/components/ClinicUserEdit.js
```diff
@@ -49,89 +75,19 @@ var ClinicUserEdit = React.createClass({
...
-    e.preventDefault();
-    this.validateForm();
-    if(this.state.errorMessage){
-      return;
-    }
-    var name = this.refs.name.value;
-    var dateString = this.refs.year.value+'-'+this.refs.month.value+'-'+this.zeroPad(this.refs.day.value);
-    var email = this.refs.email.value;
-    var mrn = this.refs.mrn.value;
+  handleNext: function(values, dispatch) {
```
jebeck: 2nd arg isn't used...

lib/components/ClinicUserEdit.js
```diff
@@ -157,72 +113,33 @@ var ClinicUserEdit = React.createClass({
...
       </div>
     );
   },
 
+  renderDateError: function(){
+    const {fields: {fullName, month, day, year}} = this.props;
+    return ((month.touched || month.value) && !month.active) &&
+      ((day.touched || day.value) && !day.active) &&
+      ((year.touched || year.value) && !year.active) && year.error &&
+      <div className={styles.validationError}>{year.error}</div>;
```
jebeck: Similar question about last bit of logic (the final `&&`), and this is a _huge_ logic statement. Can you break it up for transparency/readability?

lib/components/ClinicUserEdit.js
```diff
@@ -157,72 +113,33 @@ var ClinicUserEdit = React.createClass({
...
         <span>
-          {this.props.errorMessage}{this.state.errorMessage}<i className={styles.iconClose} onClick={this.props.dismissCreateCustodialAccountError}></i>
+          {this.props.errorMessage}<i className={styles.iconClose} onClick={this.props.dismissCreateCustodialAccountError}></i>
         </span>
       </div>
     );
   },
 
+  renderDateError: function(){
+    const {fields: {fullName, month, day, year}} = this.props;
```
jebeck: Not using `fullName` here, no need to destructure it out. (My AirBnb trained eye sees such things easily now 😜)

lib/components/ClinicUserSelect.js
```diff
@@ -41,6 +42,10 @@ var ClinicUserSelect = React.createClass({
     }
   },
 
+  handleOnChange: function(userId) {
+    this.props.setTargetUser(userId, {eventName: metrics.CLINIC_SEARCH_SELECTED});
+  },
```
jebeck: So I guess I see the issue (raised in an earlier comment) about the lack of tests for this. The root cause of the problem is how we don't yet have unit tests around React components in the uploader (only around the redux implementation). The tests will be pretty easy to add once we're set up for them, but setting up for React tests _and_ writing them may significantly expand the scope of this task.

Curious as to your thoughts on this @jh-bate, balancing "scope creep" against realizing that if we merge as-is we're not including tests for the addition of some metrics which we've tried a bit harder in the uploader to do (thus far).

lib/redux/actions/async.js
```diff
@@ -477,9 +483,22 @@ export function clickDeviceSelectionDone() {
           } else {
             let newProfile = actionUtils.mergeProfileUpdates(allUsers[uploadTargetUser], updates);
             dispatch(syncActions.updateProfileSuccess(newProfile, uploadTargetUser));
+            if (isClinicAccount) {
+              _.forEach(targetDevices[uploadTargetUser], function(device){
```
jebeck: This dispatches a metric for each target device even if it isn't _newly_ added, if I'm reading this right...? That seems like it will overreport the metric. I see the complication here with how the action works, so maybe the functionality/limitations should be discussed with BA?

lib/components/ClinicUserSelect.js
jebeck: Chatted with @jh-bate a bit about this, and the two of us are agreed that perhaps the next uploader task that touches any React component should include setting up React component testing and writing some tests for that component (exactly what these tests look like to be decided in future discussion) as part of the task. Then as we touch components in the uploader we can try to gradually add tests. Thoughts @krystophv? Sound workable to you?

lib/components/ClinicUserEdit.js
krystophv: Logical && will return the rightmost value if all values are truthy. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_AND_() (line 5). 
I think you're right about `touched`, I'll revise it for clarity.

lib/components/ClinicUserEdit.js
krystophv: I don't have strong feelings, non-anonymous is fine (and probably preferable actually) to me.

lib/components/ClinicUserSelect.js
krystophv: Sounds good to me.

lib/components/ClinicUserEdit.js
jebeck: Ah, interesting, OK. Can you put the <div> in parens though? It's a bit hard to scan what that final value _is_. (Also it used to be that JSX _had_ to be in parens, way way back in the early days of React, so it's an old habit to me.)

lib/redux/actions/async.js
krystophv: Discussed and he's aware and ok with it.

krystophv:
@jebeck - design review and PR updates complete

jebeck:
LGTM, 👍

PR #315, opened 2016-08-05
updates to avoid peer deps warnings with npm 3.x

jebeck:
@gniezen perhaps try to update npm then reinstall your node modules from this branch, confirm no warnings, and merge? No changes that need testing - only change here is in devDeps.


gniezen:
Just the usual deprecation warnings from sub-dependencies like minimatch which I'm guessing we can't do anything about? Also, just double-checking whether you intended for the `express` version to stay the same, with `sinon-chai` being a new added dependency?

jebeck:
Yeah, I think the `express` just got corrected alphabetically in the list when I ran `npm install --save-dev` on `sinon-chai` and that's why it's in the diff ;)

Also correct that unfortunately we can't do anything about the warnings from sub-dependencies...

gniezen:
Merging, and not building a new release as testing is not required.

PR #316, opened 2016-08-05
VCA update email

krystophv:
Makes appropriate API calls to update user account information along with profile information and re-send signup confirmation email as necessary when a clinical account modifies a patient's email address.

### DIFF
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -167,6 +167,18 @@ api.user.updateProfile = function(userId, updates, cb){
       return cb(err);
     }
     var newProfile = actionUtils.mergeProfileUpdates(profile, updates);
+    var emails = _.get(updates, 'emails');
+    if (_.isArray(emails) && emails.length === 1){
+      return async.series([
+        function(callback){return tidepool.updateCustodialUser({username: emails[0], emails: emails}, userId, callback);},
+        function(callback){return api.user.addProfile(userId, newProfile, callback);},
+        function(callback){return tidepool.signupStart(userId, callback);}
+      ], function(err, results){
+        if (err){ cb(err); }
+        cb(null, results[1]);
+      }
+      );
+    }
     return api.user.addProfile(userId, newProfile, cb);
   });
 };
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -169,13 +169,22 @@ api.user.updateProfile = function(userId, updates, cb){
     var currentEmail = _.get(profile, 'emails[0]');
     var newProfile = actionUtils.mergeProfileUpdates(profile, updates);
     var emails = _.get(updates, 'emails');
+    // check to see if we have a single email address that also needs to be updated
     if (_.isArray(emails) && emails.length === 1 && emails[0] !== currentEmail){
       return async.series([
-        function(callback){return tidepool.updateCustodialUser({username: emails[0], emails: emails}, userId, callback);},
-        function(callback){return tidepool.addOrUpdateProfile(userId, newProfile, callback);},
-        function(callback){return tidepool.signupStart(userId, callback);}
+        function(callback){
+          return tidepool.updateCustodialUser({username: emails[0], emails: emails}, userId, callback);
+        },
+        function(callback){
+          return tidepool.addOrUpdateProfile(userId, newProfile, callback);
+        },
+        function(callback){
+          return tidepool.signupStart(userId, callback);
+        }
       ], function(err, results){
-        if (err){ cb(err); }
+        if (err) {
+          cb(err);
+        }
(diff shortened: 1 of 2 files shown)


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;XfHSKoKc&#x2F;15-ux-0-dev-2-tst-1-uploader-ui-add-an-email-address-to-a-custodial-account-after-it-s-been-created"> &lt;ux:0 dev:2 tst:1&gt; [uploader] [UI] Add an email address to a custodial account after it&#39;s been created</a>

krystophv:
ping @jebeck for review

lib/core/api.js
```diff
@@ -166,8 +166,21 @@ api.user.updateProfile = function(userId, updates, cb){
...
     var newProfile = actionUtils.mergeProfileUpdates(profile, updates);
-    return api.user.addProfile(userId, newProfile, cb);
+    var emails = _.get(updates, 'emails');
+    if (_.isArray(emails) && emails.length === 1 && emails[0] !== currentEmail){
+      return async.series([
+        function(callback){return tidepool.updateCustodialUser({username: emails[0], emails: emails}, userId, callback);},
+        function(callback){return tidepool.addOrUpdateProfile(userId, newProfile, callback);},
+        function(callback){return tidepool.signupStart(userId, callback);}
+      ], function(err, results){
+        if (err){ cb(err); }
```
jh-bate: looking at the rest of the code this style of single line statements doesn't match and should be updated to reflect the existing code I feel

lib/core/api.js
```diff
@@ -166,8 +166,21 @@ api.user.updateProfile = function(userId, updates, cb){
     if(err){
       return cb(err);
     }
+    var currentEmail = _.get(profile, 'emails[0]');
     var newProfile = actionUtils.mergeProfileUpdates(profile, updates);
-    return api.user.addProfile(userId, newProfile, cb);
+    var emails = _.get(updates, 'emails');
+    if (_.isArray(emails) && emails.length === 1 && emails[0] !== currentEmail){
+      return async.series([
+        function(callback){return tidepool.updateCustodialUser({username: emails[0], emails: emails}, userId, callback);},
```
jh-bate: as per note below - single line style doesn't seem to match

lib/core/api.js
```diff
@@ -166,8 +166,21 @@ api.user.updateProfile = function(userId, updates, cb){
     if(err){
       return cb(err);
     }
+    var currentEmail = _.get(profile, 'emails[0]');
```
jh-bate: is this statement all good if there is no email?

lib/core/api.js
```diff
@@ -166,8 +166,21 @@ api.user.updateProfile = function(userId, updates, cb){
...
     var newProfile = actionUtils.mergeProfileUpdates(profile, updates);
-    return api.user.addProfile(userId, newProfile, cb);
+    var emails = _.get(updates, 'emails');
+    if (_.isArray(emails) && emails.length === 1 && emails[0] !== currentEmail){
```
jh-bate: do you think its worth pulling this out into its own function? just thinking that its kinda hard to tell what is happening here at a glance and maybe being in own function it would be a little clearer,  or having a brief comment as to whats going on. Thoughts?

lib/core/api.js
```diff
@@ -166,8 +166,21 @@ api.user.updateProfile = function(userId, updates, cb){
...
+      return async.series([
+        function(callback){return tidepool.updateCustodialUser({username: emails[0], emails: emails}, userId, callback);},
+        function(callback){return tidepool.addOrUpdateProfile(userId, newProfile, callback);},
+        function(callback){return tidepool.signupStart(userId, callback);}
+      ], function(err, results){
+        if (err){ cb(err); }
+        cb(null, results[1]);
```
jh-bate: what does `results[1]` represent?

lib/core/api.js
krystophv: Yeah, `_.get` is always safe, apparently. I had to check it's behavior in `console` to be sure myself before feeling good about this line.

lib/core/api.js
krystophv: With `async.series`, the results are indexed matching the array of input functions, so this represents the return from `tidepool.addOrUpdateProfile`, which is the return for the case that there aren't emails or the email doesn't need to be updated.

lib/core/api.js
krystophv: If we were using it in more than one place, I'd say it might be a candidate for moving to a `util` function, but as is, a comment might be more appropriate.

jh-bate:
LGTM

PR #318, opened 2016-08-16
change npm run karma-watch command to npm run test-watch

jh-bate:
@jebeck


jebeck:
FYI @gniezen @krystophv - Jamie wanted to change the command name, and I agree `test-watch` is better. Will be standard across all front-end repos as soon as all PRs are merged.

PR #319, opened 2016-08-17
Debugging Animas missing records 

gniezen:
When there is a missing record, we should print out the record before and just after the missing record so we have a better indication of where it went missing.

See https://trello.com/c/skctayOf for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ $ npm run node-tests
 To run just the UI tests in PhantomJS with webpack & Karma watching all files for changes and both rebundling the app and re-running the tests on every change, use:
 
 ```bash
-$ npm run karma-watch
+$ npm run test-watch
 ```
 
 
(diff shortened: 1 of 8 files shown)


gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;skctayOf&#x2F;21-uploader-bug-animas-upload-fails-due-to-missing-records"> [uploader][bug] Animas upload fails due to missing records</a>

gniezen:
@jebeck I'm pinging you for code review as I remember you running into the "missing record" issue once. This PR should make it a bit easier to debug in future. (This was prompted by the user report that we couldn't investigate further as there were issues with the pump).

jebeck:
LGTM 👍

gniezen:
Merging, but not building a new release as this is a minor change and not user-facing

PR #322, opened 2016-08-18
add a before_install to get 3.x version of npm

jebeck:
Will self-merge if build passes. Since we've updated Bob to 3.x npm we need to be building with that on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ language: node_js
 node_js:
   - "0.12.7"
   - "stable"
+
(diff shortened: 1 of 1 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;0QAfSTOP&#x2F;26-viz-tideline-blip-uploader-update-travis-to-use-npm-3-x-now-that-we-ve-updated-bob-to-use-that"> [viz] [tideline] [blip] [uploader] update Travis to use npm 3.x now that we&#39;ve updated Bob to use that</a>

PR #325, opened 2016-09-07
Initial work on Medtronic driver 

gniezen:
This PR creates a new Medtronic driver and adds support for:

- manual boluses (https://trello.com/c/OAHqN8Ng)
- calculator boluses (https://trello.com/c/4Gn33urX)

This PR also refactors our USB HID library (`hidDevice.js`). The USB HID library, originally written by IOET while developing the initial version of the Bayer Contour Next driver, was modelled on our USB serial library. However, this introduced unnecessary complexity (buffers, interval timers etc.) that has now been removed. This also necessitated changes to the Bayer Contour Next driver as to not use an interval timer, which can cause unpredictable results.


gniezen:
@jebeck Ready for review.

lib/drivers/medtronicDriver.js
```diff
@@ -0,0 +1,746 @@
...
+          }
+      });
+    },
+
+    processData: function (progress, data, cb) {
+      debug('in processData');
+      data.settings.deviceId = data.settings.modelNumber + '-' + serial;
+      cfg.builder.setDefaults({ deviceId: data.settings.deviceId});
+      // TODO: get all time changes and most recent event
+      cfg.tzoUtil = new TZOUtil(cfg.timezone, new Date().toISOString(), []);
```
jebeck: Just as a note for the future, I _think_ you'll need two instances of `tzoUtil` because CGM data stream and pump data stream need to be bootstrapped separately. Just mentioning now as a seed to plant for code organization/architecture, perhaps.

lib/medtronic/cli/blob_loader.js
```diff
@@ -0,0 +1,107 @@
...
+ */
+function loadFile(filePath, tz, userid) {
+  fs.readFile(filePath, 'utf8', function(error, data) {
+    console.log(intro, 'Reading', filePath);
+    if (error) {
+      console.log(intro, 'Error reading Medtronic blob', error);
+      return;
+    }
+
+    var json = JSON.parse(data);
```
jebeck: Ah, so is the blob the parsed-into-JavaScript-objects raw records?

lib/medtronic/medtronicSimulator.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  function simpleSimulate(e) {
+    ensureTimestamp(e);
+    events.push(e);
+  }
+
+  return {
+    bolus: function(event) {
+      if(currWizard && currWizard.jsDate.getTime() === event.jsDate.getTime()) {
```
jebeck: I just `git grep`ed and I believe these are the only instances of use of `getTime()`. I dislike the method due to the imprecision of its name (almost makes it sound like it's only getting the clock hours or something...). I would be in favor of `jsDate.valueOf()` instead - we use this a bit more and it returns the same value (a hammertime).

lib/medtronic/medtronicSimulator.js
```diff
@@ -0,0 +1,121 @@
...
+        currWizard.bolus = event;
+        currWizard = currWizard.done();
+        simpleSimulate(currWizard);
+      }
+      simpleSimulate(event);
+      setCurrBolus(event);
+    },
+    wizard: function(event) {
+      if(currBolus && currBolus.jsDate.getTime() === event.jsDate.getTime()) {
```
jebeck: Ditto re: `getTime()`.

lib/medtronic/medtronicSimulator.js
```diff
@@ -0,0 +1,121 @@
...
+        event = event.done();
+        simpleSimulate(event);
+      }
+      setCurrWizard(event);
+    },
+    getEvents: function() {
+      function filterOutInvalidData() {
+        return _.filter(events, function(event) {
+
+          //filter out zero boluses
```
jebeck: Just skipping the filtering out for now? Why commented out?

lib/medtronic/processData.js
```diff
@@ -0,0 +1,378 @@
...
+      .with_bgInput(bgInput)
+      .with_carbInput(carbInput)
+      .with_insulinOnBoard(iob)
+      .with_insulinCarbRatio(carbRatio)
+      .with_insulinSensitivity(isf)
+      .with_bgTarget({
+        low: bgTarget.low,
+        high: bgTarget.high
+      })
+      .with_units('mg/dL'); //TODO: see if we can find units
```
jebeck: You read my mind re: TODO ;)

lib/medtronic/processData.js
```diff
@@ -0,0 +1,378 @@
...
+
+            if(programmed !== amount) {
+              // dual bolus was cancelled
+              var actualDuration = Math.round((amount / (programmed * 1.0)) * duration);
+              bolus = bolus.with_expectedExtended(programmed)
+                        .with_expectedDuration(duration)
+                        .with_duration(actualDuration);
+            }
+
+            i+=1; // advance to next bolus for normal portion
```
jebeck: So as mentioned before, I think this sequencing (or it may be the sequencing in relation to wizard events, not sure if affects both or just one) can be interrupted. How do we know/check here that after advancing we really do have a bolus record, or the right bolus record? Perhaps the situation is actually way better than I thought! Do you get the `bolusRecords` all in one batch, so they are all guaranteed to be boluses? The problem we had in CareLink parsing was that we were dealing with all types at once and sometimes an alarm would intervene in the sequence (by log index).

lib/medtronic/processData.js
```diff
@@ -0,0 +1,378 @@
...
+
+    if (larger) {
+      var duration = record.head[7] * 30 * sundial.MIN_TO_MSEC;
+      var amount = struct.extractBEShort(record.head,3)/40.0;
+      var programmed = struct.extractBEShort(record.head,1)/40.0;
+      var iob = struct.extractBEShort(record.head,5)/40.0;
+      if (duration > 0) {
+        //TODO: put iob in payload?
+
+        if((record.date[2] & 0xC0) == 0x80) {
```
jebeck: Seems like there should be some constants factored out here. What is `0xC0`? What is `0x80`? Not easy to read what the logic is as it stands...

lib/medtronic/processData.js
```diff
@@ -0,0 +1,378 @@
...
+
+  if(debug && isBrowser) {
+    savePages(pages);
+  }
+
+  for(var i = 0; i < pages.length; i++) {
+    var page = pages[i];
+    console.log('Page',i,common.bytes2hex(page));
+    var index = 0;
+    while(index < 1024) {
```
jebeck: Ditto re: constant, although this one's not that opaque.

lib/medtronic/processData.js
```diff
@@ -0,0 +1,378 @@
...
+      var record = {};
+      record.head = struct.extractBytes(page,index,2);
+      if(record.head[0] === 0) {
+        console.log('End of page.');
+        break;
+      }
+      record.type = getType(record.head[0]);
+      if(record.type.value === 0x55) {
```
jebeck: Ditto again if possible to name the constant value.

lib/medtronic/cli/blob_loader.js
gniezen: Yes. At the moment the "blob" is just an array of of arrays where I'm storing each page of binary data separately, but I'll be adding other metadata (e.g. model number) to the file in future, which is why I'm storing it as JSON.

lib/medtronic/medtronicSimulator.js
gniezen: I was just considering using `.time` instead, but it's probably safer to compare hammertimes than strings.

lib/medtronic/processData.js
gniezen: I don't get the bolus records all in one batch, but in this part I've filtered to just look at bolus records. As they're reverse chronological, I'm expecting to always see a normal portion bolus record after the dual portion bolus record.

lib/medtronic/processData.js
gniezen: Ah, yes - the constant was already there: `RECORD_TYPES.UNKNOWN_55.value`. I've also added a TODO to support this record type properly and not just throw an error if found. (The  record body length is variable, which makes it a bit tricky)

lib/medtronic/medtronicSimulator.js
gniezen: Wasn't sure if it's needed for Medtronic boluses - have added it back in as it doesn't hurt. Should DRY out this function sometime for all drivers - Insulet, Animas and Tandem.

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;OAHqN8Ng&#x2F;14-0-ux-5-dev-0-tst-uploader-medtronic-manual-boluses"> &lt;0:ux 5:dev 0:tst&gt;[uploader] Medtronic manual boluses</a>

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4Gn33urX&#x2F;30-uploader-medtronic-calculator-data"> [uploader] Medtronic calculator data</a>

jebeck:
LGTM 👍

lib/drivers/medtronicDriver.js
gniezen: I just came across this comment again while cleaning up my inbox, and it turns out that's exactly what I had to do. @jebeck , you're a smart 🍪!  😄

jebeck: 😊

PR #326, opened 2016-09-14
Support for different Medtronic models and fix for calculator boluses

gniezen:
The 523 model (used for development) has some differences compared to other pump models:
- Some records are larger
- Some units from the pump need to be divided by 40.0 instead of 10.0 (`strokesPerUnit`) to get the correct value

This PR also fixes some issues with calculator/wizard boluses and adds support for the other pump models.

### DIFF
--- a/lib/medtronic/processData.js
+++ b/lib/medtronic/processData.js
@@ -104,10 +104,14 @@ var RECORD_TYPES = {
 };
 
 var BITMASKS = {
-  LEFT_TWO: 0xc0,   // b11000000
-  RIGHT_SIX: 0x3f,  // b00111111
-  RIGHT_FIVE: 0x1f, // b00011111
-  RIGHT_SEVEN: 0x7f // b01111111
+  LEFT_TWO: 0xc0,     // b11000000
+  RIGHT_SIX: 0x3f,    // b00111111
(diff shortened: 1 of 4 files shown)


gniezen:
Pinging @jebeck for review.

gniezen:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4Gn33urX&#x2F;30-uploader-medtronic-calculator-data"> [uploader] Medtronic calculator data</a>

lib/medtronic/processData.js
```diff
@@ -27,7 +27,8 @@ var savedFileEntry, fileDisplayPath;
 var debug = (typeof __DEBUG__ === 'undefined') ? false : __DEBUG__;
 var isBrowser = typeof window !== 'undefined';
 
-var larger = true; //TODO: this should be based on model number
+var larger = false; // model 523 has larger records
```
jebeck: 523 has larger records than which other model(s)? Just for a bit more clarity here.

gniezen: This was bothering me too. I've actually refactored the code in commit 358057cdfb9f0322c on `gniezen/medtronic-basal` to be clearer and not use global variables, i.e., it's stored in `data.settings` instead.

PR #328, opened 2016-09-16
Medtronic basals

gniezen:
This will add support for:
- scheduled basals
- temp basals
- suppressed info
- suspend/resume records

Since basal schedules at a specific moment in time are required to build suppressed info and temp basals, this PR also adds support for getting the current basal schedules as well as basal schedule history.

It also refactors some common functionality out of the Animas and CareLink drivers. 

For more details, see https://trello.com/c/zlAnS0iw


gniezen:
[[uploader] Medronic basals](https://trello.com/c/zlAnS0iw/13-uploader-medronic-basals)

gniezen:
Pinging @jebeck for review.

jebeck:
My brain is feeling a bit mushy this evening, so I skipped over the simulator and simulator tests stuff for now - will come back to that later (perhaps tomorrow?) since you said there's no rush, and I know those are the tricky bits.

lib/medtronic/medtronicSimulator.js
```diff
@@ -66,6 +77,111 @@ exports.make = function(config){
...
+          var startTime = common.computeMillisInCurrentDay(currBasal);
+          var endTime = startTime + currBasal.duration;
+
+          for(var i in currSchedule) {
+            if((startTime < currSchedule[i].start) && (endTime > currSchedule[i].start)) {
+              // there was a schedule change _during_ the temp basal
+              var adjustedDuration = currSchedule[i].start - startTime;
+              var oldDuration = currBasal.duration;
+              currBasal.with_duration(adjustedDuration)
+                       .with_expectedDuration(oldDuration);
```
jebeck: IMO, this is not a case for the use of `expectedDuration`, and I don't believe we've used it as such for schedule changes during temps in the old CareLink parser. I also just checked in my Tandem data because I conveniently had a schedule-crossing basal running last night, and the first segment of the temp-that-crosses-boundaries does not have an `expectedDuration` there. (Now, perhaps we don't have that info for Tandem, can't remember, but still I think `expectedDuration` does not exactly belong in this case.) However, it does beg the question of how to represent it when the temp basal as a whole is cut short. Say I program a temp basal lasting 6 hours. It crosses a schedule boundary after 1.25 hours and gets split into two segments, then I cancel it four hours in. I don't think we have a way to represent the original "collective" `expectedDuration` in that case with our current data model...

test/node/medtronic/testMedtronicSimulator.js
```diff
@@ -0,0 +1,504 @@
...
+        simulator.basal(basal1);
+        simulator.basal(tempBasal);
+        simulator.basal(basal2);
+        expect(simulator.getEvents()).deep.equals([
+          basal1.done(),
+          expectedTempBasal
+        ]);
+      });
+
+      it('checks for schedule change', function() {
```
jebeck: Add another test case here of a temp basal crossing midnight, possibly with and without the rate actually *changing* at midnight.

lib/medtronic/medtronicSimulator.js
```diff
@@ -66,6 +77,111 @@ exports.make = function(config){
...
+                deliveryType: 'scheduled',
+                rate: currSchedule[i].rate
+              };
+
+              var newTemp = config.builder.makeTempBasal()
+                .with_time(newJsDate.toISOString())
+                .with_duration(oldDuration - adjustedDuration)
+                .with_deviceTime(sundial.formatDeviceTime(newJsDate))
+                .set('index', currBasal.index)
+                .set('suppressed', newSuppressed); // TODO: build array of suppressed objects
```
jebeck: Not sure I understand your comment here. Are we changing the nested suppressed to an array of them?

lib/medtronic/medtronicSimulator.js
```diff
@@ -66,6 +77,111 @@ exports.make = function(config){
...
+            var suppressed = {
+              type: 'basal',
+              deliveryType: 'scheduled',
+              rate: currBasal.rate
+            };
+            event.suppressed = suppressed;
+          }
+
+          if (!event.rate && event.duration === 0) {
+            // temp basal was cancelled
```
jebeck: Confused about the conditions in the `if`. Is the `duration` of 0 the indicator that it was canceled? And was it therefore cancelled immediately, before it had any effect? If so, perhaps we just toss it?

lib/commonFunctions.js
```diff
@@ -19,6 +19,42 @@ var sundial = require('sundial');
...
 var annotate = require('./eventAnnotations');
 
+/**
+ * Computes the number of milliseconds after midnight on the date specified.
+ *
+ * @param dateTime DateTime object to figure out millis from
+ * @returns {number} number of millis in current day
+ */
+exports.computeMillisInCurrentDay = function(e){
+  var msFromMidnight = sundial.getMsFromMidnight(e.time, e.timezoneOffset);
```
gniezen: So I moved this function from the CareLink driver, and while writing the tests just now I started wondering if we shouldn't be using `clockDriftOffset` instead of `conversionOffset` in this function? Or am I misunderstanding how it works?

I've added some tests to `testCommonFunctions.js` - let me know if they're sufficient or not.

lib/drivers/medtronicDriver.js
```diff
@@ -346,16 +392,33 @@ module.exports = function (config) {
...
+          if (message.messageLength >= 33) {
+            // This is a Medtronic message
             if (message.bytes.slice(33)[0] === ASCII_CONTROL.ACK ) {
               clearTimeout(abortTimer);
               return callback(true);
             }
+
+            if(fetchingHistory) {
+              if(payloadLength && payloadLength < 256) {
+                message.resultsInReply = message.bytes.slice(33);
```
gniezen: Good idea, done.

lib/drivers/medtronicDriver.js
```diff
@@ -636,16 +741,26 @@ module.exports = function (config) {
       debug('in processData');
       data.settings.deviceId = data.settings.modelNumber + '-' + serial;
       cfg.builder.setDefaults({ deviceId: data.settings.deviceId});
-      // TODO: get all time changes and most recent event
-      cfg.tzoUtil = new TZOUtil(cfg.timezone, new Date().toISOString(), []);
       progress(0);
-      proc.init(cfg);
+
       proc.processPages(data, function (err, records) {
+        var dateRecords =  _.filter(records, function(event) {return event.jsDate;}); // filter out objects without dates
```
gniezen: `RESULT_DAILY_TOTAL`, `UNKNOWN_6C`, `MODEL_522_RESULTS_TOTALS` and `UNKNOWN_6E` have 2-byte date fields only. `UNABSORBED_INSULIN` doesn't have a date field at all.

None of these are in our data model, so can be ignored for now. `RESULT_DAILY_TOTAL` doesn't necessarily occur at midnight so can't be used as a new-day event.

lib/medtronic/processData.js
```diff
@@ -417,23 +425,231 @@ function buildBGRecords(records) {
...
+
+      var suspendResume = cfg.builder.makeDeviceEventSuspendResume()
+        .with_deviceTime(sundial.formatDeviceTime(suspendEntry.jsDate))
+        .with_reason({suspended: 'manual', resumed: 'manual'}) // TODO: get reason
+        .set('index', suspendEntry.index);
+      cfg.tzoUtil.fillInUTCInfo(suspendResume, suspendEntry.jsDate);
+
+      var resumeEntry = suspendResumeRecords[i+1];
+      if(resumeEntry && resumeEntry.type.value === RECORD_TYPES.PUMP_RESUME.value) {
+        i += 1; // increment counter, as we have a matching resume
```
gniezen: When you suspend before uploading, i.e., it's the last record? It's not necessary to suspend for uploading, but possible. I have not seen two suspends in succession, no.

lib/medtronic/processData.js
```diff
@@ -417,23 +425,231 @@ function buildBGRecords(records) {
...
+                                 RECORD_TYPES.SELECT_BASAL_PROFILE], records).reverse();
+  var postrecords = [];
+  var scheduleChanges = [];
+  var activeSchedule, prevRecord;
+
+  // when we record schedule changes, we need to pre-fill with the
+  // most recent schedules, and remember that they only change for the
+  // *previous* schedules
+  var scheduleChange = _.clone(cfg.settings.basalSchedules);
+  scheduleChanges.push(scheduleChange,scheduleChange);
```
gniezen: To make line 522 work. Basically, you have to look two schedule changes back to see what changed, that is, compare the previous schedule change (A) against the one before it (B). For the most recent schedule change, nothing changed, so A=B. I tried to explain it in the comment, but realize it may not be clear enough. If you agree with the approach, I can change the comment to try and explain it better.

lib/medtronic/medtronicSimulator.js
```diff
@@ -66,6 +77,111 @@ exports.make = function(config){
...
+    basal: function(event) {
+      ensureTimestamp(event);
+
+      if (currBasal != null) {
+
+        if(currBasal.deliveryType === 'temp' && event.deliveryType === 'temp' && event.duration !== 0) {
+          // temp basal was updated
+          currBasal.with_expectedDuration(currBasal.duration)
+                   .with_duration(Date.parse(event.time) - Date.parse(currBasal.time));
+          event.suppressed = _.clone(currBasal.suppressed);
```
gniezen: On the pump you can actually edit a running temp basal's duration. As such, I'm stopping the old temp basal, starting a new one with the new duration, and copying over the suppressed info. Should this rather result in the entire edited temp basal being added to the suppressed info? It's these kind of things I'm a bit unclear about since the docs don't exist yet.

lib/medtronic/medtronicSimulator.js
gniezen: Ah, so we only use `expectedDuration` for cancelled temp basals, not for schedule changes. For now, I'll add it to the payload. (We do the same for Tandem IIRC).

lib/medtronic/medtronicSimulator.js
gniezen: Another example of why I need that suppressed info doc :) I wasn't sure how we build up suppressed info (array or nested) so this comment just acts as a reminder for me that I should do this as soon as the doc is available.

lib/medtronic/medtronicSimulator.js
gniezen: When a temp basal is canceled, the pump sends a temp basal record with duration 0. We use the time this record was sent to calculate the actual duration.

lib/medtronic/medtronicSimulator.js
gniezen: Tested it on the pump again and the "edit" functionality only works for temp basals where the rate is set, not for percent temp basals. If you press "Set/edit" while a rate temp basal is running, it doesn't do anything.

test/node/medtronic/testMedtronicSimulator.js
gniezen: Done.

lib/medtronic/medtronicSimulator.js
```diff
@@ -66,6 +77,111 @@ exports.make = function(config){
...
+        }
+
+        if(currBasal.deliveryType === 'temp' && currPumpSettings) {
+          var currSchedule = currPumpSettings.basalSchedules[currPumpSettings.activeSchedule];
+          var startTime = common.computeMillisInCurrentDay(currBasal);
+          var endTime = startTime + currBasal.duration;
+
+          for(var i in currSchedule) {
+            if((startTime < currSchedule[i].start) && (endTime > currSchedule[i].start)) {
```
gniezen: OK, I've added some logic to handle temp basals crossing midnight with and without schedule changes. The tests pass, but it's not quite as clean as I would like. If there's a better way to do this, please point me in the right direction :)

lib/commonFunctions.js
jebeck: No, we definitely want to avoid including the `clockDriftOffset` because including it actually effects a *decrease* in our ability to match up to the basal schedules. If you think about anything particular instance where a user adjusted for clock drift, the device times prior to that *included* the drift when the pump was ticking through the basal schedule. If we apply the correction for drift retroactively, we do produce technically more accurate timestamps, but they match up with the schedules less well because we're being smarter than the pump was, if that makes sense.

LMK if you want to talk this through. It's a tricky bit, and I'm probably not doing it justice here.

Tests aren't quite sufficient, but I'll comment on specific inline.

lib/medtronic/medtronicSimulator.js
jebeck: OK, this was news to me and may require some thought re: suppressed. Not something we've actually considered before.

In your second comment you seem to contradict yourself. In the second second, do you mean 'If you press edit while a *percent* basal is running'...?

lib/medtronic/medtronicSimulator.js
jebeck: Perhaps a comment explaining this?

lib/medtronic/processData.js
jebeck: I spent quite a while looking at this this morning. I remember this stuff being tricky, but I still can't say I understand the code. :/

I think tests would help, but I'm sure the test fixtures would be incredibly laborious to create. Perhaps we need to just QA the hell of out this stuff?

test/node/testCommonFunctions.js
```diff
@@ -111,4 +111,38 @@ describe('commonFunctions.js', function(){
...
+    });
+
+    it('rounds to nearest 15 minutes for clock skew', function(){
+      basal.with_time('2015-11-05T17:05:00.000Z')
+           .with_conversionOffset(420000);
+      console.log(basal);
+      expect(common.computeMillisInCurrentDay(basal)).to.equal(61200000);
+    });
+
+    it('checks that rounding does not result in more than 24 hours', function(){
```
jebeck: Can you explain how this tests that it doesn't result in > 24 hours? How would you get this test to fail by changing the logic of `computeMillisInCurrentDay`? I'm not seeing it, but it admittedly does my head in a bit to try to think through this stuff with your positive offset from UTC since I'm used to thinking about negative offsets.

In any case, this is also not really testing this at the right point. I really intended for tests to be added to sundial...but now looking at the `sundial.getMsFromMidnight` I'm fairly convinced that it can't screw up. The trickier case where I had to be careful about this is when a milliseconds-from-midnight function is taking a timezone as input, not a timezone offset. (Because the timezone can be associated with more than one offset in a single day, so you have to be careful about how you're finding the previous midnight that you're subtracting...)

(The other thing that's getting in the way of readability here, IMO, is using the `conversionOffset` in more than one test set up... I think it'd be better to isolate what's being tested a bit more and if needed have one "kitchen sink" test that throws all the weird things together.)

lib/commonFunctions.js
jebeck: Just thought of another thing to point you to that may help explain...

Well first, remember that the "equation" for timestamps is this (from the BtUTC doc):
```
deviceTime = time + timezoneOffset + conversionOffset
```

Notice that `clockDriftOffset` is not part of it. We track it as a remainder from the rounding we do with `timezoneOffset`s, but that's it. And if you compare [TZOUtil line 67](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/TimezoneOffsetUtil.js#L67) with [TZOUtil line 70](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/TimezoneOffsetUtil.js#L70) you'll see that we don't do the same remaindering process with `conversionOffset`, so we are actually essentially *including* the `clockDriftOffset` when a `conversionOffset` is involved.

The resulting messy logic in `computeMillisInCurrentDay` is because of this difference: when `conversionOffset` is nonzero, `clockDriftOffset` is included in the calculation of `time`, which makes matching up against schedules harder.

You might ask: does this suggest that we took the wrong approach with `conversionOffset`? (I am asking myself this 😛) But I think from a data accuracy standpoint we can't do any better. `conversionOffset`s result from large changes in display time due to incorrect pump set up. The idea of separating out part of that change as "clock drift" (by doing the same rounding and remaindering process we do with `timezoneOffset`) feels really artificial and arbitrary.

/end more-than-you-wanted-to-know

gniezen: Makes sense, thanks for the explanation :)

lib/medtronic/processData.js
gniezen: I agree, it's very unintuitive. I think QA would be the best way to make sure it works as expected, as we'll be able to compare against CareLink driver.

lib/medtronic/medtronicSimulator.js
gniezen: Indeed. When a _percent_ temp basal is running, the "Set/edit" option doesn't do anything. With a _rate_ temp basal, you can edit both the duration and the rate itself, after which it stops the current temp basal and starts a new one.

lib/medtronic/medtronicSimulator.js
gniezen: Done.

test/node/medtronic/testMedtronicSimulator.js
```diff
@@ -0,0 +1,574 @@
...
+          settings,
+          basal1.done(),
+          expectedTempBasal1,
+          expectedTempBasal2
+        ]);
+      });
+
+      it('checks temp basal over midnight without schedule change', function() {
+
+        settings.basalSchedules.standard[0].start = 7200000; // only change at 2am
```
gniezen: Yep, total sense. I've updated the test which then showed that I should be checking that the rate actually changes.

jebeck: Looked at the commit with the code update and test 👍

test/node/testCommonFunctions.js
jebeck: Kk.

gniezen:
@jebeck I've made more changes to this PR:
- also handles nested suppressed basals
- puts schedule names in suppressed scheduled basals
- restarting temp basals on resume

Are you ok with reviewing the new commits, or is there a better way to review this? (e.g. by splitting changes of into a new PR). I haven't split it yet, as `medtronic-device-event` is also branched off here and wanted to keep the number of branches (and potential conflicts) to a minimum.

jebeck:
Yeah I think I can just look at the new changes, that's fine @gniezen

test/node/medtronic/testMedtronicSimulator.js
```diff
@@ -0,0 +1,669 @@
...
+
+        expect(simulator.getEvents()).deep.equals([
+          settings,
+          basal1.done(),
+          expectedTempBasal1,
+          expectedTempBasal2
+        ]);
+      });
+
+      it('restarts temp basal after resume, with schedule change during suspend', function() {
```
jebeck: Maybe add another test case re: *not* resuming temp basal if it would have run its course completely by the time of the resumption?

jebeck:
Note that since `gniezen/medtronic` is not a protected branch, you can merge without me reviewing the next updates. Go ahead and do so if it's expedient for getting a test release out or something. I was almost going to select "Approve" to give you that freedom but then realized I wouldn't have to since `master` is the only protected branch...

lib/medtronic/medtronicSimulator.js
```diff
@@ -65,7 +83,220 @@ exports.make = function(config){
...
+              }
+            } else {
+              newBasal = config.builder.makeSuspendBasal();
+            }
+
+            var newJsDate = Date.parse(currBasal.deviceTime) + adjustedDuration;
+            newBasal.with_time(new Date(Date.parse(currBasal.time) + adjustedDuration).toISOString())
+              .with_duration(oldDuration - adjustedDuration)
+              .with_deviceTime(sundial.formatDeviceTime(newJsDate))
+              .with_timezoneOffset(currBasal.timezoneOffset)
```
gniezen: kk, fixed in commit 3aa6319

lib/medtronic/medtronicSimulator.js
```diff
@@ -65,7 +83,220 @@ exports.make = function(config){
...
+
+      if (currBasal != null) {
+
+        if(!currBasal.isAssigned('duration')) {
+          // calculate current basal's duration
+          var duration = Date.parse(event.time) - Date.parse(currBasal.time);
+          currBasal.duration = duration;
+        }
+
+        if (event.deliveryType === 'temp' && event.rate == null && event.duration === 0) {
```
gniezen: Oh wow, I had to look twice before I saw it. It's true what they say about being so deep in your own code that you don't notice the obvious anymore. And indeed, readability much easier after that tiny refactor.

lib/medtronic/medtronicSimulator.js
```diff
@@ -65,7 +83,220 @@ exports.make = function(config){
...
+              resumeBasal.index = event.resumeIndex;
+              delete resumeBasal.duration; // we don't know the new duration yet
+
+              // finish up the temp basal before the suspend
+              if(currBasal.duration) {
+                currBasal.with_expectedDuration(currBasal.duration);
+                var duration = Date.parse(event.time) - Date.parse(currBasal.time);
+                currBasal.with_duration(duration);
+
+                resumeBasal.duration = currBasal.expectedDuration - currBasal.duration - currStatus.duration;
```
gniezen: I implemented "restart temp basal on resume" after seeing that the CareLink driver does resume temp basals after a manually entered `suspend`. Just double-checked on the pump UI itself and the temp basal does indeed resume after a suspend. Phew, you had me worried for a bit there ;)

test/node/medtronic/testMedtronicSimulator.js
gniezen: Added the test, and that exposed that I wasn't properly checking if the temp basal completed before the resume, so fixed that too in commit 9d4085a6

lib/medtronic/medtronicSimulator.js
```diff
@@ -206,16 +211,44 @@ exports.make = function(config){
...
           };
           if(currBasal.deliveryType === 'temp') {
             // nest suppressed scheduled basal in temp basal
-            suppressed.suppressed = currBasal.suppressed;
+            event.suppressed.suppressed = currBasal.suppressed;
+
+            // also, restart temp basal on resume
+            if(currStatus && currStatus.index === event.index) {
+              // if the indexes are the same, the suspended basal was
+              // created from the same record as the suspend/resume event
```
jebeck: OK, that makes sense. Thanks for the explanation.

gniezen:
Merging this, so that only new changes are properly reflected in #333 for review.

lib/medtronic/medtronicSimulator.js
```diff
@@ -239,42 +239,46 @@ exports.make = function(config){
...
-              resumeBasal.deviceTime = sundial.formatDeviceTime(Date.parse(currStatus.deviceTime) + currStatus.duration);
-              resumeBasal.index = event.resumeIndex;
-              delete resumeBasal.duration; // we don't know the new duration yet
-
-              // finish up the temp basal before the suspend
-              if(currBasal.duration) {
-                currBasal.with_expectedDuration(currBasal.duration);
-                var duration = Date.parse(event.time) - Date.parse(currBasal.time);
-                currBasal.with_duration(duration);
+            if(currBasal.duration > (Date.parse(currStatus.time) - Date.parse(currBasal.time) + currStatus.duration)) {
```
jebeck: Trying to check my understanding here...

`Date.parse(currStatus.time) - Date.parse(currBasal.time)` is the duration of the `temp` basal prior to the suspension...

So the total of `(Date.parse(currStatus.time) - Date.parse(currBasal.time) + currStatus.duration)` is the duration of the temp prior to suspend and the duration of the suspend, yes?

If you can think of a good variable name, that whole chunk could be more readable pulled out prior to the `if`, like:

```JavaScript
var tempBeforeSuspensionAndSuspensionDuration = (Date.parse(currStatus.time) - Date.parse(currBasal.time) + currStatus.duration);
if (currBasal.duration > tempBeforeSuspensionAndSuspensionDuration)
```

But y'know, only if you can come up with a better variable name than the one I just did ;)

jebeck:
Have looked at all changes since the last review. Just the one comment that's probably mostly a check of my understanding (and potential readability fix).

lib/medtronic/medtronicSimulator.js
gniezen: Yep, it's the duration from when the temp basal started to when the suspend ended. How about `tempStartToResumeDuration`? Fixed in commit 4dfc03df.

jebeck: Yeah that name works for me, 👍

PR #329, opened 2016-09-19
Use dataservices endpoint, not platform

darinkrauss:
@gniezen Use the dataservices endpoints (rather than platform; using 'platform' for the entry was a mistake since we know there are multiple "platform services" that need to be exposed).

Subtle changes to the API since we have to change the endpoints anyhow.

Also removed old Clinic endpoints.

This can only be deployed after the Delete Errant Uploads work (which includes the platform related separation) is deployed to prd.

CC @jh-bate

### DIFF
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -494,7 +494,7 @@ api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb, uploa
         post_dataset_create(uploadItem, function(err, dataset){
           if(_.isEmpty(err)){
             api.log('created dataset');
-            datasetId = _.get(dataset, 'uploadId');
+            datasetId = _.get(dataset, 'data.uploadId');
             if(_.isEmpty(datasetId)){
               api.log('created dataset does not include uploadId');
               return callback(new Error(format('Dataset response does not contain uploadId.')));
--- a/main.js
+++ b/main.js
@@ -58,14 +58,6 @@ var contextMenus = [
   },
   {
     type: 'radio',
-    id: 'Clinic',
-    title: 'Clinic',
-    contexts: contexts,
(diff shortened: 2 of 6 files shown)


darinkrauss:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;AQ8qrDTf&#x2F;31-0-ux-2-dev-1-tst-uploader-update-dataservices-endpoint-and-new-dataset-return-value"> &lt;0:ux 2:dev 1:tst&gt; [uploader] Update dataservices endpoint and new dataset return value</a>

lib/core/api.js
```diff
@@ -494,7 +494,7 @@ api.upload.toPlatform = function(data, sessionInfo, progress, groupId, cb, uploa
         post_dataset_create(uploadItem, function(err, dataset){
           if(_.isEmpty(err)){
             api.log('created dataset');
-            datasetId = _.get(dataset, 'uploadId');
+            datasetId = _.get(dataset, 'data.uploadId');
```
gniezen: What necessitated this change?

main.js
```diff
@@ -58,14 +58,6 @@ var contextMenus = [
   },
   {
     type: 'radio',
-    id: 'Clinic',
```
gniezen: Oh nice, I forgot that we can remove this now!

lib/core/api.js
darinkrauss: I changed the format of the response to the create dataset API. It is a bit more JSONAPI like so that we can return data or errors plus other metadata about the request. I think this makes for a more flexible API.

darinkrauss:
Yes, I will merge into master and deploy a new uploader once the backend code is on prd.

PR #330, opened 2016-10-05
Fix for Animas missing records bug

gniezen:
See https://trello.com/c/skctayOf for more details.

In short: It's possible that a wizard record is completely empty. We only checked if all records are in sequence (e.g. 406, 407, 408) _after_ we delete empty records (which meant that we're getting a "missing record" error). Empty records are usually only found at the end, not in between other records, which is why this issue has not presented itself during testing. The solution is to delete empty records after the sequence check.

We also need to update the code that matches bolus records with wizard records, as we now know that some of these wizard records may be empty and cannot be matched up.

### DIFF
--- a/lib/drivers/animasDriver.js
+++ b/lib/drivers/animasDriver.js
@@ -1500,8 +1500,6 @@ module.exports = function (config) {
           firstPass = false;
 
         }, function(err, result) {
-          // remove empty and invalid records
-          _.remove(result, function (item) { return item.empty || item.jsDate === null;});
           cb(null,result);
         });
       }
@@ -2042,6 +2040,10 @@ module.exports = function (config) {
         }
 
         if(!missing) {
+
+          // remove empty and invalid records
+          _.remove(records, function (item) { return item.empty || item.jsDate === null;});
+
           return records;
         } else {
           return false;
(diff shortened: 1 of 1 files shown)


gniezen:
Pinging @jh-bate for review?

gniezen:
[[uploader][bug] Animas upload fails due to missing records](https://trello.com/c/skctayOf/22-uploader-bug-animas-upload-fails-due-to-missing-records)

gniezen:
Pinging @jebeck for review instead.

lib/drivers/animasDriver.js
```diff
@@ -1601,16 +1599,16 @@ module.exports = function (config) {
     var wizard = null;
     for(var b in data.wizardRecords) {
       var wizarddatum = data.wizardRecords[b];
-      var bolusdatum = postrecords[b];
+      var bolusdatum = _.findWhere(postrecords, { 'index': wizarddatum.index, 'syncCounter': wizarddatum.sync_counter });
```
jebeck: I think it's best to stick to `_.find` in general. Same args...the only difference is `_.findWhere` will do a deep comparison where applicable, which is less efficient (and we had a problem once with this in blip *ages* ago, so I have a knee-jerk "avoid lodash `*where` methods!" reaction).

PR #331, opened 2016-10-07
Download binary blob and POST data, even if upload fails

gniezen:
The Medtronic driver is able to provide us with both the raw binary (blob) data and processed (POST) data. The changes in this PR allows you to download both files, even if the upload has failed.

### DIFF
--- a/lib/components/Upload.js
+++ b/lib/components/Upload.js
@@ -119,6 +119,45 @@ export default class Upload extends Component {
     }
   }
 
+  getDebugLinks(data) {
+
+    if(Array.isArray(data.post_records)) {
+
+      let filename = 'uploader-processed-records.json';
+      let jsonData = JSON.stringify(data.post_records, undefined, 4);
+      let blob = new Blob([jsonData], {type: 'text/json'});
+      let dataHref = URL.createObjectURL(blob);
+
+      let binary_link = null;
+      if(data.pages) {
+        let filenameBinary = 'binary-blob.json';
+        let jsonDataBinary = JSON.stringify(data.pages, undefined, 4);
+        let blobBinary = new Blob([jsonDataBinary], {type: 'text/json'});
+        let dataHrefBinary = URL.createObjectURL(blobBinary);
+        binary_link = (
+          <a href={dataHrefBinary}
+            className={styles.dataDownloadLink}
+            download={filenameBinary}
+            data-downloadurl={['text/json', filenameBinary, dataHrefBinary].join(':')}>
+            Binary blob
+          </a>
+        );
+      }
+      return (
+        <div>
+          <a href={dataHref}
+            className={styles.dataDownloadLink}
+            download={filename}
+            data-downloadurl={['text/json', filename, dataHref].join(':')}>
+            POST data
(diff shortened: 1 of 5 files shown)


gniezen:
Pinging @jebeck for review. This will be part of the upcoming `medtronic-device-event` PR as well, but thought you may want to look at this separately beforehand to make sure I'm doing this right?

lib/components/Upload.js
```diff
@@ -119,6 +119,45 @@ export default class Upload extends Component {
...
+
+    if(Array.isArray(data.post_records)) {
+
+      let filename = 'uploader-processed-records.json';
+      let jsonData = JSON.stringify(data.post_records, undefined, 4);
+      let blob = new Blob([jsonData], {type: 'text/json'});
+      let dataHref = URL.createObjectURL(blob);
+
+      let binary_link = null;
+      if(data.pages) {
```
jebeck: Wondering a bit about the nested `if` here. Do you truly want to *not* provide the binary download dump if `data.post_records` isn't there? (Or perhaps the latter is guaranteed to be there at least as an empty array...but still reads a bit funny if there isn't a logical connection, but perhaps there is and I'm missing it.)

jebeck: Also wonder if the same Array check may be useful here?

lib/components/Upload.js
```diff
@@ -338,23 +377,20 @@ export default class Upload extends Component {
...
-        let dataHref = URL.createObjectURL(blob);
-        dataDownloadLink = (
-          <a href={dataHref}
-            className={styles.dataDownloadLink}
-            download={filename}
-            data-downloadurl={['text/json', filename, dataHref].join(':')}>
-            Download POST data
-          </a>
-        );
+      if (__DEBUG__ && this.props.upload.data) {
```
jebeck: Perhaps throw the `!_.isEmpty` check back in? Again, not sure what's guaranteed here...been too long since I was elbow deep in this stuff.

lib/core/device.js
```diff
@@ -357,10 +357,7 @@ device.detectAll = function(cb) {
 device.upload = function(driverId, options, cb) {
   var dm = this._createDriverManager(driverId, options);
   dm.process(driverId, function(err, result) {
-    if (err) {
-      return cb(err);
-    }
-    return cb(null, result.post_records);
+    return cb(err, result);
```
jebeck: Ah, so this is how you're getting the data through even in the failure case!

gniezen:
[[uploader] Medtronic device events](https://trello.com/c/6b5oFTZl/17-uploader-medtronic-device-events)

gniezen:
[[uploader] Button to get raw binary data from pump in debug mode](https://trello.com/c/1IFzNbHP/42-uploader-button-to-get-raw-binary-data-from-pump-in-debug-mode)

lib/components/Upload.js
gniezen: You're right, they shouldn't be nested. Have also changed to use Array check.

gniezen:
Addressed all comments, merging.

PR #332, opened 2016-10-07
Add checks for Data Storage

hntrdglss:
Hey, @krystophv and @jebeck! Very excited to submit my first pull request. The functional changes here are pretty minimal, but I feel like I learned a lot about the structure of the Uploader, data models, and account types through this enhancement. Here are a few points I'd like to specifically call out for feedback:

- **Naming conventions**: if you have suggestions for naming anything differently (in particular the `DataStorageCheck` [page](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/no-dsa-check?expand=1#diff-f43d6abf0dd446d37fa39faa2fbd71fbR22) and [component](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/no-dsa-check?expand=1#diff-6adb8bfb0bd1a9b07a1faf0491490354R26)), please let me know.

- **Memberships**: from what I've seen, it looks like an account always has it's own ID in the memberships array. I did a check for [`memberships.length > 1`](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/no-dsa-check?expand=1#diff-487ce994eadf6647e8b5c39b770c9ec9R41) on accounts that do not have a patient object in their profile. That seemed like the simplest and least computational check, but if this isn't always assured, I can change it to something like `_.difference(_.keys(memberships), loggedInUser).length`, so we check the array for user IDs not of the current user.

- **CSS Override**: I ran into a CSS specificity issue on the "Set up data storage" button, due to [this selector](https://github.com/tidepool-org/chrome-uploader/blob/master/styles/core/buttons.module.less#L22). It was overriding my [padding and margin declarations](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/no-dsa-check?expand=1#diff-e36fff856d6e3db38812efd11885c47eR36) on the `A` element. Rather than using `!important`, I set the button style on a [`DIV` and wrapped that in the link](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/no-dsa-check?expand=1#diff-6adb8bfb0bd1a9b07a1faf0491490354R38). I couldn't find any similar instances of this situation—looks like this is a rather unique button—but let me know if I should resolve this differently. (I also tried `a.buttonCta`, but that resulted in an error. I guess element-based selectors aren't allowed in CSS Modules.)

- **Updates to fixtures**: I had done some work with updating the fixtures to include a clinic account and no-DSA variants, but that started to convolute the code changes for this specific functionality. It could probably tie in to some work on [this tech debt](https://trello.com/c/mmMR0qpw) card. I'll submit that in a separate PR.

Thanks!


lib/redux/actions/async.js
```diff
@@ -140,8 +140,12 @@ export function doLogin(creds, opts) {
         memberships: results[2]
       }));
 
-      // detect if a VCA here and redirect to clinic user select screen
+      // detect if a DSA here and redirect to data storage screen
+      if (!personUtils.userHasDSA(results[0].user, results[1], results[2])) {
```
jebeck: I think the approach here could be simplified considerably by getting rid of the utility function and doing a `getState` check. Since `loginSuccess` is already dispatched, I believe the state should be updated with the results, and then you can just check for the length of the `targetUsersForUpload` array. Perhaps give that a try and see if it covers all the same cases. I think it will. If you think about it, we're already doing the kind of logic that is in that utility function in our reducers when we update the state based on the results of logging in...

lib/components/DataStorageCheck.js
```diff
@@ -0,0 +1,44 @@
...
+* This program is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+* FOR A PARTICULAR PURPOSE. See the License for more details.
+*
+* You should have received a copy of the License along with this program; if
+* not, you can obtain one from Tidepool Project at tidepool.org.
+* == BSD2 LICENSE ==
+*/
+
+var _ = require('lodash');
```
jebeck: For future reference, we write new components in ES6 :)

lib/components/DataStorageCheck.js
```diff
@@ -0,0 +1,44 @@
...
+      <div>
+        <div className={styles.main}>
+          <a className={styles.linkCentered}
+            href={urls.ADD_NEW_PATIENT}
+            target="_blank">
+            <div className={styles.buttonCta}>
+              Set up data storage
+            </div>
+          </a>
+          <p className={styles.paragraph}>Or, ask the person you are uploading for to grant you access to upload.<br/> <a href={urls.HOW_TO_SHARE_KB_ARTICLE}>How?</a></p>
```
jebeck: Locally the "How?" is not showing up for me as a clickable link (both interaction-wise and styling wise). Double-check? I inspected, and definitely the `href` is getting dropped somewhere, not sure why.

lib/redux/constants/otherConstants.js
```diff
@@ -46,5 +47,7 @@ export const steps = {
 
 export const urls = {
   DRIVER_DOWNLOAD: 'http://tidepool.org/downloads/',
-  HOW_TO_UPDATE_KB_ARTICLE: 'https://tidepool-project.helpscoutdocs.com/article/6-how-to-install-or-upgrade-the-tidepool-uploader-gen'
+  HOW_TO_UPDATE_KB_ARTICLE: 'http://support.tidepool.org/article/6-how-to-install-or-upgrade-the-tidepool-uploader-gen',
+  ADD_NEW_PATIENT: 'https://blip.tidepool.org/patients/new',
```
hntrdglss: Excellent point! I'll look for that.

lib/redux/actions/async.js
hntrdglss: Ah! Yes, that **would** be preferred. When I first tried solving this, I did it as part of `retrieveTargetsFromStorage()`, but moved it earlier to short circuit some additional executions. Totally missed that I should have just been able to check `targetUsersForUpload`. 👍

lib/components/DataStorageCheck.js
hntrdglss: Oops! Yes, I had the wrong variable name. Caught the issue on Friday, but forgot to push. 😞

hntrdglss:
@jebeck Regarding naming: that was my qualm, too. This is more a roadblock than a check. I like `NoUploadTargets`/`NO_UPLOAD_TARGETS`, so I went with that. @krystophv, if you have any other suggestions, I'm all ears!

lib/components/DataStorageCheck.js
jebeck: We don't really have any plans for converting old code. The idea has just been to move forward with ES6 and perhaps make a judgment call when making modifications to old code. That is, if the modifications seem large enough, port to ES6. If small, just make changes in ES5.

Here in the uploader we also have to be a bit careful because the CLI tools used for uploading from certain file types need ES5 (due to our continued use of node v0.12, though of course that may change soon as well). So we have to be careful that anything in the require chain for the CLI tools does not use ES6 things that node v0.12 doesn't support. This is actually broken on master, but @gniezen has patched on his Medtronic branch.

lib/components/LoggedInAs.js
```diff
@@ -67,18 +68,33 @@ var LoggedInAs = React.createClass({
   },
 
   renderChooseDevices: function() {
+    var title = '';
     var uploadInProgress = this.props.isUploadInProgress;
+    var isDisabled = uploadInProgress;
 
-    if(this.props.isClinicAccount){
+    if (this.props.isClinicAccount) {
```
jebeck: That's a comment on the whitespace change 😉

lib/components/LoggedInAs.js
```diff
@@ -67,18 +68,33 @@ var LoggedInAs = React.createClass({
...
       return null;
     }
+
+    if (_.isEmpty(this.props.targetUsersForUpload) ||
+        this.props.targetUsersForUpload.length < 1) {
```
jebeck: Looks like we initialize `targetUsersForUpload` as an empty array, so I think just one of these checks should be sufficient, no? My preference would be to keep the `_.isEmpty`.

lib/components/DataStorageCheck.js
jebeck: So I see references to this in the latest commit message re: updates after design review, but when I run locally after pulling the "How?" is still not a link for me.

lib/redux/actions/async.js
```diff
@@ -140,8 +141,13 @@ export function doLogin(creds, opts) {
         memberships: results[2]
       }));
 
-      // detect if a VCA here and redirect to clinic user select screen
+      // detect if a DSA here and redirect to data storage screen
+      const { targetUsersForUpload } = getState();
+      if (!targetUsersForUpload.length) {
```
jebeck: I'm not a fan of the truthiness/lack of consistency here. Can we use the same check as above, perhaps `_.isEmpty`?

lib/redux/actions/sync.js
```diff
@@ -132,6 +132,14 @@ export function setForgotPasswordUrl(url) {
   };
 }
 
+export function setNewPatientUrl(url) {
```
jebeck: Add unit tests for this action creator plz.

lib/components/DataStorageCheck.js
jebeck: Disregard comment about not seeing proper styling for "How?" My bad, could have sworn loaded fresh version but was still looking at older version. Reloading extension locally fixed.

jebeck:
Please add the requested action creator and reducer unit tests before merging, but otherwise good 😎

lib/components/LoggedInAs.js
jebeck: Heh, yeah, there might have been a stage when we weren't defaulting `targetUsersForUpload` to an empty array I suppose? Go ahead and change there too.

PR #333, opened 2016-10-11
Medtronic device events

gniezen:
Adds support for:
- alarms
- prime events (cannula/tubing)
- reservoir changes (rewind events)
- status events
- time change events

For more detail, see https://trello.com/c/6b5oFTZl

Note: #328 should be merged first. This PR also includes the code to download the binary blob reviewed in #331.


gniezen:
[[uploader] Medtronic device events](https://trello.com/c/6b5oFTZl/17-uploader-medtronic-device-events)

gniezen:
[[uploader] Button to get raw binary data from pump in debug mode](https://trello.com/c/1IFzNbHP/42-uploader-button-to-get-raw-binary-data-from-pump-in-debug-mode)

gniezen:
Pinging @jebeck for review.

lib/medtronic/processData.js
```diff
@@ -638,19 +615,170 @@ var buildBasalRecords = function (records) {
...
+    var type = fixed === 0 ? 'tubing' : 'cannula';
+
+    var deviceEvent = cfg.builder.makeDeviceEventPrime()
+      .with_deviceTime(sundial.formatDeviceTime(primedatum.jsDate))
+      .with_primeTarget(type)
+      .with_volume(amount)
+      .set('index', primedatum.index);
+
+    if(type === 'cannula' && fixed !== amount) {
+      // it's possible to cancel a cannula prime,
```
jebeck: Huh. I wonder if that means our data model should be expanded to have an `expectedVolume` for primes... Of dubious clinical relevance, however...

lib/medtronic/processData.js
```diff
@@ -638,19 +615,170 @@ var buildBasalRecords = function (records) {
...
+      .set('index', rewinddatum.index);
+    cfg.tzoUtil.fillInUTCInfo(deviceEvent, rewinddatum.jsDate);
+
+    deviceEvent = deviceEvent.done();
+    postrecords.push(deviceEvent);
+  });
+  return postrecords;
+}
+
+function buildTimeChangeRecords(records, mostRecent) {
```
jebeck: So you haven't tested yet with a 530G that has CGM data, correct?

lib/medtronic/processData.js
```diff
@@ -638,19 +615,170 @@ var buildBasalRecords = function (records) {
...
+          default:
+            alarmRecord = alarmRecord.with_alarmType('other');
+            var payload = {alarm_id: alarmValue};
+            if(alarmText !== 'unknown') {
+              payload.alarm_text = alarmText;
+            }
+            alarmRecord = alarmRecord.with_payload(payload);
+        }
+        break;
+      case RECORD_TYPES.LOW_RESERVOIR.value:
```
jebeck: Does MedT not provide any info about how much insulin is left to stick in the payload?

lib/medtronic/medtronicSimulator.js
```diff
@@ -222,19 +266,25 @@ exports.make = function(config){
             };
             event.suppressed = suppressed;
 
+            if(!event.rate) {
+              event = event.with_rate(currBasal.rate * event.percent);
```
jebeck: Worth doing a check here first to see if `event.percent` exists? And throw an error if it's not there? It won't exist if the `temp` is manual, I imagine...

lib/medtronic/medtronicSimulator.js
```diff
@@ -299,20 +349,25 @@ exports.make = function(config){
...
-        currWizard = currWizard.done();
-        simpleSimulate(currWizard);
+      if(currWizard) {
+        // A wizard event is always followed by a bolus event.
+        // Usually the timestamp is the same, but sometimes the bolus timestamp
+        // is a second later. Here we also check that the bolus event is at least
+        // within 5 seconds of the wizard event.
+        var bolusTime = event.jsDate.valueOf();
+        var wizardTime = currWizard.jsDate.valueOf();
+        if ( (bolusTime >= wizardTime) && ( bolusTime < (wizardTime + 5000)) ) {
```
jebeck: Did you get this threshold from the CareLink code? (Honestly can't remember.)

lib/medtronic/processData.js
gniezen: Nope, we're still trying to get 530G to upload successfully - so far only the 523 works (as it's the one I have access to).  CGM data will only be addressed in https://trello.com/c/vMtU8xk, which has not been prioritized yet.

lib/medtronic/processData.js
gniezen: I just checked the CSV and indeed the amount is listed there, so I'll try to identify it in the raw data and add to payload.

gniezen: Added in commit 43a741d7f5.

lib/medtronic/medtronicSimulator.js
gniezen: Can't hurt, I've now put in a check with commit cfdbec4ffd

lib/medtronic/medtronicSimulator.js
gniezen: No, I didn't. The bolus record timestamp and wizard timestamp are usually exactly the same, but if they happen at the edge of a second the bolus timestamp may be recorded during the next second. I chose 5 seconds just to make sure that even if there's some short delay that somehow occurs between them they'll still match up. If you think this can cause potential issues, we could change it from 5s to 1s? (Note that this is not an instance of the wizard timestamp only being recorded after the initial bolus finishes, as the wizard record occurs first.)

lib/medtronic/processData.js
jebeck: K, just be aware that the bootstrapping may be "fun" for the CGM data 😉 . At least for how we got the data through the CSV, we had to bootstrap the CGM stream of data separately because it's separate on the device - different index and records it's own time changes.

PR #341, opened 2016-11-30
fail unsupported MedT pump models (all but *23 & 530G/Veo)

jebeck:
What it says on the tin, *plus* I also removed some dead/redundant code in the CareLink driver that was doing a late check for incorrect CareLink creds (and updated testing instructions accordingly). The new check in the redux thunk for CareLink fetch precedes it and works just fine in my testing, so I cleaned it up a bit. (Couldn't help myself when I saw the redundancy.)

The way I shoehorned in the failure of unsupported pump models is probably not ideal, but it's expedient, and we're hoping the CareLink code goes away quite soon, so I thought it plenty sufficient. There are tests around where I changed the logic in the utility for removing overlapping uploads (which was and is the only place we look at model numbers) and tests around the redux action utility that handles errors, as I had to make sure a `code` on a passed-in error object would take precedence over the default error code configured. I can't imagine that change having any unexpected ill effects - only good ones (i.e., we may start surfacing more detailed error messages where we had seen default-ish ones).

Ping @gniezen for review.


jebeck:
[<ux:1 dev:3 tst:1> Medtronic *22 uploads via CareLink should fail gracefully](https://trello.com/c/AFhl4YKo/60-ux-1-dev-3-tst-1-medtronic-22-uploads-via-carelink-should-fail-gracefully)

PR #340, opened 2016-11-30
Add Medtronic serial number field on device selection

hntrdglss:
FYI: **I *won't* be merging this, since it lacks functionality from @gniezen's side.** Purely to ease reviewing a smaller piece of the bigger puzzle. I'll close the PR once reviewed. Not much to it, but I used it as a chance to learn more about how devices are set up and uploaded! 🎉

@gniezen, a few of metric things from the [Trello card](https://trello.com/c/TiTc17UL/) I didn't see a good place to plug in to here, since it's based the master branch as we discussed:
 - Medtronic upload attempted (I went ahead and [defined here](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/medtronic-sn?expand=1#diff-447a548497d1fa1e1d01bd6f9e53ac2aR47))
 - Medtronic upload successful (I went ahead and [defined here](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/medtronic-sn?expand=1#diff-447a548497d1fa1e1d01bd6f9e53ac2aR48))
 - Medtronic upload failed (I went ahead and [defined here](https://github.com/tidepool-org/chrome-uploader/compare/hntrdglss/medtronic-sn?expand=1#diff-447a548497d1fa1e1d01bd6f9e53ac2aR49))
 - ~~and also the "remember me" checkbox, but that'll be implemented in a [later card](https://trello.com/c/hYZjKBPK/)~~

If I can help adding those on another branch, or help any with integrating this into the overall branch, please let me know!

PR #343, opened 2016-11-30
Add Medtronic serial number field on device selection

hntrdglss:
This PR adds a serial number field for direct Medtronic device uploads. The serial number field should require no more or fewer than six digits, and should ONLY allow numeric input.

It also adds new metrics when clicking on "Choose devices" on the user menu and when the "Done" button is clicked on the device selection page.

See  [Trello card](https://trello.com/c/TiTc17UL/) for more info.

@krystophv: mind taking a look when you have time? @gniezen reviewed and merged in his branch from the previous PR (#340), so it'd be nice to have another set of eyes.


lib/components/Upload.js
```diff
@@ -119,6 +137,32 @@ export default class Upload extends Component {
     }
   }
 
+  onMedtronicSerialNumberInputChange() {
+    const { refs } = this;
+    let serial_number = refs.medtronic_serial_number;
```
krystophv: I think I would prefer in this instance not to rely on grabbing the value from `refs` when it's being passed into the function through the event (via `event.target.value`). Unless there's something about the `ref` pattern here that I'm missing.

lib/components/Upload.js
```diff
@@ -293,6 +343,28 @@ export default class Upload extends Component {
...
+    if (_.get(upload, 'source.driverId', null) !== 'Medtronic') {
+      return null;
+    }
+
+    return (
+      <div>
+        <div className={styles.textInputWrapper}>
+          <p>Enter your 6 digit serial number found on the back of your pump.</p>
+          <input
+            type="number"
```
krystophv: A strange side-effect of using the `number` input type to limit input on this field is that 'e' is actually a valid input due to scientific notation. Not sure if it'll come up in the field, but we might need to do some additional filtering/validation on the handler for the input update.

lib/components/Upload.js
hntrdglss: No specific reason, that's just how the CareLink inputs are handled, so I used the same for consistency.

lib/components/Upload.js
hntrdglss: Ooh, good catch. Probably unlikely as you mentioned, but I can definitely add something to `onMedtronicSerialNumberInputChange`.

hntrdglss: So tricky! `input type="number"` doesn't emit an `onChange` event when the input is `123e` until something comes AFTER the `e` (like `123e4`), so it was still allowing the `e` to be entered. I switched the `input` type to `text`, to resolve that.

krystophv:
The fact that it doesn't emit a change event until something comes after the 'e' is pretty incredible (and actually sounds like a bug). The solution seems pretty robust. LGTM.

PR #346, opened 2016-12-07
No user name

jh-bate:
@gniezen see https://trello.com/c/aCrFMbXm for details. 

Turns out if `VCA` had data storage attached then we would see it in the searchable accounts and it would be preselected at the top of the list


jh-bate:
[VCA Uploader should recognize when an account has setup a DSA and ignore it [Uploader] <0:ux 1:dev 1:tst>](https://trello.com/c/aCrFMbXm/67-vca-uploader-search-bar-should-not-be-pre-filled-with-a-user-name-uploader-0-ux-1-dev-1-tst)

test/browser/redux/reducers/users.test.js
```diff
@@ -562,6 +562,42 @@ describe('users', () => {
...
+        expect(users.targetUsersForUpload(undefined, {
+          type: actionTypes.SET_ALL_USERS,
+          payload: { user, profile, memberships }
+        })).to.deep.equal(['d4e5f6', 'x1y2z3', '888']);
+      });
+      it('should handle normal accounts', () => {
+        const profile = {a: 1};
+        expect(users.targetUsersForUpload(undefined, {
+          type: actionTypes.SET_ALL_USERS,
+          payload: { user, profile, memberships }
```
gniezen: It took me a while to find where `user` and `memberships` are defined for this specific test. Just want to make sure that you did intend to use the ones defined at the beginning of the test suite?

PR #347, opened 2016-12-07
Medtronic remaining work

gniezen:
See https://trello.com/c/qkp0Mfj6 for more details.


gniezen:

gniezen:
@jebeck There are still two outstanding items on the checklist for the attached card - I'll put them in a separate PR. Hoping to to get this stuff merged into the main Medtronic branch in the meanwhile.

jebeck:
Remove your npm shrinkwrap file too please? I assume you didn't mean to check that in...

lib/redux/actions/async.js
```diff
@@ -236,6 +236,7 @@ export function doDeviceUpload(driverId, opts, utc) {
     const version = versionInfo.semver;
     const { devices, os, targetTimezones, uploadTargetUser } = getState();
     const targetDevice = _.findWhere(devices, {source: {driverId: driverId}});
+    opts = opts || {};
```
jebeck: Perhaps move this into the function signature? ES6 includes default params that are super nice/convenient: `export function doDeviceUpload(driverId, opts = {}, utc)`

lib/drivers/medtronicDriver.js
```diff
@@ -1004,42 +1009,46 @@ module.exports = function (config) {
...
-                simulator.prime(datum);
-              } else if (datum.subType === 'reservoirChange') {
-                simulator.rewind(datum);
-              } else if (datum.subType === 'timeChange') {
-                simulator.changeDeviceTime(datum);
-              }
-              break;
-            default:
-              debug('[Hand-off to simulator] Unhandled type!', datum.type);
+        try {
```
jebeck: Why is this part in a `try` block too? All I can think of is if `postrecords` is empty, but that should be fine instead of the `for` loop... 🤔 Perhaps because of errors thrown in the simulator, I guess?

lib/medtronic/processData.js
```diff
@@ -236,12 +233,21 @@ var buildWizardRecords = function (records) {
...
-      var correctionRaw = [twosComplement(struct.extractByte(record.body,5) & BITMASKS.RIGHT_FOUR)];
-      correctionRaw = correctionRaw.concat(twosComplement(struct.extractByte(record.body,7)));
+      var correctionRaw = [twosComplement(struct.extractByte(record.body,5) & BITMASKS.RIGHT_FOUR, 8)];
+      correctionRaw = correctionRaw.concat(twosComplement(struct.extractByte(record.body,7), 8));
       recommended.correction = struct.extractBEShort(correctionRaw,0) / cfg.settings.strokesPerUnit;
 
       recommended.net = struct.extractBEShort(record.body,11) / cfg.settings.strokesPerUnit;
       iob =  struct.extractBEShort(record.body,9) / cfg.settings.strokesPerUnit;
+
+      units = 'mg/dL'; // TODO: figure out these units when we get hold of pump model with smaller record format
```
gniezen: I think only pumps older than 523? So there is the potential for me to take out those sections of code if we decide to not support those devices - I'm expecting that we'll have a better idea after alpha testing, and get it sorted out before we go to production.

gniezen:
@jebeck `npm-shrinkwrap.json` is necessary to force the `difflet` package that I'm using in the comparison CLI tool to use a specific version of a required package.

It looks like `mversion` also updates the version field in the shrinkwrap file, so if you're just referring to the version update line, i could try to remove that?

lib/medtronic/processData.js
```diff
@@ -236,12 +233,21 @@ var buildWizardRecords = function (records) {
...
+    }
+
+    if(units == 'mmol/L') {
```
gniezen: Yeah, it will be sent as 78 if the value is 7.8 mmol/L. Not sure how exactly floating point math will cause issues in this specific instance, but I've seen enough strange things with floating point to agree that it's better to play it safe.

.bookignore
```diff
@@ -2,7 +2,7 @@ build/
 config/
 dist/
 fonts/
-images/
+./images/
```
gniezen: No, I'm still on El Capitan. This is a weird one - my guess is that we're somehow not using the same version of Gitbook, even though I removed my global installation and did a `rm -rf` and reinstallation of local modules.

lib/redux/actions/async.js
gniezen: Ha, I totally forgot that I'm allowed to use ES6 in that file, even though there are ES6 `const`s everywhere! :)

jebeck:
Huh, for some reason GitHub didn't e-mail me w/any of your responses :( Sounds fine if the shrinkwrap is necessary, I guess. I didn't see a diff at all for it - suppressed for being too big, so that's why I assumed it was your first check-in.

PR #352, opened 2016-12-13
Search bug

jh-bate:
@hntrdglss this *finally* resolves the issue we were having if you could take a quick look


jh-bate:
[<0:ux 3:dev 1:tst> [Bug] VCA Uploader search bar doesn't accept some characters in some situations](https://trello.com/c/t9opqsm5/73-0-ux-3-dev-1-tst-bug-vca-uploader-search-bar-doesn-t-accept-some-characters-in-some-situations)

PR #354, opened 2016-12-14
Medtronic CGM data

gniezen:
See https://trello.com/c/vMtU8xkK for more details


gniezen:
[Medtronic CGM data [uploader] [dev: 5 tst: 2 ux: 0]](https://trello.com/c/vMtU8xkK/19-medtronic-cgm-data-uploader-dev-5-tst-2-ux-0)

gniezen:
@jebeck Phew, finally ready for review. Yes, I did underestimate the difficulty of this one. This was mostly related to very few CBG values having timestamps, and having to interpolate times. Doing this across time changes also provided a couple of headaches ;)

lib/drivers/medtronicDriver.js
```diff
@@ -975,87 +1044,113 @@ module.exports = function (config) {
...
+
+          var cgmDateRecords =  _.filter(cgmRecords, function(event) {return event.jsDate;}); // filter out objects without dates
+          if(cgmDateRecords.length === 0) {
+            console.log('No CBG records found on pump');
+          } else {
+            var cgmMostRecent = sundial.applyTimezone(cgmDateRecords[cgmDateRecords.length-1].jsDate, cfg.timezone).toISOString();
+            console.log('CGM time changes:');
+            var cgmResult = proc.buildTimeChangeRecords(cgmRecords, cgmMostRecent, 'cgm');
+            // only pump time change events are uploaded
+            cfg.cgmTzoUtil = cgmResult.tzoUtil;
```
jebeck: See my big comment on https://trello.com/c/pGmiXf4s regarding the potential challenge of bootstrapping both the pump and CGM data if/when a user goes on a pump vacation but continues using the CGM sensor *and* has a time change (or changes that don't cancel each other out) during that pump vacation period.

jebeck: One of the partial solutions we stuck in the CareLink driver was basically [a length check on the time changes arrays for the pump vs. CGM data](https://github.com/tidepool-org/chrome-uploader/blob/master/lib/drivers/carelinkDriver.js#L327-L343). Just sticking something like that in to get an idea of whether the same problem occurs with the direct data reading would maybe be a start?

lib/medtronic/processData.js
```diff
@@ -139,24 +165,39 @@ var init = function(config, settings) {
   cfg.settings = _.clone(settings);
 };
 
+var updateConfig = function(config) {
```
jebeck: I tracked this down earlier, and now seeing it's a new add... What's the motivation behind `_.clone`-ing the config in the driver in the first place? Mutation all over the place isn't necessarily my first design choice but if memory serves it's quite common in the device code - common enough that it stuck out at me when I saw you adding properties to `cfg` in the driver and *then* calling this update function.

lib/medtronic/processData.js
```diff
@@ -426,6 +467,84 @@ function buildBGRecords(records) {
   return postrecords;
 }
 
+function buildCGMRecords(events) {
```
jebeck: I'd suggest that @pazaan might be better suited to reviewing the intricacies here, if y'all haven't already discussed them.

lib/drivers/docs/medtronicCGM.md
```diff
@@ -0,0 +1,98 @@
...
+ - `*[ ]` TODO: needs implementation!
+ - `[ ]` unavailable in data protocol and/or not documented in spec and not yet implemented)
+
+### Required if Present
+
+#### CBG
+
+  - `[x]` cbg values
+  - `[?]` units of cbg values (read from device, not hard-coded)
+  - `[ ]` out-of-range values (LO or HI)
```
gniezen: Eric confirmed that HI values are not synced from the meter to his pump either, and that readings of 20 and below always register as control solution tests, which are not synced from the meter to the pump either.

lib/drivers/medtronicDriver.js
```diff
@@ -428,33 +451,32 @@ module.exports = function (config) {
                     return callback(err, null);
                   }
                   if (result) {
-                    if(result.resultsInReply) {
 
-                      var header = struct.extractBytes(result.bytes, 0, 2);
-                      console.log('Last header: ', common.bytes2hex(header));
-                      if(header[0] !== 0x51 && header[1] !== 0x01) {
+                    var header = struct.extractBytes(result.bytes, 0, 2);
+                    console.log('Last header: ', common.bytes2hex(header));
```
gniezen: I was thinking of leaving it in until after alpha testing, to make it easier to debug any issues. That reminds me that I should tell Brandon so that he doesn't start measuring upload times as part of alpha testing, as the logging as a big effect on time to upload.

lib/drivers/medtronicDriver.js
```diff
@@ -975,87 +1044,113 @@ module.exports = function (config) {
...
+          cfg.tzoUtil = result.tzoUtil;
+
+          var cgmDateRecords =  _.filter(cgmRecords, function(event) {return event.jsDate;}); // filter out objects without dates
+          if(cgmDateRecords.length === 0) {
+            console.log('No CBG records found on pump');
+          } else {
+            var cgmMostRecent = sundial.applyTimezone(cgmDateRecords[cgmDateRecords.length-1].jsDate, cfg.timezone).toISOString();
+            console.log('CGM time changes:');
+            var cgmResult = proc.buildTimeChangeRecords(cgmRecords, cgmMostRecent, 'cgm');
+            // only pump time change events are uploaded
```
gniezen: This comment meant that only *pump* time change events are upload, not *CGM* time change events, as I expected them to be the same. After your comment on the Trello card I now realize that it's possible they may be different, so I'm now changing it to upload both.

gniezen: Oh, and I'm also sticking the device type (`cgm`/`pump`) in the `timeChange` payload.

lib/medtronic/processData.js
gniezen: Hmm, good point. IIRC I created a separate `cfg` object that I could store to a file as part of the binary blob. Now that this functionality is part of `Upload.js`, it should be fine to remove and simplify the code somewhat.

Update: A lot trickier than I thought, haha. Turns out I defined `settings` twice, which took a while to figure out. Changes in commit 75a5328

lib/medtronic/processData.js
gniezen: Ah yes, that's a good idea! @pazaan , will you please review the `buildCGMRecords` function? I'll send you a binary blob containing CGM data to use until you get your BCN Link meter.

lib/drivers/docs/medtronicCGM.md
jebeck: Oh, that applies to HI CGM readings as well? I think I misunderstood that discussion because I thought you guys were just talking about smbg readings.

lib/medtronic/processData.js
```diff
@@ -426,6 +464,84 @@ function buildBGRecords(records) {
...
+  var recordsSinceTimestamp = null;
+
+  // we have to work through the CGM records in reverse, as sensor timestamps
+  // affect previous values and are subject to time changes
+
+  for( var i = events.length - 1 ; i >= 0; i-- ) {
+    var event = events[i];
+
+    if(event.jsDate != false) {
+      console.log('CGM event:',sundial.formatDeviceTime(event.jsDate), event.type.name, common.bytes2hex([event.head]),common.bytes2hex(event.date),common.bytes2hex(event.body),common.bytes2hex(event.descriptor) );
```
pazaan: Should `debug` be used instead of `console.log`?

lib/medtronic/processData.js
```diff
@@ -426,6 +464,84 @@ function buildBGRecords(records) {
...
+
+          var isig = struct.extractShort(event.descriptor, offset * 2);
+
+          if(isig !== 0x22 &&  isig > 255) {
+            var cbg = (isig & 0x01) + (event.body[offset] << 1);
+
+            var date = sundial.applyOffset(start, -(recordsSinceTimestamp * 5));
+            recordsSinceTimestamp += 1;
+
+            console.log('CBG:', sundial.formatDeviceTime(date), common.bytes2hex(struct.extractBytes(event.descriptor, offset * 2, 2)),',',common.bytes2hex([struct.extractByte(event.body,offset)]),',',(isig & 0x01) + (event.body[offset] << 1),',',(isig  & 0xFFFE) / 100.0);
```
pazaan: Replace `(isig & 0x01) + (event.body[offset] << 1)` with `cbg`?

lib/medtronic/processData.js
```diff
@@ -426,6 +464,84 @@ function buildBGRecords(records) {
...
+        recordsSinceTimestamp = 0;
+      }
+
+      if(event.body) {
+        var offset = event.body.length;
+        var index = event.index - offset;
+
+        while(offset >= 0) {
+
+          var isig = struct.extractShort(event.descriptor, offset * 2);
```
pazaan: `isig` doesn't seem to really be the isig, but a composite structure. It would be great if we could have a comment that links to where the structure is defined.
My rationale for this is that as a new developer, I can see that the code is doing something, but I have no way of validating if it's doing it correctly.

lib/medtronic/processData.js
```diff
@@ -426,6 +464,84 @@ function buildBGRecords(records) {
...
+              .with_deviceTime(sundial.formatDeviceTime(event.jsDate))
+              .with_units('mg/dL')      // TODO: hard-coded in mg/dL on device?
+              .set('index', event.index);
+        cfg.cgmTzoUtil.fillInUTCInfo(record, event.jsDate);
+        record = record.done();
+        postrecords.push(record);
+      }
+    }
+
+  };
```
pazaan: Unnecessary semicolon.

lib/medtronic/processData.js
```diff
@@ -426,6 +464,84 @@ function buildBGRecords(records) {
...
+        }
+
+      }
+    } else {
+      console.log('Skipping', event);
+    }
+
+    if(event.type.value === CBG_RECORD_TYPES.SENSOR_CAL_BG.value) {
+      if(struct.extractShort(event.descriptor, event.descriptor.length - 12) === 0x22) {
+        var record = cfg.builder.makeDeviceEventCalibration();
```
pazaan: Re-declaration of `record`.

lib/medtronic/processData.js
```diff
@@ -426,6 +464,84 @@ function buildBGRecords(records) {
...
+
+            var date = sundial.applyOffset(start, -(recordsSinceTimestamp * 5));
+            recordsSinceTimestamp += 1;
+
+            console.log('CBG:', sundial.formatDeviceTime(date), common.bytes2hex(struct.extractBytes(event.descriptor, offset * 2, 2)),',',common.bytes2hex([struct.extractByte(event.body,offset)]),',',(isig & 0x01) + (event.body[offset] << 1),',',(isig  & 0xFFFE) / 100.0);
+            if(cbg >= 40) { // ignore magic numbers
+              var record = cfg.builder.makeCBG()
+                    .with_value(cbg)
+                    .with_deviceTime(sundial.formatDeviceTime(date))
+                    .with_units('mg/dL')      // TODO: check if mmol/L cbg is also hard-coded in mg/dL
```
pazaan: Make `mg/dL` and `mmol/L` consts?

lib/medtronic/processData.js
gniezen: Yes, eventually. At the moment there are lots of `console.log`s just to make things easier to debug during alpha testing. I plan on removing most of them before going to prod, or change them to only show up if the app is built in debug mode.

lib/medtronic/processData.js
gniezen: Agreed, I'm renaming `isig` to `descriptor`. The structure doesn't really lend itself to being defined in a packet diagram, so I'm adding a description of the structure as a comment.

lib/medtronic/processData.js
gniezen: Ah, these are one of the log statements I'm planning on removing after alpha testing.

lib/medtronic/processData.js
gniezen: Not a bad idea, but then we'll have to go back and do it for all the other drivers too? ;) I guess in this specific instance I prefer consistency to prevent confusion, and keep in mind that the data services API will validate the units on ingestion.

lib/medtronic/processData.js
gniezen: Ooh, good catch :)

lib/drivers/docs/medtronicCGM.md
gniezen: Aargh, sorry, you're right - I'm getting confused. I will change it to `[?]`, as the only way to test would be to build some kind of walkabout box.

PR #358, opened 2017-01-03
Handle no Medtronic CGM records on pump

gniezen:
Discovered a bug while trying to upload a binary blob of Veo data where there are no available CGM records on pump. Pinging @pazaan for review.

UPDATE: also fixes a truthiness bug when temp basals have a zero rate.

### DIFF
--- a/lib/medtronic/processData.js
+++ b/lib/medtronic/processData.js
@@ -25,7 +25,7 @@ var common = require('../commonFunctions');
 var annotate = require('../eventAnnotations');
 var TZOUtil = require('../TimezoneOffsetUtil');
 
-var debug = (typeof __DEBUG__ === 'undefined') ? false : __DEBUG__;
+var debug = isBrowser ? require('../bows')('MedtronicDriver') : console.log;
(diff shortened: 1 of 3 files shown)


gniezen:
[Medtronic CGM data [uploader] [dev: 5 tst: 2 ux: 0]](https://trello.com/c/vMtU8xkK/19-medtronic-cgm-data-uploader-dev-5-tst-2-ux-0)

lib/medtronic/processData.js
```diff
@@ -25,7 +25,7 @@ var common = require('../commonFunctions');
 var annotate = require('../eventAnnotations');
 var TZOUtil = require('../TimezoneOffsetUtil');
 
-var debug = (typeof __DEBUG__ === 'undefined') ? false : __DEBUG__;
+var debug = isBrowser ? require('../bows')('MedtronicDriver') : console.log;
```
pazaan: `isBrowser` isn't defined until the line below, so it will always be false here.

gniezen:
@pazaan After you've merged the branch you can delete it (to keep things tidy).

PR #359, opened 2017-01-04
Bugfix: Now correctly handles cancelled temp basals running over midnight

gniezen:
Even if a temp basal is cancelled, we still need to figure out if there were any scheduled basal changes while the temp basal was running to make sure the suppressed info is correct. 

The issue here was that if a temp basal runs over midnight, we should make sure that we don't look for schedule changes after midnight if the temp basal was already cancelled before midnight.

Why is this so tricky? Well, schedule changes are stated in "milliseconds since midnight", which is why there are edge cases if a temp basal runs over midnight.

I added unit tests for:
- a temp basal that should run over midnight but is cancelled before a schedule change at midnight
- a temp basal runs over midnight and is cancelled after a schedule change at midnight


gniezen:
[[uploader] Medtronic basals](https://trello.com/c/zlAnS0iw/13-uploader-medtronic-basals)

PR #360, opened 2017-01-11
[WIP] 600-series Proof of Concept

pazaan:
Code that can read Basal patterns from a Medtronic 600-series pump and upload them to the `dataservices` API.

@jebeck and/or @darinkrauss - Can I please get a general code review? I'll get @gniezen to check out the comms parts later.


lib/redux/reducers/devices.js
```diff
@@ -25,6 +25,21 @@ const devices = {
     source: {type: 'device', driverId: 'Medtronic'},
     enabled: {mac: true, win: true}
   },
+  medtronic600: {
+    instructions: 'Connect your Bayer Contour Next Link 2.4 to your computer',
```
darinkrauss: Might need to change wording here. I have no idea how to determine if my Next Link is "2.4" or not. (I assume that is a version number, but it isn't clear.)

lib/redux/reducers/devices.js
```diff
@@ -25,6 +25,21 @@ const devices = {
...
+    image: {
+      'src': '/images/bayer_contour_next_link_24.jpg',
+      'height': 52,
+      'width': 143,
+      'alt': 'Bayer Contour Next Link 2.4'
+    },
+    key: 'medtronic600',
+    name: 'Medtronic - 630g, 640g or 670g',
```
darinkrauss: From my quick look at their website, it is a captial "G", not lowercase "g".

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+              return callback(true);
+            }
+            return callback(false);
+          });
+        },
+        (valid) => {
+          if (valid instanceof Error) {
+            throw valid;
+          }
+          return (valid !== true);
```
darinkrauss: Why are we comparing against `true` here? Shouldn't it just be 'return !value;'

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+  }
+
+  sendMessage(hidDevice) {
+    // eslint-disable-next-line no-unused-vars
+    return new Promise((resolve, reject) => {
+      let pos = 0;
+      const message = this.payload;
+
+      async.doWhilst(
+        (callback) => {
```
darinkrauss: Looking at our existing code, we include the 'function' keyword here. Do we want to continue that for this new code?

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+  // Override send(), because we do an 'optional double read' after sending the request
+  send(hidDevice, readTimeout = BCNLMessage.READ_TIMEOUT_MS, sendDelay = 0, get80response = true) {
+    // eslint-disable-next-line no-unused-vars
+    return new Promise((resolve, reject) => {
+      // Timeout might be zero, but the callback will fire anyway
+      this.sendMessage(hidDevice)
+        .then(() => BCNLMessage.wait(sendDelay))
+        .then(() => this.readMessage(hidDevice, readTimeout))
+        .then((response) => {
+          // TODO - throw here if the 0x81 doesn't check out.
```
darinkrauss: Obviously need to deal with TODOs at some point.

darinkrauss:
Just a couple of general comments, but nothing specific to the actual implementation or comms.

pazaan:
Closing review so that it doesn't accidentally get merged.

lib/drivers/medtronic600Driver.js
pazaan: Yes, TODO's will be dealt with as part of the "production readiness" Tello card :)

lib/drivers/medtronic600Driver.js
pazaan: Good question. I know that using ES6 syntax is a departure from the existing code in Uploader, but I've been careful to make sure that I'm only using language features that work with the currently configured Babel version. I figure that we'd like to move towards ES6 syntax (as seen in viz/blip), but I'm happy to get @jebeck thoughts here.

lib/redux/reducers/devices.js
pazaan: Good point. It was lower case in certain parts of the website, but all the official documentation shows it in upper case. Will fix.

lib/redux/reducers/devices.js
pazaan: `Contour Next Link 2.4` is the model: https://www.contournext.com/our-products/contour-next-products/contour-next-link-24-meter/

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+    this.linkMAC = null;
+    this.pumpMAC = null;
+    this.key = null;
+  }
+
+  get serialNumber() {
+    return this.modelAndSerial.replace(/\d+-/, '');
+  }
+
+  get iv() {
```
pazaan: IV is a standard abbreviation in cryptography, in a similar way that we commonly use cgm, sg, bg, etc.
See https://en.wikipedia.org/wiki/Initialization_vector and https://nodejs.org/api/crypto.html#crypto_crypto_createcipheriv_algorithm_key_iv

lib/drivers/medtronic600Driver.js
jebeck: I was almost going to comment on the ES6 syntax as well - mostly it was the use of classes that caught my eye. In general I definitely am supportive of using ES6 in new code (though we have a policy in the front-end stuff of not mixing code - so if you're editing an old ES5 file don't sprinkle a bit of ES6 syntax in there even if you can...it's jarring & confusing to read like that). But in this case I feel like Gerrit's opinion should carry a lot of weight since you and him are going to be the ones going back and forth between ES5 and ES6 driver code, so maybe a bit of a discussion to be had when he gets back. (I suspect he won't be *violently* opposed, but that's about as far as I'm comfortable guessing at his opinion.)

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+    },
+
+    uploadData(progress, data, cb) {
+      progress(0);
+
+      const sessionInfo = {
+        deviceTags: ['insulin-pump'],
+        deviceManufacturers: ['Medtronic'],
+        deviceModel: '640g',
+        deviceSerialNumber: '1055866',
```
pazaan: Yes, this is just the proof of concept code. This will be removed in the next round of changes.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+              callback(err);
+            });
+        },
+        (result) => {
+          if (result instanceof Error) {
+            throw result;
+          } else {
+            if (result === null) {
+              this.pumpSession.radioChannel = 0;
+              reject(new Error('Cannot connect to pump. Are you nearby?'));
```
pazaan: You're right. The wording here is terrible. Will fix :) Errors are surfacing in the UI, but only in the "Error details" section. I'll look to make user-actionable messages like this more obvious.

package.json
```diff
@@ -82,7 +82,7 @@
     "sinon": "1.17.3",
     "sinon-chai": "2.8.0",
     "style-loader": "0.13.1",
-    "webpack": "1.12.14",
+    "webpack": "1.14.0",
```
pazaan: I have two separate Vagrant boxes - one for the 600-series code, and the other for making hotfix changes and builds, so it doesn't make too big of a problem for me. That said, the change here was non-destructive as far as I can tell, so I wouldn't be opposed to cherry-picking this into `master`.

lib/drivers/medtronic600Driver.js
pazaan: I was just being really explicit (using the type check as well), but since the `instanceof Error` catches the only other case, I think `!valid` is fine. @jebeck, is there a preference for this?

lib/drivers/medtronic600Driver.js
jebeck: Ok, sounds fine then.

lib/drivers/medtronic600Driver.js
jebeck: I'd say try to get a feel for the "standard" in the uploader code and follow that? I know, for example, that cheddar was rather fond of `x != undefined` as a checked for both `null` and `undefined` (rather than `x !== undefined` which only fails for `undefined`).

Can't remember off the top of my head what's most common in the upload for what you were doing here except that the hyper-explicit `valid !== true` is *not* it... 😉

package.json
gniezen: I second getting this into master asap. @jebeck - what kind of testing would be needed to ensure this doesn't cause any regressions?

lib/redux/reducers/devices.js
gniezen: I agree with Lennart that the only way to distinguish this meter from the other BCN Link is the "2.4" model number. What I do think will be confusing is selecting it from the "Choose Devices" screen, where it shows up as 
screen shot 2017-01-16 at 13 31 07

I'll refer this to #design on Slack.

lib/drivers/medtronic600Driver.js
gniezen: I'm not opposed to using ES6 code, but it does currently break our CLI tools. While developing the Tandem driver I was itching to use ES6 features (like spread syntax) but decided against it keep thing simple: only using ES6+ components in the React code, and keeping all drivers ES5-only.

It's probably a good idea to schedule a short discussion to talk through the pro's and cons to reconsider switching to using ES6 in driver code - maybe as part of all hands this week?

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1221 @@
...
+const async = require('async');
+const sundial = require('sundial');
+const crypto = require('crypto');
+const crcCalculator = require('../crc.js');
+
+const isBrowser = typeof window !== 'undefined';
+// eslint-disable-next-line no-console
+const debug = isBrowser ? require('../bows')('Medtronic600Driver') : console.log;
+
+class ExtendableError extends Error {
```
gniezen: What is the purpose of this class? Are you trying to reverse the effects of the stack trace being stringified and passed up to the UI in e.g. https://github.com/tidepool-org/chrome-uploader/blob/master/lib/redux/actions/utils.js#L78-L86?

lib/drivers/medtronic600Driver.js
jebeck: Ah, I forgot about the problem with the CLI tools! Sorry for missing that...

package.json
jebeck: I would say just request some general "smoke testing" of the uploader functionality? (Click around the UI, verify that a couple of devices upload.) It doesn't feel high-risk to me. If the apps builds (which it does), everything's probably fine.

lib/drivers/medtronic600Driver.js
pazaan: It's so that when an error is thrown, the stack trace shows the correct subclass name (otherwise it just shows `Error`). See http://stackoverflow.com/questions/31089801/extending-error-in-javascript-with-es6-syntax

PR #363, opened 2017-01-18
[WIP] Medtronic 600-series proof of concept

pazaan:
@gniezen are you able to have a look through and make sure that this looks okay?


lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+      progress(0);
+
+      const sessionInfo = {
+        deviceTags: ['insulin-pump'],
+        deviceManufacturers: ['Medtronic'],
+        deviceModel: data.settings.pumpModel,
+        deviceSerialNumber: data.settings.pumpSerial,
+        deviceId: `${data.settings.pumpModel}:${data.settings.pumpSerial}`,
+        start: sundial.utcDateString(),
+        timeProcessing: 'utc-bootstrapping',
```
gniezen: Just a note that this will need to change to `timeProcessing: cfg.tzoUtil.type` when you implement UTC bootstrapping.

gniezen:
[Medtronic 6-series, part 1: Proof of Concept [uploader] <ux:0 dev:5 tst:0>](https://trello.com/c/FY4NELSm/27-medtronic-6-series-part-1-proof-of-concept-uploader-ux-0-dev-5-tst-0)

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+  static buildPayload(commandType, pumpSession, payload) {
+    const payloadLength = (payload === null) ? 0 : payload.length;
+    const payloadBuffer = Buffer.alloc(MinimedMessage.ENVELOPE_SIZE + payloadLength);
+
+    /* eslint-disable lodash/prefer-lodash-method */
+    payloadBuffer.writeUInt16BE(MinimedMessage.MINIMED_HEADER, 0); // Q\x03
+    payloadBuffer.write('000000', 2); // Pump serial. '000000' for 600-series
+    payloadBuffer.fill(0, 8, 18); // Unknown bytes (hardcoded in CLPro)
+    payloadBuffer.writeUInt8(commandType, 18);
+    payloadBuffer.writeUInt32LE(pumpSession.envelopeSequenceNumber += 1, 19);
```
jebeck: My assumption is that we rolled our own originally because back in the node 0.10 days Buffer didn't include everything we needed.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+        deviceModel: data.settings.pumpModel,
+        deviceSerialNumber: data.settings.pumpSerial,
+        deviceId: `${data.settings.pumpModel}:${data.settings.pumpSerial}`,
+        start: sundial.utcDateString(),
+        timeProcessing: 'utc-bootstrapping',
+        tzName: cfg.timezone,
+        version: cfg.version,
+      };
+
+      const postRecords = [{
```
jebeck: I'd support continuing to use the object builder for the consistency in error messages it provides across drivers.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+      // We use sendMessage instead of super.send() because we want to pull raw
+      // data only, and we can't determine response types until after we've checked
+      // them. The CNL can send the messages in different orders.
+      this.sendMessage(hidDevice)
+        .then(() => BCNLMessage.wait(sendDelay))
+        .then(() => this.readMessage(hidDevice, readTimeout))
+        .then((response) => {
+          response1 = response;
+        })
+        .then(() => BCNLMessage.wait(sendDelay))
```
pazaan: I'm following the delays set out in the CLPro code.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+  get packedLinkMAC() {
+    return Buffer.from(this.linkMAC, 'hex').swap64().toString('binary');
+  }
+
+  get packedPumpMAC() {
+    return Buffer.from(this.pumpMAC, 'hex').swap64().toString('binary');
+  }
+
+  getHMAC() {
+    const paddingKey = 'A4BD6CED9A42602564F413123';
```
pazaan: Yes, it is 😄

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+  static get ENVELOPE_SIZE() {
+    return 33;
+  }
+
+  static get MINIMED_HEADER() {
+    return 0x5103; // Q\x03
+  }
+
+  static oneByteChecksum(buffer) {
+    // eslint-disable-next-line no-bitwise
```
pazaan: For this driver, I'm using the `.eslintrc` from `viz`, which is much stricter. My thought is that explicitly calling out `no-bitwise` is a good rule of thumb, even in files where we do a lot of bitwise operations, because it helps reduce any logic bugs.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+    this.pumpSession = pumpSession;
+  }
+
+  static buildPayload(commandType, pumpSession, payload) {
+    const payloadLength = (payload === null) ? 0 : payload.length;
+    const payloadBuffer = Buffer.alloc(MinimedMessage.ENVELOPE_SIZE + payloadLength);
+
+    /* eslint-disable lodash/prefer-lodash-method */
+    payloadBuffer.writeUInt16BE(MinimedMessage.MINIMED_HEADER, 0); // Q\x03
+    payloadBuffer.write('000000', 2); // Pump serial. '000000' for 600-series
```
pazaan: Yes, that's correct. The driver gets the serial number of the pump from the paired BCN Link. 
The pump can be paired with multiple BCNL's, but a BCNL can only be paired with one pump.
It's also possible to create an "ad hoc" pairing, but after discussion with Brandon, we're sticking with the "already paired" model.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+    /* eslint-enable lodash/prefer-lodash-method */
+
+    // Now that we have written the message, calculate the CRC
+    const checksum = MinimedMessage.oneByteChecksum(payloadBuffer);
+    payloadBuffer.writeUInt8(checksum, 32);
+    return payloadBuffer;
+  }
+}
+
+class TelDMessage extends MinimedMessage {
```
pazaan: From the CLPro code. The newer protocol messages are identified as `TelD`, and the encrypted sections of the message as `ComD`.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+
+class JoinNetworkRequest extends TelDRequest {
+  static get READ_TIMEOUT_MS() {
+    return 10000;
+  }
+
+  constructor(pumpSession) {
+    const payloadBuffer = Buffer.alloc(26);
+
+    /* eslint-disable lodash/prefer-lodash-method */
```
pazaan: Lodash is being used elsewhere in the file. This eslint rule is asking "Why are you using the object's `fill` method, when you could use `lodash`?". I'm using the object's fill method because it looks more consistent in this block of code.

lib/drivers/medtronic600Driver.js
pazaan: Yes, I am. I just did it this way for the PoC.

lib/drivers/medtronic600Driver.js
jebeck: You might not want to carry over the lodash part of the viz .eslintrc. We made a decision to prefer lodash everywhere for consistency, but if you don't want to use lodash as the default in the uploader, it doesn't make sense to have that linting extension.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+        });
+    });
+  }
+}
+
+module.exports = (config) => {
+  const cfg = _.clone(config);
+  const driver = new MM600SeriesDriver(cfg.deviceComms);
+
+  // REVIEW - discuss http://eslint.org/docs/rules/no-param-reassign...
```
pazaan: It's not worth trying to fix existing drivers, but I'm wondering if cloning objects is a good best practice in this and any new drivers. Thoughts @jebeck?

jebeck:
I think you and @gniezen should work out the details of linting prefs driver-side. no-param-reassign makes a lot of sense in React code but we also almost never need to work around it because it just enforces largely functional patterns we were already using.

lib/drivers/medtronic600Driver.js
pazaan: I added this to my latest changes.

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+    const minimedPayloadSize = this.payload.readUInt16LE(0x1C);
+    // Check the payload's checksums
+    const minimedPayload =
+      Buffer.from(this.payload.slice(0x00, MinimedMessage.ENVELOPE_SIZE + minimedPayloadSize));
+    const expectedChecksum = minimedPayload[0x20];
+    const calculatedChecksum =
+      // eslint-disable-next-line no-bitwise
+      (MinimedMessage.oneByteChecksum(minimedPayload) - expectedChecksum) & 0xFF;
+
+    if (expectedChecksum !== calculatedChecksum) {
```
pazaan: Good catch! I wasn't handling the different 600-series "comms modes" and their respective tear-downs correctly. Fixed in my history message update.

lib/drivers/medtronic600Driver.js
gniezen: Interesting. I wonder if the same approach is used with the 5/7-series devices?

lib/drivers/medtronic600Driver.js
gniezen: Hmm, should we maybe consider renaming it to something less obtuse?

pazaan: Changed to `NGP`

lib/drivers/medtronic600Driver.js
```diff
@@ -0,0 +1,1418 @@
...
+
+    for (let i = 0; i < numItems; i++) {
+      const highInteger = this.decryptedPayload.readUInt16BE(0x06 + (i * 9));
+      const highDecimal = this.decryptedPayload.readUInt16BE(0x08 + (i * 9));
+      const lowInteger = this.decryptedPayload.readUInt16BE(0x0A + (i * 9));
+      const lowDecimal = this.decryptedPayload.readUInt16BE(0x0C + (i * 9));
+
+      targets.push({
+        start: this.decryptedPayload[0x0E + (i * 9)] * 30 * sundial.MIN_TO_MSEC,
+        high: parseFloat(`${highInteger}.${highDecimal}`),
```
pazaan: Debugged and fixed over Slack.

lib/drivers/medtronic600Driver.js
pazaan: Removed the lodash part of the .eslintrc.

pazaan:
[Medtronic 6-series, part 1: Proof of Concept [uploader] <ux:0 dev:5 tst:0>](https://trello.com/c/FY4NELSm/7-medtronic-6-series-part-1-proof-of-concept-uploader-ux-0-dev-5-tst-0)

PR #365, opened 2017-01-26
Medtronic initial settings

gniezen:
While we do record all pump settings changes, this PR adds the initial pump settings at the first data point. We have a record of both the old and the new setting at a given time, so we build up the settings history by starting from the current settings and working our way back through the changes. This means that the first "old" settings will be the initial settings.


gniezen:
[[uploader] [dev: 3 tst: 1 ux: 0] Medtronic device settings](https://trello.com/c/qsKqeuFh/18-uploader-dev-3-tst-1-ux-0-medtronic-device-settings)

PR #369, opened 2017-02-02
Medtronic Direct bugs uncovered with Darin's data

gniezen:
Darin has noticed gaps in his data, which is due to a communication error that failed silently. During the comms error an empty packet is returned, where the checksum doesn't fail because it's also zero. This PR correctly fails the checksum check if the packet length is 0 and automatically retries. I confirmed from @darinkrauss 's console logs that it works.

This PR also fixes another issue uncovered with his data, where a suspended basal crossed two schedule changes and extended past midnight, which is now taken into account during schedule change checks.


test/node/medtronic/testMedtronicSimulator.js
```diff
@@ -933,6 +933,128 @@ describe('medtronicSimulator.js', function() {
...
+          timezoneOffset: 0,
+          conversionOffset: 0,
+          activeSchedule: 'standard',
+          units: { 'bg': 'mg/dL' },
+          basalSchedules: {
+            standard: [
+              {
+    						'start': 0,
+    						'rate': 1
+    					},
```
pazaan: Weird formatting/spacing.

PR #371, opened 2017-02-09
OmniPod: drop unbootstrappable alarms that fail UTC lookup

jebeck:
As I think you'll see pretty quickly @gniezen the changes here ended up being more (i.e., completely) isolated to the OmniPod driver than I(/we?) initially thought. The issue is that since the `fillInUTCInfo` method *mutates* objects (and *also* returns them, but that's mostly for testing purposes), it's not up to the TZOUtil to "drop" things on the floor. The driver has to do the dropping, so I had to update the OmniPod driver code to look for the incomplete (i.e., non-mutated) resulting objects (in this case: alarms) after the call to `fillInUTCInfo`.

There's a small risk here that there's some other case (i.e., a non-alarm event in OmniPod or an event in another device) where lookup will fail, but in that case the `.done()` step of the object builder will throw the error for us when there's no `time` field on the object, so we'll still get a surfaced error in those cases (and then we can figure out what the cause is and potentially drop them from the upload in the same way).

Let me know if this makes sense, happy to chat about it if not.

### DIFF
--- a/lib/drivers/insuletDriver.js
+++ b/lib/drivers/insuletDriver.js
@@ -788,12 +788,19 @@ module.exports = function (config) {
       }
       // handle non-history alarms
       else {
-        // alarm.alarm.alarm is not a typo!
-        if (alarm.alarm.alarm_text != null) {
-          alarmValue = alarm.alarm.alarm;
+        // will occur for non-history alarms that aren't bootstrappable
+        // since we're no longer erroring on failure to look up UTC info
+        if (postalarm.time == '**REQUIRED**') {
+          postalarm = null;
         }
         else {
-          postalarm = null;
+          // alarm.alarm.alarm is not a typo!
+          if (alarm.alarm.alarm_text != null) {
+            alarmValue = alarm.alarm.alarm;
+          }
+          else {
+            postalarm = null;
+          }
         }
       }
       var alarmText = getNameForValue(ALARM_TYPES, alarmValue);
--- a/lib/drivers/insuletDriver.js
+++ b/lib/drivers/insuletDriver.js
@@ -790,17 +790,15 @@ module.exports = function (config) {
       else {
         // will occur for non-history alarms that aren't bootstrappable
         // since we're no longer erroring on failure to look up UTC info
-        if (postalarm.time == '**REQUIRED**') {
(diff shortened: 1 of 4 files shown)


jebeck:
[<0:ux 2:dev 1:tst> [bug] [uploader] Un-bootstrappable alarm causes OmniPod upload to fail (under rare circumstances)](https://trello.com/c/ra5u5jWP/85-0-ux-2-dev-1-tst-bug-uploader-un-bootstrappable-alarm-causes-omnipod-upload-to-fail-under-rare-circumstances)

lib/drivers/insuletDriver.js
```diff
@@ -788,12 +788,19 @@ module.exports = function (config) {
       }
       // handle non-history alarms
       else {
-        // alarm.alarm.alarm is not a typo!
-        if (alarm.alarm.alarm_text != null) {
-          alarmValue = alarm.alarm.alarm;
+        // will occur for non-history alarms that aren't bootstrappable
+        // since we're no longer erroring on failure to look up UTC info
+        if (postalarm.time == '**REQUIRED**') {
```
gniezen: Very nitpicky, but `===` instead of `==` ?

lib/drivers/insuletDriver.js
```diff
@@ -788,12 +788,19 @@ module.exports = function (config) {
...
         }
         else {
-          postalarm = null;
+          // alarm.alarm.alarm is not a typo!
+          if (alarm.alarm.alarm_text != null) {
+            alarmValue = alarm.alarm.alarm;
+          }
+          else {
+            postalarm = null;
+          }
```
gniezen: Uhm, would it maybe improve the readability here if we use
```
if {}
else if {}
else {}
```
instead of the nested ifs and elses?

lib/drivers/insuletDriver.js
jebeck: Good catch. Fixed.

lib/drivers/insuletDriver.js
jebeck: Yup, done.

jebeck:
I'll update the test strategy. I think you won't be at stand-up today, so thoughts about merge sequencing? I'll touch base with @pazaan too.

PR #373, opened 2017-02-15
Misc. fixes for Medtronic Direct

gniezen:
- Sometimes pump returns an empty page with zero header - we now catch this.
- Sometimes there is a CGM reading without SENSOR_TIMESTAMP packet
- Sometimes there is an invalid (stale?) entry in the carb ratios for pump setting changes
- Handles case where temp basal is cancelled at the exact same time a new basal schedule starts


gniezen:

lib/medtronic/processData.js
```diff
@@ -663,6 +663,10 @@ function getCarbRatios(encoded, units) {
       break;
     }
     var startTime = carbRatio.offset * 30 * sundial.MIN_TO_MSEC;
+    if((carbRatios.length > 0) && (startTime < carbRatios[carbRatios.length-1].start)) {
+      debug('Invalid start time for carb ratio:');
```
pazaan: Was there extra info you wanted to put here after the `:`?

lib/medtronic/processData.js
```diff
@@ -472,7 +472,7 @@ function buildBGRecords(records) {
 function buildCGMRecords(events) {
 
   var postrecords = [];
-  var start = null;
+  var start = events[events.length-1].jsDate;
```
pazaan: Are all events guaranteed to have a `jsDate`?

lib/medtronic/processData.js
gniezen: Ah, good catch!

lib/medtronic/processData.js
gniezen: For CGM events, yes. And in this specific instance, putting in a check wouldn't really help much, as `start` will then either be `null` or `undefined`.

lib/medtronic/medtronicSimulator.js
```diff
@@ -103,7 +103,7 @@ exports.make = function(config){
               break;
             }
 
-            if (common.computeMillisInCurrentDay(cancelled) < currSchedule[i].start ) {
+            if (common.computeMillisInCurrentDay(cancelled) <= currSchedule[i].start ) {
```
gniezen: @pazaan I'm adding one more change to this PR. What happened here is that if a temp basal is cancelled at the exact same time that a new basal schedule is starting, this if statement would not be triggered,  hence the adding of `=`. I'm merging already, but ping me if there any concerns.

PR #375, opened 2017-02-21
Handle temp basal being suspended and not resumed before upload

gniezen:
@darinkrauss noticed during deduplicator testing that if a temp basal is suspended, and not resumed before upload, additional incorrect events will be added to the data.

This PR contains a fix and relevant tests.


gniezen:

PR #378, opened 2017-03-02
MedT Direct bugfix: Comms failures

gniezen:
Eric has been experiencing numerous comms failures, which we suspect is due to low battery levels, but will occur in production if we don't address them now.

This bugfix:
- verifies the number of parts received per page
- will read past zeroes to see if there is more data 
- adds a page checksum verification (CRC16-CCITT of data pages) in addition to the packet checksum verification (simple LSB sum).
- will retry if failing to open connection
- trims null characters when checking model number


gniezen:
[<0:ux 5:dev 2:tst> [platform] Implement Jellyfish de-duplication algorithm in Platform for Medtronic support](https://trello.com/c/GExalbvy/50-0-ux-5-dev-2-tst-platform-implement-jellyfish-de-duplication-algorithm-in-platform-for-medtronic-support)

lib/drivers/medtronicDriver.js
```diff
@@ -481,19 +486,25 @@ module.exports = function (config) {
...
-                      payload[32] = 0x00; // mask out checksum itself
-                      var checksum = result.bytes[32];
-                      debug('Calculated checksum:', _sum_lsb(payload));
-                      debug('Checksum from packet:', checksum);
-                      if( (_sum_lsb(payload) !== checksum) || (decoded.message.length === 0)) {
-                        debug('Invalid checksum or null data');
-                        return callback(new Error(COMM_ERROR_TEXT));
-                      };
+
+                      if(decoded.message.length === 2 && decoded.message[0] === ASCII_CONTROL.NAK) {
```
pazaan: `decoded` will have a `message` field if it's an `instanceof Error` (as has been added above), but the behaviour may not be what you think...

lib/drivers/medtronicDriver.js
```diff
@@ -481,19 +486,25 @@ module.exports = function (config) {
...
+                        console.log('nak received');
+                        nak = true;
+                      } else {
+                        page.set(decoded.message, length);
```
pazaan: If `decoded` is an `instanceof Error`, `decoded.message` will be a string.

lib/drivers/medtronicDriver.js
```diff
@@ -481,19 +486,25 @@ module.exports = function (config) {
 
                     } else {
                       var decoded = decodeMessage(packet.parser, result.bytes);
```
pazaan: If `decodeMessage` is changed to throw, try here, otherwise a type check will be required.

lib/drivers/medtronicDriver.js
```diff
@@ -419,7 +419,11 @@ module.exports = function (config) {
 
   function decodeMessage (parser, message) {
     var response = struct.unpack(message, 0, 'b', ['recordType']);
-    _.assign(response, parser(message));
+    var result = parser(message);
+    if(result instanceof Error) {
+      return result;
```
gniezen: OK, I've changed it to `throw` instead and wrapped the `decodeMessage` calls in `try/catch` blocks.

PR #379, opened 2017-03-02
Bug fix: Animas mmol/L 

gniezen:
Fixes an issue where an SMBG record from the linked meter would be converted twice if in mmol/L.

### DIFF
--- a/lib/animas/animasLogic.js
+++ b/lib/animas/animasLogic.js
@@ -15,8 +15,8 @@
  * == BSD2 LICENSE ==
  */
 
- Number.prototype.toFixedNumber = function(significant, truncating){
-   var pow = Math.pow(10,significant);
+ Number.prototype.toFixedNumber = function(decimalPlaces, truncating){
+   var pow = Math.pow(10, decimalPlaces);
    if(truncating) {
       return +( Math.trunc(this*pow) / pow );
    } else {
--- a/lib/animas/animasLogic.js
+++ b/lib/animas/animasLogic.js
@@ -15,9 +15,13 @@
  * == BSD2 LICENSE ==
  */
 
- Number.prototype.toFixedNumber = function(significant){
+ Number.prototype.toFixedNumber = function(significant, truncating){
    var pow = Math.pow(10,significant);
-   return +( Math.round(this*pow) / pow );
+   if(truncating) {
+      return +( Math.trunc(this*pow) / pow );
+   } else {
+     return +( Math.round(this*pow) / pow );
+   }
  };
 
  var roundAndNotNegative = function(number) {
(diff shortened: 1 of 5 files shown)


gniezen:
@darinkrauss I was thinking you may be the best person to review this as you already know the background of the issue. If not, let me know ASAP.

darinkrauss:
LGTM! :shipit:

gniezen:
[Animas mmol/L support [uploader][bug]](https://trello.com/c/Ry3Cz0eC/14-animas-mmol-l-support-uploader-bug)

gniezen:
Hi @darinkrauss and @jebeck ! I've added one more change to this PR to resolve an issue Howard discovered during testing where we can be up to 0.05 (in either direction) off when calculating the suggested amount for a wizard bolus. Whoever can attend to this first, please do and ping me so I can get a new dev build ready for testing asap.

lib/animas/animasSimulator.js
```diff
@@ -202,13 +202,14 @@ exports.make = function(config){
     smbg: function(event) {
 
       if(currSMBG != null && (event.subType === 'manual') && (currSMBG.subType === 'linked')) {
+        var linkedBG = currSMBG.value;
         if(event.units === 'mmol/L') {
           // linked value is always in mg/dL, so we have to convert to
           // mmol/L with one significant digit if we want to compare to
           // Animas mmol/L manual BG value
-          currSMBG.value = (currSMBG.value/18.01559).toFixedNumber(1);
+          linkedBG = (currSMBG.value/18.01559).toFixedNumber(1);
```
jebeck: Could you factor out the conversion as a constant? In viz we've even tried to make sure we only define it once in the repo. Is it being used much or anywhere else currently? (My guess is maybe not as the conversion in most cases is happening back-end...)

lib/animas/animasLogic.js
```diff
@@ -15,9 +15,13 @@
  * == BSD2 LICENSE ==
  */
 
- Number.prototype.toFixedNumber = function(significant){
+ Number.prototype.toFixedNumber = function(significant, truncating){
```
jebeck: I guess there's no tests around the stuff in this file at present, so I don't think it's an absolute need. But I do think tests would help make the subtleties of what you're doing here easier to grok!

Related: maybe a different name for the `significant` parameter because unless I'm *not* understanding (possible!) it's not significant digits but rather number of decimal places, so maybe just `decimals` or similar? This is also not a distinction I would have remembered had I not looked it up recently (when the distinction between `toFixed` and `toPrecision` was relevant to me).

lib/animas/animasSimulator.js
gniezen: The only other place is in a local function in the CareLink code: https://github.com/tidepool-org/chrome-uploader/blob/master/lib/carelink/common.js#L134. Since we'll be deprecating that eventually I'll put the function in a common place, but not touch the CareLink driver.

lib/animas/animasLogic.js
gniezen: OK, changing it to `decimalPlaces`.

jebeck:
:shipit:

PR #386, opened 2017-03-16
Medtronic Direct: Support basal profile with many rates

gniezen:
This PR adds support for a Medtronic basal schedule with many (30+) rates.

### DIFF
--- a/lib/drivers/medtronic/medtronicDriver.js
+++ b/lib/drivers/medtronic/medtronicDriver.js
@@ -161,7 +161,13 @@ module.exports = function (config) {
       }
       ctr += struct.pack(bytes, ctr, '7bi', 0, 0, 0, partsPerPage, 0x10, 0x10, 0x00, payload.length+secondPacketLength);
     } else {
-      ctr += struct.pack(bytes, ctr, '7bi', 0, 0, 0, 0, 0, 0, 0, payload.length);
+      // for current basal profiles we are not sending parameters, but we're requesting
+      // three records instead of the usual one
+      if(partsPerPage === 3) {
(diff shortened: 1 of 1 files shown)


gniezen:

lib/drivers/medtronic/medtronicDriver.js
```diff
@@ -161,7 +161,13 @@ module.exports = function (config) {
       }
       ctr += struct.pack(bytes, ctr, '7bi', 0, 0, 0, partsPerPage, 0x10, 0x10, 0x00, payload.length+secondPacketLength);
     } else {
-      ctr += struct.pack(bytes, ctr, '7bi', 0, 0, 0, 0, 0, 0, 0, payload.length);
```
pazaan: This could be changed using new insight into the meanings of these bytes to make the code easier to understand (especially from a sense of purpose), but it's not critical to it working.

lib/drivers/medtronic/medtronicDriver.js
```diff
@@ -576,7 +611,14 @@ module.exports = function (config) {
 
           console.log('Raw packet received:', common.bytes2hex(packet));
 
-          message = extractPacketIntoMessage(packet.slice(MAGIC_HEADER.length));
+          var discard = MAGIC_HEADER.length;
+          if(replyNr > 0 && firstPacket) {
```
gniezen: IIRC only during the `open_connection` phase.

lib/drivers/medtronic/medtronicDriver.js
gniezen: I agree. However, I can't get the number of bytes in #388 to match up with the actual packet, so I'm going to defer this to a new PR after we've had a chance to discuss.

PR #387, opened 2017-03-17
Trying to fix requested page numbers for CGM pages

gniezen:
It seems that for some instances of CGM data, page 0 does not exist anymore, and you have to read from e.g. page 30-36 if there are 6 pages of data. I think this is because the pump does not hold more than 32 pages of CGM data, but the page counter keeps increasing.  E.g., one user's pump has a page counter of 285.

Note: This is still WIP until I've had a chance to test it on some pumps.

Other notable changes in this PR:
- refactored `buildMedtronicPacket`
- reads battery level from pump
- uploads current pump settings at time of upload AND at the time when it changed


gniezen:

PR #390, opened 2017-03-23
Medtronic Direct checksum fix

gniezen:
*Problem*: In some instances, even requesting the same packets 5 times after getting a checksum mismatch, the pump still sends the exact same packets. 

*Solution*:
It will retry 5 times, after which it will assume the data from the pump itself is corrupted. It will then just return the data it already has. Since the pump sends the newest data first, this means it will have all current data back to the date the data got corrupted, and all older (possibly corrupted) data will be ignored. This seems to be what CareLink does as well.


gniezen:

gniezen:
😂

PR #391, opened 2017-03-28
MedT Direct: Fix wizard carb units and pump settings

gniezen:
Two changes in this PR:

- for the current pump settings, use the the time at which they were actually activated, instead of the current device time
- always convert carb input and ratio to grams, as the data model doesn't currently support different carb units (e.g. exchanges) for wizard records


gniezen:

PR #392, opened 2017-04-03
MedT Direct page checksum mismatches: Flush and retry at end

gniezen:
This PR marks a page as invalid when the page checksum fails, and retries at the end of the transfer. It also implements a flush routine so that we can get rid of any extra packets that get sent during the transfer which are invalid. That is, the page would be larger than the expected 1024 or 2048 bytes.


gniezen:

PR #394, opened 2017-04-06
MedT Direct: Throw away a corrupt CGM record if it's the last one in the stream

gniezen:
It's possible for the last CGM record on the pump to sometimes be corrupted. If so, we check if it is indeed the last record, in which case we throw it away and upload the rest of the records. If it's not the last record, we discard that corrupted page (and previous pages), and start reading CGM data from the next page.


gniezen:

PR #396, opened 2017-04-20
Bug fixes after latest round of alpha tests

gniezen:
This PR:

- updates list of record types to include all known ones
- checks NAK bytes and retries if it's non-zero
- flushes and clears message buffers on retry
- also retries with single record and tri-record requests (not only history records)
- handles edge case where temp basal extends past midnight with schedule change shortly after midnight (with unit test)
- handles edge case where temp basal is followed by two LGS alarms, and then temp basal resumes


gniezen:

PR #397, opened 2017-04-27
WIP: Electron

krystophv:


.babelrc
```diff
@@ -1,3 +1,32 @@
...
+        "transform-es2015-classes"
+      ],
+      "presets": ["react-hmre"]
+    },
+    "test": {
+      "plugins": [
+        ["resolver", { "resolveDirs": [ "app/node_modules" ]}],
+        ["webpack-loaders", { "config": "webpack.config.test.js", "verbose": false }],
+        "babel-plugin-rewire",
+        ["transform-define", {
```
krystophv: Since we're not running the tests through webpack first, we need to perform the work that the webpack-define plugin would have in the babel transpile.

app/actions/async.js
```diff
@@ -671,16 +665,26 @@ export function checkUploadTargetUserAndMaybeRedirect() {
...
 
 export function clickAddNewUser(){
   return (dispatch, getState) =>{
     dispatch(syncActions.setUploadTargetUser(null));
-    dispatch(syncActions.setPage(pages.CLINIC_USER_EDIT, undefined, {metric: {eventName: metrics.CLINIC_ADD}}));
+    dispatch(setPage(pages.CLINIC_USER_EDIT, undefined, {metric: {eventName: metrics.CLINIC_ADD}}));
+  };
+}
+
+export function setPage(page, actionSource = actionSources[actionTypes.SET_PAGE], metric) {
```
krystophv: The new setPage function. This is a thin wrapper around the `react-redux-router` push function which creates an action to navigate to a new page (pushes the path to the browser history object) which also tacks on our meta/metric information into the location change `state` property, from which it is retrieved in middleware for tracking.

app/index.js
```diff
@@ -0,0 +1,28 @@
...
+import configureStore from './store/configureStore';
+import './app.global.css';
+import '../styles/main.less';
+import { ipcRenderer } from 'electron';
+
+const store = configureStore();
+const history = syncHistoryWithStore(hashHistory, store);
+store.dispatch(push('/'));
+
+ipcRenderer.on('action', function(event, action) {
```
krystophv: This is the communication mechanism for receiving actions dispatched from the `main` Electron process. `action` should always be the resulting object from an action creator.

app/utils/drivers.js
```diff
@@ -15,10 +15,11 @@
  * == BSD2 LICENSE ==
  */
 
-/* global __REDUX_DEV_UI__ */
+import os from 'os';
 
-if (__REDUX_DEV_UI__ === true) {
-  module.exports = require('./Root.dev');
+const platform = os.platform();
+if (platform === 'darwin'){
```
krystophv: The drivers detection is split like this because the two code paths depend on node modules which will only be installed/available on a particular OS and will error out if you try to require them on a non-supported OS.

app/utils/metrics.js
```diff
@@ -20,13 +20,19 @@ import _ from 'lodash';
...
 export function createMetricsTracker(api) {
-  return ({ getState }) => (next) => (action) => {
+  return () => (next) => (action) => {
     if (_.get(action, 'meta.metric', null) !== null) {
       api.metrics.track(
         _.get(action, 'meta.metric.eventName', NONE_PROVIDED),
         _.get(action, 'meta.metric.properties', {})
       );
     }
+		if (_.get(action, 'payload.state.meta.metric', null) !== null) {
```
krystophv: This is to extract meta information from the route change/page navigation events.

styles/components/LoggedInAs.module.less
```diff
@@ -75,13 +76,17 @@
...
+
 .dropdown {
   border: 1px solid @gray-light;
   cursor: default;
   position: absolute;
   background: white;
-  top: 45px;
-  left: 431px;
+  top: 42px;
+  left: ~"calc(100% - 164px)";
```
krystophv: `~` tells LESS not to process the next value. Since we're bundling Chromium, it's safe to use `calc()`. (actually, `calc()` has pretty solid support across browsers now anyway)

test/app/actions/async.test.js
```diff
@@ -121,9 +113,16 @@ describe('Asynchronous Actions', () => {
           meta: {source: actionSources[actionTypes.SET_NEW_PATIENT_URL]}
         },
         {
-          type: actionTypes.SET_PAGE,
-          payload: {page: pages.LOGIN},
-          meta: {source: actionSources[actionTypes.SET_PAGE]}
+          type: '@@router/CALL_HISTORY_METHOD',
```
krystophv: Almost all the `SET_PAGE` actions are now `@@router/CALL_HISTORY_METHOD`, which represents a page navigation.

test/lib/testLocalStore.js
```diff
@@ -49,16 +48,19 @@ describe('localStore [node.js version for testing]', function() {
   });
 
   describe('localStore.getItem', function() {
-    var data = {blocks: {mine: ['a', 'b'], yours: [1,2,3]}};
-    localStore = localStore(data);
+    var data = 'awesome_data';
```
krystophv: Fun fact: the `localStore` api specifies that values can only be strings.

gniezen:
@krystophv 21k lines added and 282 files touched, no biggie ;) Now that all my code is merged into this PR, I guess there's no way to know which parts you authored unless I went through it on a commit-by-commit basis? Then again, we didn't really touch the same files, so it shouldn't be too hard to remember.

krystophv:
@gniezen Yeah, and actually a _large_ portion of the lines added are from the OS-level USB drivers being added and the `yarn.lock` files (at least that's what I felt when I went through last to add comments).

test/lib/testLocalStore.js
gniezen: What are we still using `localStore` for? Is it even still necessary?

README.md
```diff
@@ -51,82 +65,155 @@ The environment variable `DEBUG_ERROR` (boolean) controls whether or not errors
...
+
+Docs reside in several places in this repository, such as `docs/` and `lib/drivers/docs`. They are built as a static site with [GitBook](https://www.gitbook.com/ 'GitBook') and served at [developer.tidepool.io/](http://developer.tidepool.io/) via [GitHub Pages](https://pages.github.com/ 'GitHub Pages').
+
+See [this guidance on our use of GitBook at Tidepool](http://developer.tidepool.io/docs/).
+
+## Publishing
+
+Release management and application updates are handled via the Github provider in the `electron-builder` project. The recommended workflow for a new production release is as follows:
+
+1. When you're working on what might become a new release, increment the version number in `package.json` and `app/package.json` and commit/push (on the branch)
```
krystophv: I don't believe `mversion` will update the `app/package.json` file automatically and, thus, won't be a sufficiently complete solution.

app/app.html
```diff
@@ -0,0 +1,32 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Tidepool Uploader</title>
+    <script>
+      (function() {
+        if (!process.env.HOT) {
+          const link = document.createElement('link');
+          link.rel = 'stylesheet';
+          link.href = './dist/style.css';
+          // HACK: Writing the script path should be done with webpack
```
krystophv: I'll need to check upstream with the boilerplate on this one (the comment was from the boilerplate).

test/lib/testLocalStore.js
krystophv: It holds the user's authentication token (used by `platform-client` I believe)

README.md
```diff
@@ -20,15 +23,26 @@ This README is focused on just the details of getting the uploader running local
...
-1. To run it, you can choose "Launch" from the chrome://extensions page. You can also run it from the Chrome App Launcher, which Chrome may install for you whether you want it or not.
-1. To open the JavaScript console/Chrome Dev Tools, click on the `index.html` link in the section of chrome://extensions devoted to the uploader. (Note: this link will only appear after you've launched the uploader.)
-1. React components and CSS will hot load after changes (this can be confirmed by watching the JavaScript console), but changes to device drivers and other code outside of the React components will not - use 'Reload' from chrome://extensions to reload after such changes. If the compilation/hot reload of a component fails for any reason (e.g. from a syntax error) you may need to reinitialize the hot loader by reloading the extension. You will definitely need to reload any time you change the manifest.
+1. Run these two commands __simultaneously__ in different console tabs.
```
krystophv: I haven't yet run them in two different consoles. This is one of the pieces that was inherited from the boilerplate. I'm fine just including the `dev` command.

app/main.development.js
```diff
@@ -0,0 +1,338 @@
...
+
+    const extensions = [
+      'REACT_DEVELOPER_TOOLS',
+      'REDUX_DEVTOOLS'
+    ];
+
+    const forceDownload = !!process.env.UPGRADE_EXTENSIONS;
+
+    // TODO: Use async interation statement.
+    //       Waiting on https://github.com/tc39/proposal-async-iteration
```
krystophv: This is another case of being a comment directly from the boilerplate. It's probably something to look into for the future, but may not be a v1 (async generators would be somewhat of a learning experience for me).

app/main.development.js
```diff
@@ -0,0 +1,338 @@
...
+    mainWindow.show();
+    mainWindow.focus();
+    autoUpdater.checkForUpdates();
+  });
+
+  mainWindow.webContents.on('new-window', function(event, url){
+    event.preventDefault();
+    open(url, 'chrome', function(error){
+      if(error){
+        // couldn't open in chrome specifically, probably because it's not installed
```
krystophv: This is actually significantly more complex than I originally thought. The `open` command here isn't sufficient to find and launch Chrome consistently across operating systems. I found that the Lighthouse project (https://github.com/GoogleChrome/lighthouse/tree/master/chrome-launcher) has isolated their `chrome-launcher` and we'll probably need to use that. I think it should be fairly easy to use their utility to determine if Chrome isn't installed and send that metric.

app/utils/drivers.win32.js
```diff
@@ -0,0 +1,61 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import _ from 'lodash';
+//import regedit from 'regedit';
```
krystophv: There hasn't been significant testing on either option, so I wasn't comfortable dropping the code yet. Once we get into some actual alpha testing and confidence in the currently used package/method is bolstered it certainly can be dropped. I'll put a TODO here.

lib/config.js
```diff
@@ -14,9 +14,10 @@
  * not, you can obtain one from Tidepool Project at tidepool.org.
  * == BSD2 LICENSE ==
  */
-var pkg = require('./../package.json');
+var pkg = require('../package.json');
 
-var config = (typeof window !== 'undefined') ? window.config : {};
+//var config = (typeof window !== 'undefined') ? window.config : {};
```
krystophv: I'm not entirely sure, but also, given that there's a full node available inside Electron, the two environments should be nearly identical. Is there usefulness in differentiating given the same execution environment?

test/app/actions/async.test.js
```diff
@@ -121,9 +113,16 @@ describe('Asynchronous Actions', () => {
...
+          payload: {
+            args: [ {
+							pathname: '/login',
```
krystophv: I had an issue with tabs vs spaces in my editor for a moment. Thought I'd found all the places it messed with them and fixed them. Apparently not quite.

lib/config.js
krystophv: Per below, I can see that there is some usefulness in differentiating outside pure functionality. We can use the `is-electron` npm package, which combines a pair of commonly used checks to see if the current runtime environment is electron or not.

test/lib/testLocalStore.js
gniezen: Hmm, and I have a feeling that authentication token is not being refreshed as expected, leading to uploads failing if you leave the Uploader open for a long time:
https://trello.com/c/yQx6JyQB. So we can revisit this when we finally get around to fixing that bug.

README.md
gniezen: Just checked and that is indeed the case. I asked on the #electron-builder Slack channel, and the response was
>  in case of two package.json structure, root package.json should be marked as private, version not required here.

So it turns out if you run `npm version` in the `app/` directory, it will just bump the package.json there, nice. If you want it to also create a tag, you just do a `mkdir .git` in `app/` as a workaround first.

Advantages of using `npm version` includes the capability to add preversion and postversion scripts, and of course preventing typos. Disadvantages include not being able to specify `alpha` (I think), but it does have `prerelease` tags (e.g. v0.310.0-0).

Do we want to use it, or do you think it's better to bump the version manually?

app/utils/drivers.win32.js
gniezen: 👍

lib/config.js
gniezen: Ah, yes, `is-electron` looks like a good solution.

README.md
krystophv: We could use it, but we also don't want to create tags locally and push them to GitHub in this flow (necessarily) - I've been using the GitHub releases mechanism to create the tag for me (the flow recommended by `electron-builder`), as actually making the release on GitHub will prevent new build artifacts from being pushed to that version (as matched in the `app/package.json` file it seems.) 

`npm version` can be passed any valid `semver` string, so we could use it for `alpha` versions or just use `mversion` inside the `app` directory (and without the `-m` flag, to prevent it from making a tag). 

It is slightly strange to only update the `app/package.json` as I almost always expect to find the current version of an application in the root `package.json` file. 

I don't have strong feelings about an approach so if `npm version` is preferable to you, we can work that in.

app/containers/App.js
```diff
@@ -0,0 +1,221 @@
...
+import VersionCheckError from '../components/VersionCheckError';
+import Footer from '../components/Footer';
+import Header from '../components/Header';
+import UpdateModal from '../components/UpdateModal';
+
+import styles from '../../styles/components/App.module.less';
+
+export class App extends Component {
+  static propTypes = {
+    // api: PropTypes.func.isRequired
```
krystophv: I'm actually going to re-work this because the `api` moved into the `route` prop, but is still absolutely necessary for everything else to work.

lib/core/storage.js
```diff
@@ -19,71 +19,23 @@ var _ = require('lodash');
  * Create our the store we will be using
  *
  * @param {Object}  options
- * @param {Boolean} options.isChromeApp
  * @param {Object}  options.ourStore the storage system we are using
```
krystophv: The `ourStore` is always localStorage in the uploader now, but as far as this wrapper goes, it does provide a little more than `localStorage` does. This relationship could be cleaned up/compressed, but I'm pretty sure that other moving pieces of the application (including `platform-client`) expect the API expressed by this wrapper to be available.

README.md
gniezen: OK, let's leave it as is for now. We can always revisit using `npm version` if it turns out doing it manually is more error-prone.

app/components/UpdateModal.js
```diff
@@ -0,0 +1,114 @@
...
+      text = 'After clicking Install, the uploader will restart to complete the installation.';
+      actions = [
+        <button key='dismiss' className={styles.buttonSecondary} onClick={sync.dismissUpdateAvailable}>
+          Dismiss
+        </button>,
+        <button key='install' className={styles.button} onClick={this.handleInstall}>
+          Install
+        </button>
+      ];
+    } else {
```
gniezen: Just tried out the fix and it works as expected.

lib/core/api.js
```diff
@@ -27,10 +25,9 @@ var semver = require('semver');
 var sundial = require('sundial');
 var uuid = require('node-uuid');
 
-var isChromeApp = (typeof chrome !== 'undefined');
 var bows = require('../bows');
-var errorText = require('../redux/constants/errors');
-var log = isChromeApp ? bows('Api') : console.log;
```
gniezen: Fixed.

lib/core/api.js
```diff
@@ -51,7 +48,7 @@ var api = {
 
 // synchronous!
 api.create = function(options) {
-  var tidepoolLog = isChromeApp ? bows('Tidepool') : console.log;
```
gniezen: Fixed.

gniezen:
Changes look good. I've approved, but of course with the caveat that the device driver changes still need to be reviewed. And the changes before we go to prod (update platform-client link and Circle CI link)

app/main.development.js
krystophv: I'm going to say that I wouldn't be comfortable letting this bit get to `prd` without using the `chrome-launcher` utility as I'm fairly certain that the open command won't properly find the Chrome executable across platforms.

app/routes.js
```diff
@@ -0,0 +1,23 @@
...
+import SettingsPage from './containers/SettingsPage';
+import ClinicUserSelectPage from './containers/ClinicUserSelectPage';
+import ClinicUserEditPage from './containers/ClinicUserEditPage';
+import NoUploadTargetsPage from './containers/NoUploadTargetsPage';
+import api from '../lib/core/api';
+
+export default (
+  <Route path="/" component={App} api={api}>
+    <IndexRoute component={Loading} />
+    <Route path="login" component={Login}/>
```
krystophv: I'm on the fence on this one. Currently, yes, they could be replaced by the paths in the pagesMap, but if we intended to make more complex paths in the future (regex-style pattern matching), then it's likely that those wouldn't be able to be flat references in the pagesMap. Perhaps that's premature optimization though and can be changed if we ever get there.

PR #398, opened 2017-05-12
Medtronic Direct: Negative duration for temp basals bug

gniezen:
When creating a new `deviceTime` for a fabricated temp basal based on the basal schedule (created for a cancelled temp basal where there is basal schedule change before the temp basal is cancelled), Chrome applies the local timezone when `deviceTime` is timezone-naive. This fix applies the timezone offset negatively before using the `new Date()` function, adds a duration in milliseconds, then adds the timezone offset again to revert back to timezone-naive.

We also make sure that we use `deviceTime` instead of `time` when checking if a temp basal was cancelled before midnight.


gniezen:
Yes, binary blobs from two alpha users and Brandon now upload successfully. The trick was to apply the timezone offset again. Brandon has also successfully uploaded directly from his pump. Eric is rescheduling with the two alpha users to retest their pumps directly too.

gniezen:

PR #400, opened 2017-05-17
Always use sundial when adding a duration to a timestamp

gniezen:
Fixes a bug discovered during alpha where a suspend/resume during a temp basal caused an incorrect `deviceTime`. When calculating the new `deviceTime` Chrome applies a timezone when it shouldn't. The fix is to always use sundial when adding durations to timestamps, as the Date constructor and `Date.parse` are unpredictable when parsing date/time strings.


gniezen:
@jebeck Looks like I didn't find all the instances where `Date.parse` could cause an issue with `deviceTime`. I've now checked that these are the only instances in the simulator where durations are added to time.

lib/drivers/medtronic/medtronicSimulator.js
```diff
@@ -83,6 +83,17 @@ exports.make = function(config){
     events.push(e);
   }
 
+  function addDurationToTime(time, timezoneOffset, duration) {
```
jebeck: I'd consider changing the function name and first param to `deviceTime` for clarity, but up to you...

lib/drivers/medtronic/medtronicSimulator.js
```diff
@@ -335,8 +346,8 @@ exports.make = function(config){
                   event.duration = currStatus.duration;
 
                   resumeBasal = _.clone(currBasal);
-                  resumeBasal.time = new Date(Date.parse(currStatus.time) + currStatus.duration).toISOString();
-                  resumeBasal.deviceTime = sundial.formatDeviceTime(Date.parse(currStatus.deviceTime) + currStatus.duration);
+                  resumeBasal.time = addDurationToTime(currStatus.time,currStatus.timezoneOffset, currStatus.duration).toISOString();
```
jebeck: I don't _think_ the function should be used here. The subtracting then adding back `timezoneOffset` gymnastics is specific to working with `deviceTime`s, and this one's `time`. You should be able to just add the `duration` here.

lib/drivers/medtronic/medtronicSimulator.js
gniezen: You know, I had it like that and then changed it to use with `.time` as well. You're right in that the gymnastics are only necessary for `.deviceTime`, but I'm so scared of using `Date.parse` now that I only use it where it definitely can't cause any problems (e.g. when subtracting `Date.parse` from `Date.parse` to get a duration).

Is there a reason to not use sundial for `.time`, apart from the extra gymnastics?

jebeck: I would personally find it confusing to use it where it's not needed because it's obfuscating and will likely confuse future devs (or even future you 😜 ), but no, there isn't a reason not to...it's just vacuous.

If I were explaining the problem you were having before, I wouldn't say `Date.parse` was to blame. It was using `deviceTime`s _as if_ they were `time`s. `Date.parse` is always safe on `time` because `time` is a complete timezone-aware timestamp (in this case UTC) and so `Date.parse` just turns it from String to integer. `deviceTime` is **part of** a timezone-aware timestamp too, but in this case the entire timezone-aware timestamp is represented in pieces with its timezone (offset) information  "carried" externally (in `timezoneOffset`), so `Date.parse` on _just_ the `deviceTime` doesn't do the complete operation needed. Does that help?

gniezen: OK, fair enough, I see your point 😄

gniezen:
While reviewing an alpha user's data, I discovered that `resumeBasal`'s duration is not always calculated correctly, so I removed that line in commit 27b0d7a as the duration will be calculated on the next pass of the simulator when it's the current basal.

gniezen:

gniezen:
Added some additional changes after testing with alpha user binary blobs.

PR #401, opened 2017-05-18
Medtronic Direct: Prevent async.retry from running 5 x 10 = 50 times

gniezen:
This PR fixes the instance where a consistent timeout can cause 50 retries from two interacting `async.retry`s.


gniezen:

PR #402, opened 2017-05-18
Medtronic Direct: Cleanup of console log messages

gniezen:
This PR cleans up the console messages used during development and testing, and puts the ones useful for debugging behind the `__DEBUG__` switch. To see them in the console, the environment variable `DEBUG_ERROR` should be `true`.

It also plays an "underwater" sound when the download completes (.wav file found in the [public domain](http://soundbible.com/908-Underwater.html) is the sounds of a gulf toadfish).


gniezen:

gniezen:
Yes, I would love to have a logger with different logging levels, that also has the capability to send those logs back to us automatically in case of a failed upload.

PR #403, opened 2017-05-19
Medtronic Direct: Various LGS fixes

gniezen:
Working through an alpha user's binary blob with many LGS suspends provided a number of edge cases for which this PR provides fixes:

- Added LGS type 0x01 (and confirmed in CLP)
- When the suspend state is 0x00, it's actually "normal pumping" (from CLP)
- improves matching between events to prevent incomplete tuples

### DIFF
--- a/lib/drivers/medtronic/cli/blob_loader.js
+++ b/lib/drivers/medtronic/cli/blob_loader.js
@@ -46,7 +46,9 @@ function loadFile(filePath, tz, userid) {
  */
 function processMedtronic(driverMgr){
   driverMgr.process('Medtronic', function(err, result) {
-    console.log('Result:', result);
+    if(result && result.post_records) {
+      console.log('POST records:', JSON.stringify(result.post_records, null, '\t'));
+    }
     if (err) {
       console.log(intro, 'Error processing Medtronic data:', err);
       console.log(err.stack);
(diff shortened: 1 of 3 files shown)


gniezen:

lib/drivers/medtronic/cli/blob_loader.js
```diff
@@ -46,7 +46,9 @@ function loadFile(filePath, tz, userid) {
  */
 function processMedtronic(driverMgr){
   driverMgr.process('Medtronic', function(err, result) {
-    console.log('Result:', result);
+    if(result && result.post_records) {
+      console.log('POST records:', JSON.stringify(result.post_records, null, '\t'));
```
pazaan: `debug`?

test/node/medtronic/testProcessData.js
```diff
@@ -320,7 +321,8 @@ describe('processData.js', function() {
           'resumeIndex' : 4,
     			'payload': {
     				'lgs_types': [
-    					'Suspend user',
+              'Suspend low glucose',
```
pazaan: Inconsistent whitespace.

lib/drivers/medtronic/cli/blob_loader.js
gniezen: This is the CLI tool, so using `debug` isn't really that useful.

test/node/medtronic/testProcessData.js
gniezen: Thanks, fixed it just now but forgot to push before merge. Cherry-picking into main branch.

PR #404, opened 2017-05-22
MedT Direct bugfix: Missing temp basal duration

gniezen:
During an alpha test, a temp basal that was suspended and then resumed, followed by a reservoir change, resulted in a missing duration. This PR changes instances of `with_duration` so that we're always setting the duration after the initial record was created.

It also adds a regression test for the situation that triggered the missing duration in the first case.


gniezen:

PR #405, opened 2017-05-23
Electron tweaks

jebeck:
For developer happiness mainly:
- minor README updates for clarity
- logic to avoid `autoUpdater` check that produces a Webpack error on every instance of `npm run dev`
- updated the Yarn lockfile (dunno if you wanted me to include this - if not, can make a PR without it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ This README is focused on just the details of getting the uploader running local
 ## How to set it up
 
 1. Clone this repository.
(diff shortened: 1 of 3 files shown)


jebeck:
Hmm, the builds failed. Something related to the `electron` install. I see you're using `yarn --pure-lockfile` rather than the `yarn --frozen-lockfile` we landed on for blip. Is that just because it should be updated @krystophv?

krystophv:
Build failures seem to just be transient issues with the CI boxes (network issues on the Appveyor and cache problems on Circle). I'm rebuilding them now and will likely merge this later today.

PR #406, opened 2017-05-24
device implementation checklist additions

jebeck:
Take a look in particular at the OmniPod checklist and the Tandem pump one for substantive (i.e., non-template) additions. @gniezen @pazaan

### DIFF
--- a/docs/checklists/tandem.md
+++ b/docs/checklists/tandem.md
@@ -233,7 +233,7 @@ No Tidepool data model yet:
 
 ##### Device-specific? (Add any device-specific notes/additions here.)
 
-At the moment, we only process data from the most recent pump shut-down event, if any exist. This is because a subtle assumption of BtUTC is that the pump clock is always running. Unfortunately, when a Tandem device is shut down, the clock stops. There are several ways we might try to handle these shut-downs and still do bootstrapping, but we'll need to research and potentially prototype these to figure out the best approach and then implement it.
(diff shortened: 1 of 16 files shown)


docs/checklists/tandem.md
```diff
@@ -242,3 +242,17 @@ At the moment, we only process data from the most recent pump shut-down event, i
...
+Choose one of the following:
+
+  - `[x]` legacy "jellyfish" ingestion API
+  - `[ ]` platform ingestion API
+
+### Known implementation issues/TODOs
+
+*Use this space to describe device-specific known issues or implementation TODOs **not** contained in the above datatype-specific sections.*
+
+  - `*[ ]` On first upload, read back to the very beginning of the pump's history to construct a full `pumpSettings` history, allowing for fuller `pumpSettings` history (including profile changes) as well as basal schedule look-up.
```
gniezen: This is already in the checklist:

> - To build a full settings history we'll need to use the entire log record, not just the last 90 days. For example, a profile name change event only contains the new name, so you'll need to read the record where it was first created to get the original name. Until we do diff uploads (so it only happens once), or significantly improve the upload speed, reading the entire log record is not possible.

docs/checklists/tandem.md
```diff
@@ -242,3 +242,17 @@ At the moment, we only process data from the most recent pump shut-down event, i
...
+  - `*[ ]` Find a reliable way to identify (and then drop) `timeChange` events occurring after a device shutdown so that BtUTC doesn't have to be limited to records since the last pump shutdown.
```
gniezen: This is also in the checklist already:

> At the moment, we only process data from the most recent pump shut-down event, if any exist. This is because a subtle assumption of BtUTC is that the pump clock is always running. Unfortunately, when a Tandem device is shut down, the clock stops. There are several ways we might try to handle these shut-downs and still do bootstrapping, but we'll need to research and potentially prototype these to figure out the best approach and then implement it.

Maybe merge them into one bullet point?

jebeck:
Ooops, yeah, will merge those dups into the datatype-specific sections in Tandem.

PR #408, opened 2017-05-26
OmniPod: Bugfix for error codes

gniezen:
A user reported an issue that was triaged as an incorrect wizard bolus value. Further inspection showed that there was a checksum error on the relevant record. 

According to the spec, records with error codes should not be processed. In this specific instance, however, it leads to some missing details for the relevant wizard record. As these details are optional according to the data model, this PR modifies `objectBuilder` and its tests to match the data model.


lib/drivers/insulet/insuletDriver.js
```diff
@@ -652,10 +652,12 @@ module.exports = function (config) {
         continue;
       }
       if (rec.error_code) {
-        debug('Nonzero logRecord error_code!');
+        // according to the spec, a record with an error code should not be parsed
```
jebeck: This was actually on the checklist from our meeting. Maybe the checklist should be updated?

lib/drivers/insulet/insuletDriver.js
```diff
@@ -1066,16 +1068,23 @@ module.exports = function (config) {
...
+
+            if(wiz.detail.target_bg && wiz.detail.bg_correction_threshold) {
+              postwiz = postwiz.with_bgTarget({target: wiz.detail.target_bg,
+                               high: wiz.detail.bg_correction_threshold });
+            }
+
+            if(wiz.detail.corr_units_iob && wiz.detail.meal_units_iob) {
+              postwiz = postwiz.with_insulinOnBoard(common.fixFloatingPoint(wiz.detail.corr_units_iob + wiz.detail.meal_units_iob, 2));
+            }
+
```
jebeck: I assume these changes are not essential to this bugfix (since now we're just dropping the offending record), but are a low-hanging fruit improvement that you also thought would be good to keep. Is that correct?

lib/objectBuilder.js
```diff
@@ -431,13 +431,13 @@ module.exports = function () {
   function makeWizard() {
     var rec = _.assign(_createObject(), deviceInfo, {
       type: 'wizard',
-      recommended: REQUIRED,
+      recommended: OPTIONAL,
```
jebeck: These changes affect all the drivers, including those pushing up data through the platform. I did a double-check that all these fields are optional via jellyfish. Have you checked that they are all optional via platform?

jebeck:
Also, what's the deal with the failure on CircleCI? I assume Circle is only relevant for the Electron work... Is there a way to turn the builds off on Circle for non-Electron work?

lib/drivers/insulet/insuletDriver.js
gniezen: Ah, yes, of course! I'll also remove the TODO in the code.

lib/drivers/insulet/insuletDriver.js
gniezen: No, they are actually required because the record being dropped in the .ibf we have is part of a wizard entry. My approach is to make a "best effort" at uploading what we have here. Do you think we should drop the entire record instead?

lib/objectBuilder.js
gniezen: Good idea. I have just tested this by uploading MedT Direct via platform with those fields disabled and it works as expected.

lib/drivers/insulet/insuletDriver.js
jebeck: Hmm, based on the TODO comment that was in the code before and my skim of the spec last week, I thought the spec indicated that the entire record should be dropped if there's an error code, but perhaps I misunderstood. If you're confident this is to spec, then 👍

gniezen:
[[uploader][bug] OmniPod records with checksum errors should be dropped](https://trello.com/c/r7sDfxHz/186-uploader-bug-omnipod-records-with-checksum-errors-should-be-dropped)

PR #410, opened 2017-05-29
MedT Direct: Various bugfixes and updates

gniezen:
This PR contains the following bugfixes and updates:

- reads number of pump history pages as integer, not byte
- moves pump & CGM checklists to right directory and adds to summary
- flushes once on error while reading pump history - previously it could flush twice at the same time, leading to unpredictable results in terms of callbacks
- When an upload fails, the error message displayed in the UI will remind the user not to run CareLink, OpenAPS or Loop at the same time


gniezen:
@pazaan I noticed that parts of the diff is a bit icky - basically I just shuffled some code into an if/else statement to prevent the `flush()` command being called while it could still be running, which would lead to a "callback was already called" error message.

gniezen:

lib/redux/constants/errors.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
   E_HID_CONNECTION: 'Hmm, your device doesn\'t appear to be connected',
   E_INIT: 'Error during app initialization',
   E_MEDTRONIC_UNSUPPORTED: 'Tidepool does not support Minimed pumps 522, 722 or older, or the newer 6-series pumps. Sorry...',
+  E_MEDTRONIC_UPLOAD: 'Is there another device (e.g. CareLink/OpenAPS/Loop) talking to your pump? If not, see error details below or contact Tidepool Support.',
```
pazaan: Do we need an "if so" case?

gniezen: I agree, but there's not really enough space at the moment. Once we have an updated design (see https://trello.com/c/8nJtBvT7) we should consider updating this error message.

PR #411, opened 2017-05-30
MedT Direct: updated temp basal after schedule change

gniezen:
Jana discovered incorrect suppressed info while looking through Eric's uploaded MedT Direct data. I traced this back to a temp basal being updated after a schedule change. We should be checking if there were any schedule changes before the temp basal was updated. This PR is a one-liner that fixes that and includes a unit test for properly testing this scenario.


gniezen:

PR #412, opened 2017-06-01
Electron Uploader: Updating metrics

gniezen:
This PR removes the device ID from the metric event name and ensures that it's returned in the `source` property. Previously CareLink would have `undefined` as `source`.


gniezen:
[<XS> Electron Uploader - KissMetrics](https://trello.com/c/cbseB9qW/135-xs-electron-uploader-kissmetrics)

PR #414, opened 2017-06-09
Add `bg/out-of-range` annotations to `cbg` types.

pazaan:
@gniezen

### DIFF
--- a/docs/checklists/medtronic.md
+++ b/docs/checklists/medtronic.md
@@ -114,7 +114,8 @@ Device-specific? (Add any device-specific notes/additions here.)
   - `[x]` subType (`linked` or `manual`)
   - `[x]` units of value (read from device, not hard-coded)
   - `[ ]` out-of-range values (LO or HI)
-  - `[?]` out-of-range value thresholds (e.g., often 20 for low and 600 for high on BGMs)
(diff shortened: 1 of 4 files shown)


pazaan:
While starting to implement the annotations for SMBG, I realised that it doesn't make much sense to do so. The Bolus Wizard only lets you enter a number in the range of the Contour Next Link (20mg/dL to 600mg/dL). You would want both 20 and 600 to show up as valid SMBG numbers (not "Hi" or "Lo"). If you have a reading outside of this range, it's not automatically transmitted to the pump, and you couldn't enter it into the Bolus Wizard, because you wouldn't know what the number is.
Thoughts?

gniezen:
I vaguely remember coming to the same conclusion while working on SMBG records - that there isn't really a way to get out-of-range numbers, for the reasons you specified. Will you mention something to that effect in the checklist?

pazaan:
@gniezen Can you just re-review? I removed a duplicate const, just want to make sure you're happy with that.

gniezen:
LGTM! I'll merge it in so that I can start a new branch for the CGM timestamp fixes.

gniezen:

PR #416, opened 2017-06-14
Medtronic Direct: CGM timestamps

gniezen:
When there's no `SENSOR_TIMESTAMP` at the end of a CGM history, the timestamps for CBG values are different between values.

This PR ensures a new CGM history timestamp is written before reading CGM history. It also fixes a regression where a suspended pump could not be uploaded (surfacing as `type` error).


gniezen:
Oh, we are already using `SENSOR_CAL_BG` for calibration events, so I'll just change the other one.

gniezen:

PR #417, opened 2017-06-15
MedT Direct: Fill in one-second gaps

gniezen:
In 10-20% of cases a scheduled basal starts one second _after_ a temp basal duration ends. 

This PR fills in the one-second gap with a one-second temp basal equivalent to the one that preceded it (including suppressed info) so that we can calculate bolus-to-bolus ratio stats correctly.

A unit test is also included.

### DIFF
--- a/lib/drivers/medtronic/medtronicSimulator.js
+++ b/lib/drivers/medtronic/medtronicSimulator.js
@@ -398,6 +398,36 @@ exports.make = function(config){
         }
       }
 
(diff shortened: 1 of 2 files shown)


gniezen:

gniezen:
A duration of 0 in the CSV just means that the temp basal was cancelled. So what's happening is that the cancelled temp basal gets finished up and pushed on the stack before it gets to the part in the code where it fills in the one-second gap. Good catch, thanks!

PR #419, opened 2017-06-16
MedT Direct: handle case where bolus event is more than 5 seconds after bolus wizard

gniezen:
Currently we expect a bolus record to be within 5 seconds of its corresponding bolus wizard record to match them (usually they have the same timestamp, but not always). During alpha testing we discovered a bolus record that came 13 seconds (!) after the bolus wizard record. This PR changes the window to 30 seconds to make sure we can match these records.

### DIFF
--- a/lib/drivers/medtronic/medtronicSimulator.js
+++ b/lib/drivers/medtronic/medtronicSimulator.js
@@ -405,10 +405,10 @@ exports.make = function(config){
         // A wizard event is always followed by a bolus event.
         // Usually the timestamp is the same, but sometimes the bolus timestamp
         // is a second later. Here we also check that the bolus event is at least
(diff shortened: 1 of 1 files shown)


gniezen:

pazaan:
Does the timing have to do with the size of the bolus? For example, would the bolus record for an 8U bolus take longer to appear after the BWZ record than a 2U bolus, for example?
What about a square wave?

gniezen:
I thought that may be the case too. However, most of the time the bolus wizard record timestamp matches the corresponding bolus record timestamp exactly, and since it takes at least some seconds before a bolus is delivered, I would expect there to always be a non-zero difference if it waited for the bolus to complete first.

PR #420, opened 2017-06-19
MedT Direct: Old schedule drops off end

gniezen:
While testing on the 523 pump I have here, I discovered that it's possible that if the oldest record on the pump is a `CHANGE_BASAL_PROFILE_NEW`, it won't be able to match to its corresponding `CHANGE_BASAL_PROFILE_OLD` as it has dropped off the end. 

Since it's the first/oldest record (i.e., there are no basals prior to this change that can be affected), we can drop the `CHANGE_BASAL_PROFILE_NEW` in this instance.


gniezen:

PR #422, opened 2017-06-22
MedT Direct: handle temp basal with schedule changes before automatic suspend/resume

gniezen:
When there is a temp basal with a reservoir change, we need to infer the resumption of the temp basal from the prime event. We also need to take into account any scheduled changes that took place before the suspend event.

@pazaan, would you mind verifying this with the affected alpha user's binary blob and look at the events that are created? I have done this, but it will be good for someone else to double-check.


gniezen:

PR #423, opened 2017-06-26
Electron: Adding Rollbar and crashReporter for evaluation

gniezen:
This PR adds sentry.io, TrackJs and Electron crashReporter to evaluate them while we are seeing some Electron crashes during testing.


krystophv:
[[Uploader] Send error reports automatically](https://trello.com/c/cYSNiB1W/196-uploader-send-error-reports-automatically)

krystophv:
@gniezen I realize that I'm tagged for review on this, but somewhat awkwardly it seems that almost all the code changes are mine now. If you can review these changes, I can give it the 'Approved'.

krystophv:
I think I figured it out. Rollbar wraps the `window.console`. The reason that `App` was being properly collected was because it actually get's it's `bows()` instance when the component has mounted, which will come _after_ the Rollbar instantiation code. The rest of the modules get their `bows()` instance essentially during parsing/load. If I move the Rollbar instantiation to it's own module and `import` it at the very top of the renderer portion of the application, then I'm pretty sure that the `window.console` has been wrapped prior to the rest of the app making their `bows()` instances. In my latest test error, I got pretty much everything.

gniezen:
Ah, good thinking! This is going to be so useful once we get it out there!

gniezen:
Closing this one as it has been merged into #534

PR #430, opened 2017-07-05
fix first char case of medtronic driver name

DorianScholz:
new driver is not found due to case sensitive filenames on Linux


gniezen:
LGTM. We should enable devices in `devices.js` as they're tested on Linux so that they can appear on the Choose Devices screen. I should find out what testing is required for each device.

PR #435, opened 2017-07-07
Add more CBG types that can increment time

gniezen:
There are events like SENSOR_WEAK_SIGNAL and SENSOR_CALIBRATION_EVENT that sometimes increment the time with another 5 minutes. Prior to this PR we only took SENSOR_WEAK_SIGNAL into account, but there are others. A combination of the descriptor and the type have to be used to identify which ones are used.


gniezen:

PR #438, opened 2017-07-13
Support for OneTouch Ultra2

gniezen:
We've had the Ultra2 driver in the codebase for a while now, but it has always been disabled since it didn't have proper error handling and didn't adhere to the UTC bootstrapping time format.

This PR adds:
- TZOUtil across-the-the-board time
- units from meter, not hard-coded
- BGM checklist
- skipping control solution tests
- annotating HI/LOs

### DIFF
--- a/lib/drivers/onetouch/oneTouchUltra2.js
+++ b/lib/drivers/onetouch/oneTouchUltra2.js
@@ -352,10 +352,6 @@ module.exports = function (config) {
           return cb(null, data);
         }
       });
-
-      progress(100);
-      data.cleanup = true;
-      cb(null, data);
     },
 
     disconnect: function (progress, data, cb) {
--- a/lib/drivers/onetouch/oneTouchUltra2.js
+++ b/lib/drivers/onetouch/oneTouchUltra2.js
@@ -126,7 +126,7 @@ module.exports = function (config) {
       clearInterval(listenTimer);
 
       callback(null, pkt);
-    }, 1000);     // spin on this one not so quickly
+    }, 2000);     // spin on this one not so quickly
   };
 
   var probe = function(cb){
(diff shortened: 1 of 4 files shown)


gniezen:
[OneTouch Ultra 2 and OneTouch Mini](https://trello.com/c/drBwiv9K/198-onetouch-ultra-2-and-onetouch-mini)

lib/drivers/onetouch/oneTouchUltra2.js
```diff
@@ -213,43 +199,73 @@ module.exports = function (config) {
      }
 
     // get data from lines
-    var dpattern = /P "(\w+)","(\d{2}\/\d{2}\/\d{2})","(\d{2}:\d{2}:\d{2})   ","..(\d{3}).","(.)","(\d{2})",\W00\W(.+)/;
+    var dpattern = /P "(\w+)","(\d{2}\/\d{2}\/\d{2})","(\d{2}:\d{2}:\d{2})   ","(..\d{3}).","(.)","(\d{2})",\W00\W(.+)/;
```
DorianScholz: The regex for the glucose value does not capture the optional trailing question mark which signals a parity error according to the specs. Is that intentional, or should there be some specific handling of values with this flag?

gniezen: True, it's currently ignoring the parity error flag. I think the correct response would be to drop the reading if there is a parity error. I'll update this right now.

gniezen:
You're right, line 356-358 should not be there.

DorianScholz:
The two new commits look good to me.
From my perspective of a visual code review, this PR is ready to be merged.
Should I approve it, or leave that to someone that has actually tested it against the device?

gniezen:
You can approve. I will then merge it into master and create a new version of the dev uploader for Eric or Nick to test before we push to production.

gniezen:
@pazaan Dorian has already reviewed the driver, but I have made some additional changes in commit b03ba87 after I discovered that sometimes it only uploaded around 12 out of 20 records. Can you please review the additional changes? (I also need someone with write access to approve)

PR #442, opened 2017-07-25
Electron: Bayer fixes

gniezen:
I think we used to skip non-glucose records, so this is likely a regression. The meter checklist indicates that carb/insulin records on meter will be implemented in future as it requires changes to the data model.

I also changed the error message for a malformed ASTM message, as it appears when you click Upload immediately after the meter is plugged in and not yet ready for comms.


gniezen:
[[uploader] Bayer Contour Next additional types](https://trello.com/c/ZVwTT29I/211-uploader-bayer-contour-next-additional-types)

PR #446, opened 2017-07-26
Bug fixes for `finalBasal` and code cleanup

pazaan:
- Call `simulator.finalBasal()` before getting the final set of events to upload
- Code clean-up
- Use time of last history event instead of current timestamp for TZOUtil

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -872,7 +872,7 @@ class NGPHistoryParser {
     return model;
   }
 
-  buildTimeChangeRecords(mostRecent) {
+  buildTimeChangeRecords() {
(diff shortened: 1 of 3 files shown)


lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -888,6 +888,7 @@ class NGPHistoryParser {
       timeChangeEvents.push(timeChange);
     }
 
+    const lastEvent = this.events[this.events.length - 1];
```
gniezen: Just checking: is it guaranteed that the last event will always contain a timestamp?

pazaan: Yes, every 600 series event comes with a timestamp (nice, right?)

PR #448, opened 2017-07-28
Find record using pointer from error

gniezen:
When the platform returns an error, there is an index pointer to the specific record in the block sent during the API call. This PR finds that offending record using the pointer.

### DIFF
--- a/lib/core/api.js
+++ b/lib/core/api.js
@@ -392,14 +392,16 @@ function addDataToDataset(data, datasetId, blockIndex, uploadType, callback) {
       if(status.meta) {
         err.requestTrace = status.meta.trace.request;
       }
+      api.log(JSON.stringify(status, null, '\t'));
       if(status.errors && status.errors.length > 0) {
-        var hpattern = /\/(\w+)\//;
-        var toMatch = hpattern.exec(status.errors[0].source.pointer);
-        if (toMatch[1]) {
-          api.log('Offending record:', JSON.stringify(data[parseInt(toMatch[1])], null, '\t'));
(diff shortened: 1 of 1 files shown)


lib/core/api.js
```diff
@@ -392,7 +392,16 @@ function addDataToDataset(data, datasetId, blockIndex, uploadType, callback) {
       if(status.meta) {
         err.requestTrace = status.meta.trace.request;
       }
-      api.log(JSON.stringify(status));
+      api.log(JSON.stringify(status, null, '\t'));
+      if(status.errors && status.errors.length > 0) {
+        for (var i in status.errors) {
+          var hpattern = /\/(\w+)\//;
```
darinkrauss: Should this be `\d` not `\w`?

lib/core/api.js
```diff
@@ -392,7 +392,16 @@ function addDataToDataset(data, datasetId, blockIndex, uploadType, callback) {
...
+          var toMatch = hpattern.exec(status.errors[i].source.pointer);
+          if (toMatch[1]) {
+            api.log('Offending record for error',i,':', JSON.stringify(data[parseInt(toMatch[1])], null, '\t'));
```
darinkrauss: Space between parameters?

lib/core/api.js
gniezen: Yep, that will be better.

gniezen:
As this change is not user-facing, I'm going to merge this without a version bump and it will form part of the next release (which will most likely be the next iteration of the Medtronic bugfix).

PR #450, opened 2017-07-28
Update dialog state fix

krystophv:
Unravels some of the deep logic involved in the update modal dialog display to make it (hopefully) a little more readable. Also revealed a reducer that wasn't reacting to a couple of update events that it should have been which was flashing incorrect messages temporarily in some circumstances.

### DIFF
--- a/app/components/UpdateModal.js
+++ b/app/components/UpdateModal.js
@@ -48,40 +48,57 @@ export class UpdateModal extends Component {
       sync
     } = this.props;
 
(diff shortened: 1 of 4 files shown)


krystophv:
[BUG - Electron - "Install" button doesn't work when displayed after logging in.](https://trello.com/c/mVAWN9tA/208-bug-electron-install-button-doesnt-work-when-displayed-after-logging-in)

PR #451, opened 2017-08-01
Add mechanism to prepare QA builds

pazaan:
Also add `DEBUG_ERROR=true` to `dev` builds in order to make Blob downloads available.


pazaan:
@krystophv - I considered also adding an automatic commit at the end, to make the QA build completely consistent (ie, one can't change another file, then commit and push). Thoughts?

package.json
```diff
@@ -199,6 +203,7 @@
     "fbjs-scripts": "0.7.1",
     "file-loader": "0.11.1",
     "flux-standard-action": "1.2.0",
+    "git-describe": "^4.0.2",
```
krystophv: These just need to be exact versions.

scripts/prepare-qa-build.js
```diff
@@ -0,0 +1,32 @@
...
+
+const gitInfo = gitDescribeSync();
+if (gitInfo.dirty) {
+  console.log('Please start with a clean repo');
+  shell.exit(1);
+}
+
+const qaVersion = gitInfo.semverString.replace('+', '.');
+
+(async() => {
```
krystophv: This causes an error on `node 7.4.0` without running this script command through `babel` first (accomplished in the other npm scripts by using `babel-register`).

pazaan: Updated `package.json` to fix. Thanks!

krystophv:
LGTM 👍

PR #454, opened 2017-08-03
Add support for 670g SMBG history type

### LINKED ISSUE #452: SMBGs don't appear for 670G
### Description
When downloading pump data from a 670G series device, the SMBG's don't appear in the trend view or daily view.

### Environment
**Uploader version**: 0.310.0-alpha.11.135.gb0298b2e
**OS**: macOS

pazaan:
Fixes #452 
Also includes some code simplification, clarifications and warning fixes.

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -39,6 +39,9 @@ class NGPHistoryEvent {
       START_OF_DAY_MARKER: 0x07,
       END_OF_DAY_MARKER: 0x08,
       PLGM_CONTROLLER_STATE: 0x0B,
+      CLOSED_LOOP_STATUS_DATA: 0x0C,
+      CLOSED_LOOP_PERIODIC_DATA: 0x0D,
+      CLOSED_LOOP_DAILY_DATA: 0x0E,
       NORMAL_BOLUS_PROGRAMMED: 0x15,
       SQUARE_BOLUS_PROGRAMMED: 0x16,
       DUAL_BOLUS_PROGRAMMED: 0x17,
@@ -59,10 +62,10 @@ class NGPHistoryEvent {
       ALARM_CLEARED: 0x2A,
       LOW_RESERVOIR: 0x2B,
       BATTERY_INSERTED: 0x2C,
-      FOOD_MARKER: 0x2E,
-      EXERCISE_MARKER: 0x2F,
-      INJECTION_MARKER: 0x30,
-      OTHER_MARKER: 0x31,
+      FOOD_EVENT_MARKER: 0x2E,
+      EXERCISE_EVENT_MARKER: 0x2F,
+      INJECTION_EVENT_MARKER: 0x30,
+      OTHER_EVENT_MARKER: 0x31,
       BG_READING: 0x32,
       CODE_UPDATE: 0x33,
       MISSED_MEAL_BOLUS_REMINDER_EXPIRED: 0x34,
@@ -72,8 +75,10 @@ class NGPHistoryEvent {
       ACTIVE_INSULIN_CLEARED: 0x39,
       DAILY_TOTALS: 0x3C,
       BOLUS_WIZARD_ESTIMATE: 0x3D,
+      MEAL_WIZARD_ESTIMATE: 0x3E,
+      CLOSED_LOOP_DAILY_TOTALS: 0x3F,
       USER_SETTINGS_SAVE: 0x50,
-      USER_SETTINGS_RESETTO_DEFAULTSS: 0x51,
+      USER_SETTINGS_RESET_TO_DEFAULTS: 0x51,
       OLD_BASAL_PATTERN: 0x52,
       NEW_BASAL_PATTERN: 0x53,
       OLD_PRESET_TEMP_BASAL: 0x54,
@@ -109,6 +114,12 @@ class NGPHistoryEvent {
       SET_CHANGE_REMINDER_CHANGE: 0x78,
       BLOCK_MODE_CHANGE: 0x79,
       BOLUS_WIZARD_SETTINGS_SUMMARY: 0x7B,
+      CLOSED_LOOP_BG_READING: 0x82,
+      CLOSED_LOOP_OPTION_CHANGE: 0x86,
+      CLOSED_LOOP_SETTINGS_CHANGED: 0x87,
+      CLOSED_LOOP_TEMP_TARGET_STARTED: 0x88,
+      CLOSED_LOOP_TEMP_TARGET_ENDED: 0x89,
+      CLOSED_LOOP_ALARM_AUTO_CLEARED: 0x8A,
       SENSOR_SETTINGS_CHANGE: 0xC8,
       OLD_SENSOR_WARNING_LEVELS: 0xC9,
       NEW_SENSOR_WARNING_LEVELS: 0xCA,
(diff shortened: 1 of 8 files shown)


lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -315,20 +329,64 @@ class BloodGlucoseReadingEvent extends NGPHistoryEvent {
...
+
+  get bgReadingOrigin() {
+    // eslint-disable-next-line no-bitwise
+    return this.eventData[0x16] & 2 ?
+      NGPUtil.NGPConstants.BG_ORIGIN.MANUALLY_ENTERED :
+      NGPUtil.NGPConstants.BG_ORIGIN.RECEIVED_FROM_RF;
+  }
+
+  // See NGPUtil.NGPConstants.BG_CONTEXT
+  get bgReadingContext() {
```
pazaan: Can you clarify what that means?

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -126,103 +144,63 @@ class BCNLMessage {
...
+
+        message = Buffer.concat([message, packet.slice(4)], (message.length + packet.length) - 4);
+      }
+      // USB_BLOCKSIZE - 4, because we don't include the MAGIC_HEADER or the size byte
+    } while (size === (BCNLMessage.USB_BLOCKSIZE - 4));
+
+    abortTimer.stop();
+
+    debug('### READ USB DATA', message.toString('hex'));
+    return message;
```
pazaan: I had the same thought, but when I had less clearTimeout/startTimeout functions lying around, I knew it was worth it. Even more so if we ever have the cause to use `Timer.reset()`

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -126,103 +144,63 @@ class BCNLMessage {
...
+    while (pos < message.length) {
+      const bytes = Buffer.alloc(BCNLMessage.USB_BLOCKSIZE);
+      const sendLength = (pos + 60 > message.length) ? message.length - pos : 60;
+      bytes.write(BCNLMessage.MAGIC_HEADER, 0);
+      bytes.writeUInt8(sendLength, 3);
+      bytes.write(message.slice(pos, pos + sendLength).toString('binary'), 4, 'binary');
+      debug('### SENDING USB DATA', bytes.toString('hex'));
+
+      // eslint-disable-next-line no-await-in-loop, requests to devices are sequential
+      await new Promise(resolve => hidDevice.send(bytes.buffer.slice(), resolve()));
```
pazaan: Yes, that's right 😄

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -259,41 +237,29 @@ class DeviceInfoRequestCommand extends BCNLCommand {
...
-            astmInfo = Buffer.from(response1).toString('ascii');
-            new BCNLCommandResponse(response2).checkAsciiControl(BCNLMessage.ASCII_CONTROL.ENQ);
-          } else {
-            astmInfo = Buffer.from(response2).toString('ascii');
-            new BCNLCommandResponse(response1).checkAsciiControl(BCNLMessage.ASCII_CONTROL.ENQ);
-          }
-          const message = new DeviceInfoRequestResponse(astmInfo);
-          resolve(message);
-        })
-        .catch((err) => {
```
pazaan: Only if we want to handle them at that stage. With Promises, we had to always catch and rethrow (which is what I was doing here). With `async/await`, we don't need to do that any more, so the `try/catch` is in `fetchData`, where it's actually being handled.

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -259,41 +237,29 @@ class DeviceInfoRequestCommand extends BCNLCommand {
...
-          reject(err);
-        });
-    });
+    // Timeout might be zero, but the callback will fire anyway
+    // We use sendMessage instead of super.send() because we want to pull raw
+    // data only, and we can't determine response types until after we've checked
+    // them. The CNL can send the messages in different orders.
+    await this.sendMessage(hidDevice);
+    await Timer.wait(sendDelay);
```
pazaan: It *probably* isn't. In fact, I only put them in here because it was in the CLP code. I'll remove them for now and see if we get any adverse affects.

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -1640,24 +1578,19 @@ class MM600SeriesDriver extends BCNLDriver {
...
+  async getBolusWizardSettings() {
+    const settings = {};
+    let response = await new BolusWizardBGTargetsCommand(this.pumpSession).send(this.hidDevice);
+    settings.bgTarget = response.targets;
+
+    response = await new BolusWizardCarbRatiosCommand(this.pumpSession).send(this.hidDevice);
+    settings.carbRatio = response.ratios;
+
+    response =
+      await new BolusWizardSensitivityFactorsCommand(this.pumpSession).send(this.hidDevice);
```
pazaan: It breaks the eslint rule for max line size. But it wasn't as readable as it could be, so I'll change it to:
```
response = await new BolusWizardSensitivityFactorsCommand(this.pumpSession)
      .send(this.hidDevice);
```

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -1668,80 +1601,61 @@ class MM600SeriesDriver extends BCNLDriver {
...
+      const response = await new ReadBasalPatternCommand(this.pumpSession, i + 1)
+        .send(this.hidDevice);
+      const schedule = response.schedule;
+      // Only include patterns with schedules in them
+      if (schedule.length > 0) {
+        basalSchedules[NGPUtil.NGPConstants.BASAL_PATTERN_NAME[i]] = schedule;
+      }
+    }
+
+    return basalSchedules;
```
pazaan: It makes me so happy! 😹

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -1887,6 +1801,7 @@ module.exports = (config) => {
         .then(() => debug('Read pump history'))
         .then(() => {
           const threeMonthsAgo = new Date(new Date().valueOf() - 7776e6);
+          // const threeMonthsAgo = new Date(new Date().valueOf() - 181e6); // Three weeks
```
pazaan: I will remove it. I have it here because I'm in the middle of fixing other bugs, but upon further reflection, I'll make a Stash, rather than commented code.

lib/drivers/medtronic600/NGPHistoryParser.js
gniezen: In our "grab bag" payload field, it may be useful (for example for debugging purposes) to add the BG reading context you're getting in `bgReadingContext`, e.g.

    payload : { readingContext: 'entered in bolus wizard' }

in the BG records. If that's a lot of work or if you think it won't be useful, ignore I said it :)

lib/drivers/medtronic600/medtronic600Driver.js
gniezen: 👍

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Ah, good idea! I'll add that in.
Sorry, I got confused between "Data model payload" and "Medtronic protocol payload" ;)

PR #455, opened 2017-08-03
Medtronic: Bugfixes

gniezen:
Considering the following scenario:
A scheduled basal is followed by a series of temp basals. We look at the first scheduled basal event to fill in the suppressed info of the subsequent temp basals. What happens when the first scheduled basal has rolled off the end of the pump history and only the temp basals remain? 

The fix is to always fill in the suppressed info of a temp basal using the basal schedule at the time temp basal starts.

This PR also checks that we don't auto-suspend (e.g. with no delivery alarm) if we're already suspended.

### DIFF
--- a/lib/drivers/medtronic/medtronicSimulator.js
+++ b/lib/drivers/medtronic/medtronicSimulator.js
@@ -223,7 +223,7 @@ exports.make = function(config){
 
       ensureTimestamp(event);
 
-      if(event.deliveryType === 'temp') {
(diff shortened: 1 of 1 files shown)


gniezen:
[[uploader][bug] Medtronic Direct upload with Eric's pump fails platform validation](https://trello.com/c/bnEjgwgW/213-uploaderbug-medtronic-direct-upload-with-erics-pump-fails-platform-validation)

PR #456, opened 2017-08-07
Electron: Remove old kernel extensions during installation

gniezen:
It seems macOS do not like when kernel extensions are updated by just copying over the new files - the old kernel extension should be completely removed before the new one is copied in place.


gniezen:
@pazaan I'm going to merge this one already so that I can test it, but please let me know if you find any issues.

PR #457, opened 2017-08-08
Set default protocol handler

krystophv:
What it says on the tin. Not a whole lot going on here, just registering a custom protocol on launch.

Pairs well with: https://github.com/tidepool-org/blip/pull/436


krystophv:
[Electron Uploader - Blip Integration](https://trello.com/c/CTjNregM/212-electron-uploader-blip-integration)

PR #459, opened 2017-08-10
Restrict Mac install location to user-owned Applications

krystophv:
In order to avoid the issue where a user can't apply an automatic update to an App in a directory that they don't have user-level write access to (https://github.com/Squirrel/Squirrel.Mac/pull/207 which hasn't as of yet been applied/released through stable Electron), we need to restrict the user to installing the App to their $HOME/Applications directory instead of the system /Applications. This also involved upgrading `electron-builder`, which _almost_ resolved the artifact naming issue we have with multiple sets of releases (`dev` vs production) on GitHub, except that their GitHub artifactName still completely renames the system distributed artifactName. *sigh*

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,7 @@
 os: unstable
 
+image: Visual Studio 2017
+
 environment:
   APPVEYOR_CACHE_ENTRY_ZIP_ARGS: -t7z
   matrix:
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "dev": "npm run hot-server -- --start-hot",
     "prepare-qa-build": "node -r babel-register scripts/prepare-qa-build.js",
     "package": "npm run build && build -p onTagOrDraft",
-    "package-dev": "npm run build-dev && build -p onTagOrDraft --config.artifactName='${productName}-${version}-dev.${ext}' --config.nsis.artifactName='${productName}-Setup-${version}-dev.${ext}'",
+    "package-dev": "npm run build-dev && build -p onTagOrDraft --config.artifactName='tidepool-uploader-${version}-${arch}-${os}-dev.${ext}' --config.nsis.artifactName='tidepool-uploader-setup-${version}-dev.${ext}' --config.pkg.artifactName='tidepool-uploader-${version}-dev.${ext}'",
     "package-win": "npm run build && build --win --x64",
     "package-mac": "npm run build && build --mac",
     "package-linux": "npm run build && build --linux",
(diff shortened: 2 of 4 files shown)


appveyor.yml
```diff
@@ -1,5 +1,7 @@
 os: unstable
 
+image: Visual Studio 2017
```
gniezen: Just wondering what necessitated this change?

package.json
```diff
@@ -25,11 +25,11 @@
     "build-dev": "npm run build-main-dev && npm run build-renderer-dev",
     "start": "cross-env NODE_ENV=production electron ./app/",
     "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development",
-    "postinstall": "concurrently \"install-app-deps\" \"node node_modules/fbjs-scripts/node/check-dev-engines.js package.json\" && electron-rebuild --force --module-dir app",
+    "postinstall": "concurrently \"electron-builder install-app-deps\" \"node node_modules/fbjs-scripts/node/check-dev-engines.js package.json\" && electron-rebuild --force --module-dir app",
```
gniezen: Is this the only breaking change between 18 and 19? That's what [the changelog](https://github.com/electron-userland/electron-builder/blob/master/CHANGELOG.md) seems to suggest?

package.json
```diff
@@ -147,6 +149,11 @@
         },
         "dir"
       ]
+    },
+    "pkg": {
+      "artifactName": "tidepool-uploader-${version}.${ext}",
+      "allowAnywhere": false,
+      "allowRootDirectory": false
```
gniezen: Oh electron-builder, this is a pretty non-intuitive way to specify home directory only. Would've been nice if JSON supported comments so that we could make this clearer. I guess it's a temporary solution until they fix the permissions issue in Electron itself. I'm just wondering if any clinics have Macs with multiple users set up - seems unlikely.

appveyor.yml
krystophv: First release note: https://github.com/electron-userland/electron-builder/releases/tag/v19.2.2 
I just didn't question it too hard :)

package.json
krystophv: As far as I could tell, yes. The biggest change was supporting Windows Universal Apps.

package.json
krystophv: Yeah - I sort of understand, as you can gather from their documentation that these flags translate directly to flags in the Mac install API, but just from reading these I wouldn't know what their actual effect is.
I also agree that it seems highly unlikely that clinics would set up individual users for their computers. If that's the case, then we'll have to jump that hurdle when we get there. At least now it can successfully update.

PR #462, opened 2017-08-15
Build speed improvements

krystophv:
Removes some caching of `node_module` directories as I'm pretty sure that the zip/unzip of the cache is slower than the file-copy from the `yarn` global cache (which I have left as a CI cache). Sets the 32-bit windows build to come before the 64 bit as I think the default native modules are 32 bit (for whatever reason), so we might avoid building them 4 times. Also disables the artifact packaging for individual CI machines.


appveyor.yml
```diff
@@ -39,14 +37,14 @@ test_script:
   - yarn test
 #  - yarn test-e2e
   - yarn package-dev
-  - 7z a development.zip release
+#  - 7z a development.zip release
```
gniezen: Does this mean we can also remove the 7zip dependencies, which create those annoying warnings for the dependencies with different os architectures?

krystophv: This is actually a call to the CI box's installed 7zip. The package `electron-devtools-installer` depends on the `7zip` package, so we can't get rid of it.

PR #464, opened 2017-08-16
enforces that carelink shows up last

krystophv:
What it says on the tin, just pulls out carelink from a given list and tacks it on the end. Turns out we don't enforce any sort of sorting on either place (just mapping over object keys which has no guaranteed order).


krystophv:
[[Electron Uploader] Move CareLink checkbox to bottom of list](https://trello.com/c/UkEO5ymQ/227-electron-uploader-move-carelink-checkbox-to-bottom-of-list)

PR #465, opened 2017-08-16
add back in audio queues with new wav files

krystophv:
adds the new audio files and uncomments where they were once commented out.


krystophv:
[Electron Uploader should have notification sounds (earcons) for Upload Complete / Upload failed](https://trello.com/c/faV0ToYq/299-electron-uploader-should-have-notification-sounds-earcons-for-upload-complete-upload-failed)

lib/drivers/medtronic/medtronicDriver.js
```diff
@@ -1586,11 +1586,10 @@ module.exports = function (config) {
     cleanup: function (progress, data, cb) {
       debug('in cleanup');
 
-      /* TODO: put this back in when we have new sound/earcon
       if(isBrowser) {
-        var snd = new Audio('../../../sounds/underwater.wav');
+        var snd = new Audio('../../../sounds/UploadSucceeded.wav');
```
gniezen: Uhm, I think this is duplicate now. IIRC the idea was to either enable it just for Medtronic Direct, or for everything  (in DriverManager).

krystophv: Ok, cool. I was a little confused about it since I couldn't really nail down a place to put the failure case. I'll take out the audio piece from the medtronic specific case.

PR #466, opened 2017-08-18
Change permissions for .app on postinstall

gniezen:
We change the permissions on `Tidepool Uploader.app` to that of the active user, so that the auto-update mechanism doesn't try to do anything with root permissions (e.g. using `/var/root/Library/Caches` instead of `~/Library/Caches`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,10 +147,6 @@
         },
         "dir"
       ]
-    },
-    "pkg": {
-      "allowAnywhere": false,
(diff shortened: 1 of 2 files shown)


krystophv:
Am I correct that we figured out that this won't actually solve _our_ use case because we actually need the `.app` to be running with system-level permissions in order to run the `.pkg` update as the system user to update drivers?

PR #468, opened 2017-08-21
Remove beta label

gniezen:
Tiny one-liner to remove the beta label to prevent confusion with actual betas.


gniezen:
[[uploader] Remove "beta" from Production Uploader](https://trello.com/c/3fRwxot6/205-uploader-remove-beta-from-production-uploader)

PR #469, opened 2017-08-21
Support for FreeStyle Libre

DorianScholz:
This PR adds support for the CGM and BGM functionalities of the Abbott FreeStyleLibre.

What works (please test):
- upload via CLI from device (or data file for testing)
- upload via GUI from device
- bootstrap to UTC via global record index and time change records
- uploading of CBG, SMBG and Ketone data

What should work, but has not really been tested:
- annotation of HI and LO values for CBG, SMBG and Ketone

What is not implemented yet but upcoming:
- uploading via new platform API

What could be implemented in the future:
- use of DB record schemata from device to support future firmwares with new DB format
- upload notes from device
- an even more asynchronous communication, which might speed up the data download a tiny bit
- and as always, more unit tests...


DorianScholz:
@gniezen: There are a few changes in the medtronic driver that made its way from master into this branch, which have not been merged into the electron branch. I'm unsure if we need to remove them from this branch before merging?

lib/drivers/medtronic/medtronicSimulator.js
```diff
@@ -68,7 +68,9 @@ exports.make = function(config){
   };
 
   function setSuspendingEvent(event) {
-    suspendingEvent = event;
```
pazaan: Not sure if Medtronic changes should be in this PR... @gniezen?

test/lib/abbot/testFreeStyleLibreData.js
```diff
@@ -0,0 +1,130 @@
...
+  describe('non-static', () => {
+    let fsLibreData;
+
+    beforeEach(function(){
+      fsLibreData = new FreeStyleLibreData(cfg);
+    });
+
+    describe('process AAP packets', () => {
+
+      it('correctly restores 32bit records numbers', () => {
```
pazaan: eslint will probably warn that `it` scenarios should start with `should`. eg, "should correctly restore 32bit record numbers"

lib/drivers/abbott/freeStyleLibreData.js
```diff
@@ -0,0 +1,466 @@
...
+      this.oldestResultRecordNumber = Math.min(this.oldestResultRecordNumber, headerFields.recordNumber);
+    }
+
+    const dateTime = this.getDateTime(headerFields.readerTime, headerFields.userTimeOffset);
+
+    if (headerFields.recordType === DB_RECORD_TYPE.TIME_CHANGE_RESULT) {
+      const timeChangeFields = struct.unpack(aapPacket.data, offset, FORMAT.TIME_CHANGE,
+        ['oldReaderTime', 'oldUserTimeOffset', 'valid', 'unused', 'CRC16']);
+
+      // TODO: validate CRC16
```
DorianScholz: All the packets have already been checked with CRC32 checksums on the transport layer.
So any alteration of the data during transport is already being detected.
I'm guessing they added this additional CRC16 to catch DB/internal memory errors of the device?
To validate the CRC16 we would need to figure out which CRC16 they are using here on the application layer and on which part of the data.
Should we make the effort to implement this additional check to be on the very safe side?

lib/drivers/medtronic/medtronicSimulator.js
DorianScholz: Same here, just wasn't sure if I should revert these changes on this branch, since Gerrit merged them in from master.

DorianScholz:
Thanks for the fast review!
I'm all for stricter linting! Would it be possible to add it to the whole uploader, or will that break things?
Thank you for your comments, I've fixed most issues.
Will wait for Gerrit to comment on the Medtronic changes.
Please let me know what you think about the application layer CRC16.

lib/drivers/abbott/freeStyleLibreData.js
pazaan: Do you have any example data sets and their corresponding CRCs? I think it's worth doing the extra validation if it's not too hard to figure out. The Medtronic drivers similarly have a one byte checksum, and several layers of CRC8 or CCITT checksums, all of which are validated.

pazaan:
We can't add stricter linting to the whole uploader, since most of it would fail, especially the older drivers. We'll just update the drivers piece-by-piece as we need to refactor.

lib/drivers/abbott/freeStyleLibreData.js
DorianScholz: I have plenty on example data. So I will give it a try tomorrow.

DorianScholz:
Ok, then I'll add the stricter linting to the abbott directory. But there are two old drivers in there as well. How can we exclude them?

pazaan:
Try using the `overrides` directive, and `extend` the base level `.eslintrc` for the older files.
See https://eslint.org/docs/user-guide/configuring#example-configuration.
In this way, any newer files that are added will use the new, stricter, config.

gniezen:
I've merged the relevant changes from master into `krystophv/electron`, and then merged that branch into this one to make sure all the Medtronic-related changes are removed.

DorianScholz:
Great, this was what I was hoping for instead of having to revert all medtronic files on this branch!

DorianScholz:
I added the stricter linting rules to the Libre modules.
But to be able to apply `overrides` for the old files in the same directory I had to upgrade ESLint to version 4.5.0.
I hope this is possible to merge?

But still I was not able to do what @pazaan suggested, as applying an `extends` inside and `overrides` is not allowed:
> Override blocks can contain any configuration options that are valid in a regular config, with the exception of extends, overrides, and root.

So, for the two old files I've reduced all relevant linting problems to `warn` level until they are fixed.
If you find a better way to deal with this, please let me know!

DorianScholz:
@gniezen the linting failes in api.js line 335, because the referenced `data` variable does not exist. I can't really see how to fix it properly without removing the output of the record.

lib/drivers/abbott/freeStyleLibreProtocol.js
```diff
@@ -0,0 +1,665 @@
...
+              } else {
+                debug('readResponse: Received wrong sequence number', atpPacket.sequenceTx,
+                  ', expected', this.sequenceRx);
+                // try to get missing package by sending an ACK containing the next expected
+                // sequence number to the device
+                this.sendAck();
+              }
+            } else if (hidPacket.responseType === COMMAND.ACK_FROM_DEVICE) {
+
+              // ignore these for now
```
pazaan: Should this be implemented? If not, can you please update the comment with what the expected behaviour should be when this case is not ignored, or why it's ignored (similarly to https://github.com/tidepool-org/chrome-uploader/pull/469/files#diff-9f2e95903da58a2a654b636f764330d6R202)?

DorianScholz: You're absolutely right, this comment should have been a to-do. It's is not necessary to deal with this ack response, as long as all command packet reach the device properly. But to recover from a lost or garbled command packet, the user will currently have to restart the upload process. To make this smoother I should at least throw an error on receiving an unexpected ack, or really handle the ack to resend the lost packet. But so far I've never had any lost or garbled packet in either direction and completely forgot about this...

DorianScholz:
So, what are the next steps to getting this merged?
(As the CRC16 and the ACK packets aren't blockers, I'll work on them next week.)

pazaan:
[[uploader] Abbott FreeStyle Libre](https://trello.com/c/XMchviCK/226-uploader-abbott-freestyle-libre)

pazaan:
Just push the "Merge pull request" button! :)
After that, you can delete this branch, and create a new branch from `krystophv/electron` for your CRC16 and ACK (and eventually when we move to `platform`).

pazaan:
My apologies. I got over-exited by the timing of the merge. Since `krystophv/electron` will go into master (and therefore production), we need to wait to merge this branch until testing is complete, and has been approved for production in Trello.

PR #471, opened 2017-08-23
Medtronic Direct: Bugfix for timestamps out of order.

### LINKED ISSUE #470: Medtronic Direct: Timestamps out of order
A support request (https://secure.helpscout.net/conversation/417340890/2392/) showed a "Timestamps must be in order" error message when uploading a Medtronic pump.

gniezen:
Fixes #470.

Usually bolus records and their associated wizard records have the same timestamp, but sometimes the bolus record occurs a couple of seconds later. If something else happens in between (like a low reservoir alarm), the wizard record will have an earlier timestamp than the current timestamp, so we cannot use `simpleSimulate`. We can, however, still ensure that the bolus timestamp is in sequence.

### DIFF
--- a/lib/drivers/medtronic/medtronicSimulator.js
+++ b/lib/drivers/medtronic/medtronicSimulator.js
@@ -551,7 +551,7 @@ exports.make = function(config){
           }
 
           currWizard = currWizard.done();
-          simpleSimulate(currWizard);
+          events.push(currWizard);
         } else {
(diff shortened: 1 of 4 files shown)


gniezen:
[#470 ⁃ Medtronic Direct: Timestamps out of order](https://trello.com/c/tFZ5mrMN/241-470-medtronic-direct-timestamps-out-of-order)

PR #472, opened 2017-08-24
Electron on MacOS: Use .dmg for installer and then install USB drivers from .app

gniezen:
What it says on the tin. 

This PR adds quite a few new dependencies. One that could maybe be replace is `decompress`, as we only need unzip functionality.

As the shell script runs as root, I would like to have at least two people review the code - @krystophv and @pazaan ?


app/utils/drivers.darwin.js
```diff
@@ -15,19 +15,75 @@
...
+  if (isDev) {
+    // The dev mode Electron.app does not contain the drivers or update script,
+    // it gets copied into the .app during packaging.
+    console.log('Not checking driver versions in dev mode.');
+    return;
+  }
+
+  const appFolder = path.dirname(remote.app.getAppPath());
+
+  decompress(path.join(appFolder,'driver/extensions.zip'), path.join(appFolder,'driver/')).then(files => {
```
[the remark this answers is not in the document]
gniezen: No, we don't. They get built and signed by the chipset manufacturers (e.g. Prolific, SiLabs). I agree keeping them in a .zip looks tidier. I got some more details on why the certificates became invalid. Electron-builder tries to sign everything in the .app, including the kexts. But you need a special certificate (only available on request) from Apple to be able to sign kexts.They'll be changing this behaviour to ignore the kexts in a future release, so for now this is the way to go.

gniezen:
Hahaha, @krystophv, you'll notice that bca3251 only has half a commit message. That's because macOS interpreted `$` as environment variables _inside_ the message. It should have said `"change $1 to $DRIVER_PATH"`. Next time I'll try to remember to escape them like you tried with the command-line options yesterday.

package.json
```diff
@@ -25,11 +25,11 @@
...
     "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development",
-    "postinstall": "concurrently \"electron-builder install-app-deps\" \"node node_modules/fbjs-scripts/node/check-dev-engines.js package.json\" && electron-rebuild --force --module-dir app",
+    "postinstall": "concurrently \"install-app-deps\" \"node node_modules/fbjs-scripts/node/check-dev-engines.js package.json\" && electron-rebuild --force --module-dir app",
     "dev": "npm run hot-server -- --start-hot",
     "prepare-qa-build": "node -r babel-register scripts/prepare-qa-build.js",
     "package": "npm run build && build -p onTagOrDraft",
-    "package-dev": "npm run build-dev && build -p onTagOrDraft --config.artifactName=\"${productName}-${version}-${arch}-${os}-dev.${ext}\" --config.nsis.artifactName=\"${productName}-Setup-${version}-dev.${ext}\" --extraMetadata.name=\"tidepool-uploader-dev\"",
+    "package-dev": "npm run build-dev && build -p onTagOrDraft --config.artifactName='${productName}-${version}-dev.${ext}' --config.nsis.artifactName='${productName}-Setup-${version}-dev.${ext}' --extraMetadata.name='tidepool-uploader-dev'",
```
krystophv: @gniezen - This was me simply reverting to what previously worked. The conflict was a combination of going back to `dmg` from `pkg` (because the artifact gets a proper extension in the interim build step and has a double period `..` in the file name). Technically this isn't great, since Windows will leave in the single-quotes as literals, but it doesn't die when trying to reference the temp files the way that Mac will if we switch out the escaping scheme. I'm only mostly sure that there's no way to properly escape the `$` character as a command line option value inside a JSON file that will be properly cross-platform. These options could be set in a separate config file by moving the `build` key out of `package.json` into two new files, one for the production build and one for the dev build, but the files would be largely duplicates of one another (except for the addition of these keys). I try to stick to "don't repeat yourself", but wanted your opinion concerning moving this hack to two config files instead (which will hopefully help us in efforts to upgrade later).

gniezen: Maybe just leave it as is for now and then we can move to a two-file config when we upgrade later?

gniezen:
I guess since it's part of this work you could do it as part of this PR? Up to you. I'm going to add the new .dmg background images to this PR as well.

PR #473, opened 2017-08-25
Fix TZOUtil test: use parseFromFormat instead of Date constructor

gniezen:
While working on the Electron app, @krystophv discovered that Chrome 58+ changes the way they handle `Date`s to be ES6 compliant. To quote http://codeofmatt.com/2015/06/17/javascript-date-parsing-changes-in-es6/:
 
> So, to make this align with the ISO 8601 spec, this line was changed in the ES6 specification (in section 20.3.1.16). It now says:
>
>... If the time zone offset is absent, the date-time is interpreted as a local time.
>
>That's great for compliance and spec alignment. But it's a serious issue for backward compatibility. Any code that expects the Date constructor to parse a date as UTC will need to be updated.

This has potential effects on our current production codebase.

This PR fixes a unit test for TZOUtil that calls the Date constructor with a timestamp without timezone info, so it got double-adjusted. I've gone through our uploader code and I can't find any instances of where we're trying to parse non-UTC timestamps apart from the unit test itself. We use `sundial.parseFromFormat` for non-UTC timestamps, which uses Moment instead to parse it into UTC. 

Other uses of  the Date constructor in our codebase are to get current date/time and to parse "UTC" Unix timestamps. We also use the Date constructor to parse dates where we calculate the difference between two dates, which is fine if they're both local instead of UTC.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "test": "mocha test/node/ && mocha test/node/**/*.js && ./node_modules/karma/bin/karma start",
     "browser-tests": "./node_modules/karma/bin/karma start --browsers PhantomJS,Chrome",
     "node-tests": "mocha test/node/ && mocha test/node/**/*.js",
-    "test-watch": "./node_modules/karma/bin/karma start --no-single-run",
+    "test-watch": "./node_modules/karma/bin/karma start --browsers PhantomJS,Chrome --no-single-run",
     "prestart": "rm -rf _book/",
     "start": "bash ./scripts/config.sh && webpack -d --progress --colors --watch",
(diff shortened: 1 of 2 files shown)


gniezen:
Ah, I see now I was using `yarn test-watch`, which only used phantomJS and not Chrome. I've now updated all calls to the Date constructor (except `new Date()` for current date/time) to use `sundial.parseFromFormat` instead, which fixes the tests.

krystophv:
Cool. I was actually copying the changes to the Electron branch and testing on Electron 1.7.x - the latest changes seem to resolve the issues. 

LGTM 👍

PR #474, opened 2017-08-28
Add Rollbar integration for evaluation

krystophv:
This adds Rollbar error tracking. They very recently introduced a telemetry feature which operates very similarly to the "breadcrumbs" from Sentry (I think). The more complicated pieces are handling the automated source-map uploading (thankfully someone wrote a decent webpack plugin for this) and re-writing the base of the URL so we can handle the fact that a full URI to the JS bundle is actually going to be a local filesystem path and not a standard URL.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1,3 +1,4 @@
+/* global  __VERSION_SHA__ */
 import _ from 'lodash';
 import React from 'react';
 import { render } from 'react-dom';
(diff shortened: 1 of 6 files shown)


krystophv:
[[Uploader] Send error reports automatically](https://trello.com/c/cYSNiB1W/196-uploader-send-error-reports-automatically)

app/index.js
```diff
@@ -12,6 +13,33 @@ import './app.global.css';
 import '../styles/main.less';
 import { ipcRenderer } from 'electron';
 import Raven from 'raven-js';
+import Rollbar from 'rollbar/dist/rollbar.umd';
+
+let rollbar = new Rollbar({
+    accessToken: '1843589282464f4facd43f794c8201a8',
```
gniezen: Should this be replaced with `ROLLBAR_POST_TOKEN`?

app/index.js
```diff
@@ -12,6 +13,33 @@ import './app.global.css';
...
+          }
+        }
+    },
+    transform: function(payload) {
+      var trace = payload.body.trace;
+      if (trace && trace.frames) {
+        for (var i = 0; i < trace.frames.length; i++) {
+          var filename = trace.frames[i].filename;
+          if (filename) {
+            trace.frames[i].filename = 'http://dynamichost/dist/bundle.js';
```
gniezen: Do you maybe want to add a comment here to explain how this URL handling works?

app/index.js
krystophv: This is actually the client-side token, which would end up distributed with JS code delivered to a browser environment and as such it's not as sensitive as the `ROLLBAR_POST_TOKEN`, so I didn't feel the need to go with ENV variables for it.

app/index.js
krystophv: Rather than try explain it myself, I could probably just make a small note and drop the link to the documentation about it (where this snippit came from)

app/index.js
gniezen: OK, got it, just checking ;)

app/index.js
gniezen: Sounds good to me.

PR #475, opened 2017-08-28
update help menus

krystophv:
Actually point help menu links to our resources. Also adds a Check for Updates link in appropriate places.

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -223,12 +223,6 @@ app.on('ready', async () => {
           shell.openExternal('http://support.tidepool.org/');
         }
       }, {
-        label: 'Check for Updates',
-        click() {
-          manualCheck = true;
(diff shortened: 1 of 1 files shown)


krystophv:
[Uploader 2.0 Help menu](https://trello.com/c/JfZWcESh/238-uploader-20-help-menu)

krystophv:
@gniezen - It is only in the `Tidepool Uploader` menu (on darwin). But there isn't one of those for Windows, so it seems that most Electron projects drop it under the `Help` menu for Windows.

krystophv:
@gniezen - Yup, the `Help` menu update check should be (and was supposed to be - my bad) Windows only, as there isn't a `Tidepool Uploader` menu item on Windows. I moved it into the appropriate block.

PR #506, opened 2017-08-30
add UI for driver updates

krystophv:
Creates a UI component very similar to the electron update notification for the driver updates. Moves the actual install script execution behind a user click event.

There are a couple slightly strange things happening because we're trying to dispatch Redux events from a non-component module. We generally wouldn't pass around the `dispatch` function, but it was the best way I could figure out how to expose it to the `drivers` file. Ideally, we wouldn't be passing the shell options in application state, but I couldn't get access to the raw `store` to pass to the `drivers` file (in order to manually `subscribe` to the user-initiated install action), so the shell script execution gets moved into the component that actually handles the click event.


krystophv:
@gniezen I think this is ready for review now.

app/components/UpdateDriverModal.js
```diff
@@ -0,0 +1,114 @@
...
+    if(driverUpdateAvailableDismissed || driverUpdateComplete || !driverUpdateAvailable){
+      return null;
+    }
+
+    if (checkingDriverUpdate){
+      title = 'Checking for driver update...';
+    } else {
+      if (driverUpdateAvailable) {
+        title = 'Driver Update Available!';
+        text = 'After clicking Install, the uploader ask for your password to complete the installation. This window will close when completed.';
```
gniezen: the uploader _will_ ask / the uploader _asks_ ?

PR #507, opened 2017-08-31
Fix regression: clearing the suspending event after use

gniezen:
An issue reported by a user was reproduced using Eric's 530G. Using Eric's binary blob, I went back through the versions and it is a regression that was introduced in the most recent version: v0.309.0.

Basically, if a "suspending event", like an auto-off alarm occurred, there's a chance it may have been used to generate a suspended basal more than once, which was then caught by the platform validation. This PR makes sure we always clear the suspending event after use.

### DIFF
--- a/lib/drivers/medtronic/medtronicSimulator.js
+++ b/lib/drivers/medtronic/medtronicSimulator.js
@@ -68,8 +68,10 @@ exports.make = function(config){
   };
 
   function setSuspendingEvent(event) {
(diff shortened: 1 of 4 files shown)


gniezen:
[[uploader][bug] #470 ⁃ Medtronic Direct: Timestamps out of order](https://trello.com/c/tFZ5mrMN/241-uploaderbug-470-medtronic-direct-timestamps-out-of-order)

PR #510, opened 2017-09-01
Enable Tandem, Dexcom, Contour Next USB, and Omnipod on ChromeOS

willnowak:
Adds cros to the enabled map for these devices, and Supports serial ports named /dev/ttyACM.+


gniezen:
Hi @willnowak ! This is pretty cool, thanks for your PR! Unfortunately I have a bit of bad news - the release of Uploader 2.0 is imminent, and is based on Electron and not a Chrome App anymore. We made this decision last year after Google announced that they will be discontinuing Chrome Apps on all platforms except for Chrome OS. To my knowledge it's not possible to run an Electron app on ChromeOS - can you confirm that this is indeed the case?

PR #512, opened 2017-09-04
Support for Dexcom G5 CGM events on Tandem X2

gniezen:
Tandem defined new record types for Dexcom G5 events and this PR adds support for the new CBG and calibration events.

This PR also changes multi-record downloads to only work with firmware version 59814 and higher, to conform with updates to the protocol spec.

Addresses #511.


gniezen:
[#511 ⁃ [uploader] Add G5 support for Tandem pumps](https://trello.com/c/aKu0nzdf/246-511-uploader-add-g5-support-for-tandem-pumps)

lib/drivers/tandem/tandemDriver.js
```diff
@@ -2114,15 +2131,27 @@ module.exports = function (config) {
...
+        record = cfg.builder.makeDeviceEventCalibration();
+        record.with_value(value)
+              .with_deviceTime(event.deviceTime)
+              .with_units('mg/dL')      // hard-coded in mg/dL
+              .set('index', event.index);
+        cfg.tzoUtil.fillInUTCInfo(record, event.jsDate);
+        record = record.done();
+
+      } else if ((event.header_id === PUMP_LOG_RECORDS.LID_CGM_DATA.value ||
+                  (event.header_id === PUMP_LOG_RECORDS.LID_CGM_DATA_GX.value &&
```
pazaan: [suggestion] Formatting is a bit weird here...

lib/drivers/tandem/tandemDriver.js
```diff
@@ -599,6 +599,18 @@ module.exports = function (config) {
...
+      value: 0xD3,
+      name: 'CGM GX Data Sample',
+      format: 'sby.ysii',
+      fields: ['status','type','rate','rssi','value','timestamp', 'transmitter_timestamp']
+    },
+    LID_CGM_CALIBRATION_GX: {
+      value: 0xD2,
+      name: 'CGM Calibration GX',
+      format: 's..............',
+      fields: ['value']
```
pazaan: Do we want to add trailing commas for all lines? You have a trailing comma at line 2191, so just pointing it out for consistency.

gniezen: Ha, good point. I thought to myself "I should start getting into the habit of adding trailing commas. But maybe I should rather start doing it with new drivers than adding it to older drivers to be consistent. Adding trailing commas for all lines is probably more trouble than it's worth - I'll remove the trailing comma.

PR #515, opened 2017-09-07
Win32 chrome-finder fix

krystophv:
Right now, `chrome-finder` (part of the `chrome-launcher` being sliced off of Google's Lighthouse project) presumes the existence of a `PROGRAMFILES(X86)` environment variable which only exists on Windows x64 versions and tries to use it to create a path, which fails since it's `undefined` instead of a string.

This works around it by setting the env variable if it doesn't exist.


gniezen:
Did some preliminary testing after switching dev.3 to pre-release, works great!

krystophv:
#518

krystophv:
[#518 ⁃ [uploader] Windows 32 crashes when clicking on most links (like the footer)](https://trello.com/c/FFKssPse/252-518-uploader-windows-32-crashes-when-clicking-on-most-links-like-the-footer)

PR #525, opened 2017-09-13
Unify build

krystophv:
This PR covers a handful of changes. 

- Moving from the psudo-global static `__DEBUG__` to a toggle-able `DebugMode` setting allowing us to collapse the `dev` and `prod` builds into a single build
- Adding a proper checkmark next to the currently selected server in the context selection menu on the login screen
- Removing `redux-logger` in favor of the much more complete Redux Devtools available in `development` mode


app/utils/debugMode.js
```diff
@@ -0,0 +1,50 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import _ from 'lodash';
```
gniezen: We're not using lodash here, are we?

app/containers/App.js
```diff
@@ -39,6 +39,7 @@ const syncActions = actions.sync;
 import * as actionSources from '../constants/actionSources';
 import { pages, urls, pagesMap } from '../constants/otherConstants';
 import { checkVersion } from '../utils/drivers';
+import DebugMode from '../utils/debugMode';
```
gniezen: If I'm super nit-picky, I would not capitalize `debugMode` to be consistent. But I don't have strong feelings about it.

krystophv: You're right. It doesn't require the `new` keyword, so it probably shouldn't be initial cap.

app/utils/debugMode.js
krystophv: You're correct. I think I was initially, but factored it out.

krystophv:
[[Uploader] Unify dev/prod build](https://trello.com/c/jo0EfNg2/257-uploader-unify-dev-prod-build)

PR #529, opened 2017-10-03
Bugfix: Pump errors lead to unexpected data

gniezen:
This PR includes the following bugfixes:

- fixes the new debug mode so that it works with CLI tools like `blob_loader`
- checks that there is a basal event on the pump before trying to set a suspending event
- handles cleared basal schedules after a pump error


gniezen:
[Pump errors cause unexpected data on pump](https://trello.com/c/TWzhnHRM/264-pump-errors-cause-unexpected-data-on-pump)

PR #530, opened 2017-10-04
Script to submit release to antivirus whitelist

gniezen:
What it says on the tin. At the moment we'll have to run the script manually after every release. In future we may want to consider a cron job or a webhook to automate this.


krystophv:
As a more general note, it looks like Rollbar has the ability to track deploys, but requires that one hit an API endpoint on release. Since our release mechanism is a GitHub publish, there's no existing script that gets run to piggyback off of easily. If we end up needing two scripts (this one and the Rollbar deploy tracking), I wonder if it'd be worth investigating writing a small script that we host somewhere to accept the GitHub webhook for releases which will automatically execute this and potentially other scripts.

gniezen:
@krystophv I've addressed your comments. Are you happy for me to just merge this, as it contains no user-facing changes and I'm already using it to submit to antivirus whitelist?

krystophv:
@gniezen Yeah, feel free to merge.

PR #535, opened 2017-10-20
Bugfix: Precision Xtra

gniezen:
One-liner to fix a regression that prevents Precision Xtra from uploading.


gniezen:
[Precision Xtra upload fails](https://trello.com/c/PZzb3Ch2/267-precision-xtra-upload-fails)

PR #536, opened 2017-10-24
Handle bogus time change events

pazaan:
* If an NGPHistoryEvent.EVENT_TYPE.TIME_RESET event is found, skip the next NGPHistoryEvent.EVENT_TYPE.USER_TIME_DATE_CHANGE. This can occur if the pump runs out of battery, and the user sets the correct time when a new battery is installed. It is NOT a valid time change due to crossing timezones, clock drift, or DST changes
* Formatting of TODOs to make plugins happy
* Add parsing of Alarm events

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -160,7 +160,7 @@ class NGPHistoryEvent {
     return this.eventData[0];
   }
 
-  // TODO - make this a static factory (passing in eventData).
+  // TODO: - make this a static factory (passing in eventData).
   // Doing it this way causes double-initialisation
   eventInstance() {
     /* eslint-disable no-use-before-define */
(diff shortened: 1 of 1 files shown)


lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -231,12 +235,36 @@ class NGPHistoryEvent {
...
 }
 
-class UserTimeDateEvent extends NGPHistoryEvent {
+class TimeChangeEvent extends NGPHistoryEvent {
   get newTimestamp() {
     return NGPUtil.NGPTimestamp.fromBuffer(this.eventData.slice(0x0B, 0x13));
   }
 }
 
+class AlarmNotificationEvent extends NGPHistoryEvent {
```
gniezen: Do you actually make use of these alarm notification events anywhere?

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -798,10 +826,23 @@ class NGPHistoryParser {
   }
 
   processPages(pages) {
+    let skipTimeChange = false;
     // Consume the pages as we process them to save on memory.
     while (pages.length > 0) {
       const page = pages.shift();
       for (const event of NGPHistoryParser.eventsFromPageString(page)) {
+        if (event.eventType === NGPHistoryEvent.EVENT_TYPE.TIME_RESET) {
+          debug('TIME_RESET detected. Ignoring next USER_TIME_DATE_CHANGE.');
```
gniezen: Do we also need to ignore other events/alarms between `TIME_RESET` and `USER_TIME_DATE_CHANGE`? Or do we want to still record them even if their datetime is incorrect? Maybe we should annotate them?

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Not yet, but I have an outstanding card to support specific alarms (https://trello.com/c/PZfoCKuP). I did this work as part of the debugging to track down how best to solve this, but I've left it in so that I can complete that card later.

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: There are other events in there that we don't process currently (DAILY_TOTALS, other ALARMS). Currently I'm effectively ignoring them by not processing them. Perhaps I should explicitly ignore them by not pushing them. Thoughts?

gniezen: Let's make that decision when we start to process them.

PR #537, opened 2017-10-24
Script to re-create binary blob from fragments in console log

gniezen:
Sometimes we have a console log from a user, but no binary blob. This scripts takes the fragments of binary data in the console log and reconstructs the binary blob.


gniezen:
I'm going to merge this one into master as it's purely a development tool and does not have any user-facing changes.

PR #539, opened 2017-10-25
Upload data services port for platform

darinkrauss:
@gniezen @pazaan The localhost ports are changing for the platform to be inline with the deployed ports for less confusion as we add additional platform services.

Should not be merged into `master` until platform is merged into `master`.


darinkrauss:
[[platform] Download data from Dexcom API via OAuth](https://trello.com/c/MmE0hoUx/259-platform-download-data-from-dexcom-api-via-oauth)

PR #543, opened 2017-11-05
Allow linux host in Animas device config

Zanfib:
Tested on tag v2.2.2 of master tree with Gentoo Linux 64 bits up to date and the Animas Vibe pump


lib/core/device.js
```diff
@@ -365,6 +365,8 @@ device.detectUsb = function(driverId, cb) {
                found = true;
              } else if (p.comName.match('/dev/tty.+')) {
                found = true;
+             } else if (p.comName.match('/dev/ttyUSB.+')) {
```
gniezen: I don't think this is necessary anymore, as the previous `tty.+` should match ttyUSB* as well?

Zanfib: the '.+' is one character catch only, maybe with 'tty*' in the second if but I've not changed for this version because I will not touch on a code used for Mac platform (1 test for each platform I guess ?)

Zanfib: How sorry, my regular expression's are a little bit false, I've tested without this line before and never work, I will retest with last version of the code

gniezen: Yeah, we used to match on `/dev/cu.+` for macOS, but this changed to `/dev/tty.+` (very recently) when we updated to a more recent version of the `node-serialport` library to make it work the same way as Linux.

Zanfib: I see during november ! A good update with the node-serialport version 6.0.4 that solve an issue with linux too. I just tested the revert and the animas driver works well with just a oneliner to add linux platform in driver config, so my PR go to be useless ! :-)

gniezen:
Could you please mention the Linux distro and version number you tested this with, as well as which Animas pump you used?

Zanfib:
Yes, updated in first comment, Gentoo Linux (64 bits rolling release) and Animas Vibe pump.

derrickburns:
[Suppress noisy platform log messages.](https://trello.com/c/YbawaBoZ/643-suppress-noisy-platform-log-messages)

gniezen:
Closing this PR, as `linux: true` has been the case for Animas for a while now.

PR #544, opened 2017-11-06
Bugfix for Error: Suspend/resume events out of order

gniezen:
For more details, see https://trello.com/c/vKfxm83B


gniezen:
[[uploader] Error: Suspend/resume events out of order](https://trello.com/c/vKfxm83B/271-uploader-error-suspend-resume-events-out-of-order)

PR #545, opened 2017-11-06
Updating native dependencies

gniezen:
node-serialport is the only native dependency with a major version update that has breaking changes. Luckily the changes are quite small, but will require smoke testing all supported devices on at least one version of each OS.

Since both node-serialport and node-hid now has prebuilt binaries for Electron, I'm also seeing if we can do away with rebuilding native dependencies ourselves.


gniezen:
@krystophv Can you review the changes I've made to use the prebuilt binaries for the native dependencies? I'm not 100% sure if I did it correctly. Uploads are successful.

@pazaan Will you please review the other small changes I had to make to get node-serialport from v4 to v6?

gniezen:
@krystophv Ah, I was wondering how that worked.

PR #546, opened 2017-11-06
fix hanging uploader when device not connected

DorianScholz:
Abbott FreeStyle Lite driver is missing an error check in fetchData to detect when the device is not connected to the USB serial cable.
Currently the uploader just hangs forever at 70%...


gniezen:
[FreeStyle Lite upload sometimes hangs at 70%](https://trello.com/c/3lCJF3qM/33-freestyle-lite-upload-sometimes-hangs-at-70)

PR #547, opened 2017-11-06
return on callback with error

krystophv:
See Trello/Rollbar for more information.

Makes sure we return from function when throwing the callback with an error instead of continuing function execution (and adding erroneous profiles to available upload groups).


krystophv:
[Errored user profile still being added to available upload groups](https://trello.com/c/LMqqMx1x/31-errored-user-profile-still-being-added-to-available-upload-groups)

PR #548, opened 2017-11-07
Support for OneTouch Verio & Verio Flex

DorianScholz:
Support for the Verio so far only on OSX and Linux.

On OSX this needs the "disk permission helper" which is installed as a LaunchDeamon and a PrivilegedHelperTool using the driver update mechanism.
This allows using direct disk access to communicate with the device via USB SCSI commands.

On Linux direct USB communication is used after detaching the kernel USB SCSI driver.

Windows support will have to wait until Node 8 is used and has all the patches needed for direct disk access via the fs module.

The Verio gives us no time change records, so timezone is applied across the board.
But a global record number exists, which is used as the index.


DorianScholz:
Well, the Windows build is failing to compile the new "usb" dependency. Any ideas how to change this, or how to disable it on windows for now, as it is not needed there at the moment?

pazaan:
@DorianScholz I'm less concerned about that one, and more with the failure in CircleCI.

DorianScholz:
oops, I somehow missed that one.
Well there now is one device not available on windows, so the test criteria had to change...

DorianScholz:
So the windows build issue of node-usb seems to have been fixed in June in release 1.3.0, but the version on npm is still 1.2.0:
https://github.com/tessel/node-usb/issues/162
Should we pull in 1.3.0 from the github release as described in this issue?

pazaan:
We can give it a go and see if it helps. @gniezen?

gniezen:
node-usb 1.3.0 has been marked as pre-release since June, and the maintainer does not seem to be responding to requests, see e.g. https://github.com/tessel/node-usb/issues/193. This is a bit concerning. 

That said, using a pre-release of node-usb seems to be a less riskier option than using a pre-release of Electron, which are our two options here, right?

DorianScholz:
With the dependency set to 1.3.0 from git we seem to be running into the problem that this git repo is not pulled in recursively by npm and so the sub-repo with libusb is empty as described for example here:
https://github.com/tessel/node-usb/issues/150

But I have no idea how to fix this for a dependency.

gniezen:
Note: I tried using prebuilt binaries, but node-usb only has these for Node 7 & 8, not for Electron, so we have to build from source.

gniezen:
One solution could be to switch back to npm for building, as not pulling in git submodules is an issue with yarn: https://github.com/yarnpkg/yarn/issues/1488. I've tried using npm 4 locally and it seems to work fine.

gniezen:
I think I have finally found a solution: Having Appveyor build with VS2013 instead of VS2017 allows us to use node-usb v1.2.0.

gniezen:
cross post from Slack: I just noticed that node-usb v1.3.1 was pushed to npm. It seems to work fine with uploading Verio & Verio Flex to Linux, so I'm going to bump Visual Studio from 2013 to 2017 to see if Appveyor builds correctly with this version.

DorianScholz:
@gniezen Your changes look good to me, please go ahead with this.

lib/core/device.js
```diff
@@ -416,6 +441,30 @@ device.detect = function(driverId, options, cb) {
         return cb();
       }
 
+      self._deviceInfoCache[driverId] = _.cloneDeep(devdata);
+
+      self.detectHelper(driverId, options, function(err, usbDevice) {
+
```
pazaan: If we're moving towards `eslint-config-airbnb`, there shouldn't be an empty line after an opening brace here.

lib/core/device.js
```diff
@@ -416,6 +441,30 @@ device.detect = function(driverId, options, cb) {
...
+        }
+        device = _.assign(devdata, usbDevice);
+        return cb(null, devdata);
+      });
+    });
+
+  } else {
+
+    this.detectUsbSerial(driverId, function(err, devdata) {
+
```
pazaan: See above.

lib/core/device.js
```diff
@@ -416,6 +441,30 @@ device.detect = function(driverId, options, cb) {
...
+        if (err) {
+          return cb(err);
+        }
+        device = _.assign(devdata, usbDevice);
+        return cb(null, devdata);
+      });
+    });
+
+  } else {
+
```
pazaan: See above.

test/app/reducers/misc.test.js
```diff
@@ -69,8 +69,8 @@ describe('misc reducers', () => {
       });
       let expectedResult = _.pickBy(devices, filterDevicesFn('win'));
       expect(actualResult).to.deep.equal(expectedResult);
-      // because nothing currently is unavailable on Windows
-      expect(_.keys(actualResult).length).to.equal(_.keys(devices).length);
+      // because we do currently have devices unavailable on Windows
```
pazaan: Can you please clarify this comment? I'm not exactly sure what it's trying to say.

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -0,0 +1,730 @@
...
+  timeFormat: 0x00,
+  dateFormat: 0x02,
+  displayUnit: 0x04,
+};
+const PARAMETER_NAME = invert(PARAMETER_TYPE);
+const UNIT_OF_MEASURE = [
+  'mg/dL',
+  'mmol/L',
+];
+const TIMESTAMP_EPOCH = 946684800;
```
pazaan: Can we add a comment after this? Something like `// January 1, 2000 12:00:00 AM GMT`.

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -0,0 +1,730 @@
...
+  retrieveData(lbaNumber, requestData, callback) {
+    const seekOffset = lbaNumber * BLOCKDEVICE_BLOCKSIZE;
+    fs.write(
+      this.fileHandle, requestData, 0, BLOCKDEVICE_BLOCKSIZE, seekOffset,
+      (err) => {
+        if (err) {
+          debug(`retrieveData: Error writing to device '${this.devicePath}':`, err);
+          return;
+        }
+        // debug(`retrieveData: wrote ${numWritten} bytes:`, requestData);
```
pazaan: Remove commented code if it's no longer required.

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -0,0 +1,730 @@
...
+        const readBuffer = Buffer.alloc(BLOCKDEVICE_BLOCKSIZE);
+        fs.read(
+          this.fileHandle, readBuffer, 0, BLOCKDEVICE_BLOCKSIZE, seekOffset,
+          (err2) => {
+            if (err2) {
+              debug(`retrieveData: Error reading from device '${this.devicePath}':`, err2);
+              return;
+            }
+            // debug(`retrieveData: read ${numRead} bytes:`, readBuffer);
```
pazaan: Remove commented code if it's not required.

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -0,0 +1,730 @@
...
+  closeDevice(callback) {
+    return this.communication.closeDevice(callback);
+  }
+
+  checkDevice(callback) {
+    return this.communication.checkDevice(callback);
+  }
+
+  static parseApplicationLayer(data, callback) {
+    // const serviceId = data[0];
```
pazaan: Remove commented code

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -0,0 +1,730 @@
...
+      this.constructor.parseLinkLayer(linkResponseData, (applicationResponseData) => {
+        this.constructor.parseApplicationLayer(applicationResponseData, (commandData) => {
+          callback(commandData);
+        });
+      });
+    });
+  }
+
+  static buildLinkLayerFrame(applicationData) {
+    const length = 3 + applicationData.length + 3;
```
pazaan: Please add comments as to what the `3`'s mean, or add a const somewhere (eg `const MESSAGE_ENVELOPE = 3;`) that makes the meaning explicit.

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -0,0 +1,730 @@
...
+  retrieveParameterData(parameterType, callback) {
+    const linkRequestData = Buffer.alloc(USB_BULK_BLOCKSIZE);
+    const applicationData = Buffer.from([SERVICE_ID, parameterType, 0x00]);
+    this.constructor.buildLinkLayerFrame(applicationData).copy(linkRequestData);
+    this.retrieveData(LBA_NUMBER.PARAMETER, linkRequestData, (commandData) => {
+      let responseString;
+      if (parameterType === PARAMETER_TYPE.displayUnit) {
+        responseString = UNIT_OF_MEASURE[commandData.readUInt32LE()];
+      } else {
+        responseString = commandData.slice(0, commandData.length - 2).toString('utf16le');
```
pazaan: Comment or const for `2` here :)

lib/core/device.js
gniezen: `lib/core/device.js` is still ES5, so doesn't yet adhere to the new linting rules. We should probably come up with a strategy on how/if we want to migrate them over?

lib/drivers/onetouch/oneTouchVerio.js
gniezen: I'll leave it to @DorianScholz to define these constants, as he probably has a better idea what they should be :)

gniezen: Ping @DorianScholz

lib/drivers/onetouch/oneTouchVerio.js
gniezen: Ping @DorianScholz

gniezen:
[#520 ⁃ [uploader] OneTouch Verio and Verio Flex](https://trello.com/c/UhxOg1zZ/253-520-uploader-onetouch-verio-and-verio-flex)

test/app/reducers/misc.test.js
DorianScholz: Well, to be honest, this sentence is not original work, but a simple variation of the comment from line 61 stating the same thing for Mac instead of Windows ;-)
It is supposed to explain why the next line tests for the number of available devices to be less than the the number of all devices.
Do you have any proposition how this should be better explained?
Then I could change both sentences accordingly.

gniezen: Maybe "at least one device is unavailable on Windows, so available devices should be less than all devices"?

DorianScholz:
@pazaan Thanks for pointing out the magic numbers! I replaced them and improved the comments you mentioned.

pazaan:
@DorianScholz - nice! Looks good 😄

gniezen:
Just a note that it will not upload to platform successfully until https://trello.com/c/fD8JYbuZ/316-backend-add-verio-verio-flex-to-list-of-supported-devices is implemented.

gniezen:
Note: I have tried accessing the device on Windows with Node v8.2.1 (Electron v1.8.2). It does not work, so for Windows we'll have to wait for the next release of Electron that included Node v9.

PR #550, opened 2017-11-11
Medtronic 6-series auto basals

pazaan:
* Generate basal pattern from 6-series APS micro boluses
* Add support for Meal Wizard type
* Fix linting errors


lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1420,17 +1503,17 @@ class NGPHistoryParser {
...
       .with_carbInput(wizardEvent.carbInput)
       .with_insulinOnBoard(wizardEvent.iob)
       .with_insulinCarbRatio(wizardEvent.carbRatio)
       .with_insulinSensitivity(wizardEvent.isf)
-      .with_bgTarget({
+      .with_bgTarget(wizardEvent.eventType === NGPHistoryEvent.EVENT_TYPE.BOLUS_WIZARD_ESTIMATE ? {
         low: wizardEvent.lowBgTarget,
         high: wizardEvent.highBgTarget,
-      })
+      } : undefined)
```
gniezen: I'd rather we wrap this in an if statement, e.g.,

```
if (wizardEvent.eventType === NGPHistoryEvent.EVENT_TYPE.BOLUS_WIZARD_ESTIMATE) {
  wizard.with_bgTarget({
    low: wizardEvent.lowBgTarget,
    high: wizardEvent.highBgTarget,
  })
}
```
Otherwise it will send up a `bgTarget : undefined` property, when it should not be sending anything.

lib/drivers/medtronic600/medtronic600Simulator.js
```diff
@@ -128,6 +128,39 @@ class Medtronic600Simulator {
...
+      if (this.currBasal.scheduleName === 'Auto-Basal') {
+        const autoBasalDifference = Date.parse(event.time).valueOf() -
+          Date.parse(this.currBasal.time).valueOf();
+        // Auto-basal microboluses occur roughly every 5 minutes, but no more than 6 minutes apart.
+        // The pump doesn't tell us if the algorithm chose *not* to micro-bolus.
+        // If a 6 minute gap appears between micro-boluses, there must have been a gap.
+        if (autoBasalDifference > 360000) {
+          this.currBasal
+            .set('duration', 300000)
+            .set('basalEndTime', new Date(Date.parse(this.currBasal.time).valueOf() + 300000).toISOString());
```
gniezen: Using the Date constructor to parse dates is dangerous, as browser behaviour can change, as it recently did in Chrome 58+. Rather use `sundial.parseFromFormat`. For more details, see https://github.com/tidepool-org/chrome-uploader/pull/473

lib/drivers/medtronic600/medtronic600Simulator.js
```diff
@@ -128,6 +128,39 @@ class Medtronic600Simulator {
...
+            .with_time(insertedBasalDate.toISOString())
+            .with_deviceTime(sundial.formatDeviceTime(
+              sundial.parseFromFormat(this.currBasal.deviceTime).valueOf() + 300000))
+            .with_rate(0)
+            .with_scheduleName('Auto-Basal')
+            .with_duration(Date.parse(event.time) - Date.parse(this.currBasal.time) - 300000)
+            .with_conversionOffset(this.currBasal.conversionOffset)
+            .with_clockDriftOffset(this.currBasal.clockDriftOffset)
+            .with_timezoneOffset(this.currBasal.timezoneOffset)
+            .set('index', currBasalIndex + 300000)
```
gniezen: Why can't we use TZOUtil to set the offsets here? And if we're not using TZOUtil, why are we setting `index` and `jsDate`?

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -501,6 +501,15 @@ class BolusDeliveredEvent extends NGPHistoryEvent {
   get presetBolusNumber() {
     return this.eventData[0x0D];
   }
+
+  bolusFromWizard() {
```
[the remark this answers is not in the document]
pazaan: No. It's not a direct property. I'm using it as a function, since it's derived. At least that's how I've distinguished them in the past.

PR #551, opened 2017-11-14
Verio Flex Support

gniezen:
This PR adds Verio Flex support to the Verio driver. It also moves the `drivelist` dependency to `app/package.json` as it's a native dependency. This has the weird side effect of also having to add `lodash` to `app/package.json`.

Basically, if I read https://www.electron.build/tutorials/two-package-structure correctly, we need to move all `dependencies` to `app/package.json`, and only have `devDependencies` in `package.json`.


gniezen:
I was looking at the Verio & Verio Flex checklist and noticed that we don't yet discard control solution tests. We'll need to fix that before we push to production. I'll see if I can get it done.

gniezen:
@pazaan Can you just check the two commits since your review - I've added the check to discard control solution tests.

gniezen:
[#520 ⁃ [uploader] OneTouch Verio and Verio Flex](https://trello.com/c/UhxOg1zZ/253-520-uploader-onetouch-verio-and-verio-flex)

PR #552, opened 2017-11-27
Remove all references to chrome-uploader

gniezen:
What it says on the tin. This can only be merged once we've switched the repo name from `chrome-uploader` to just `uploader`.


gniezen:
Closing this in favour of #762 as it is created against develop.

PR #555, opened 2017-12-07
Improve Precision Xtra error handling and increase timeout

gniezen:
What it says on the tin. For more details, see https://trello.com/c/NRFZHzbo


gniezen:
[[uploader] PXtra times out if there are many BG records on the meter](https://trello.com/c/NRFZHzbo/295-uploader-pxtra-times-out-if-there-are-many-bg-records-on-the-meter)

PR #556, opened 2017-12-08
Delta uploads

gniezen:
Adding delta upload support for both Medtronic Direct 5/7-series (https://trello.com/c/5i0sM0wx) and Tandem (https://trello.com/c/ieSwFQq5)

Dependent on https://github.com/tidepool-org/platform-client/pull/84/

Note that we still need to implement some UI popups for delta uploads, specifically when it's the first upload that can take a long time.


gniezen:
[<M> Delta Uploads (4 of 4) - Tandem](https://trello.com/c/ieSwFQq5/120-m-delta-uploads-4-of-4-tandem)

gniezen:
[<M> Delta Uploads (3 of 4) - Medtronic 5-series](https://trello.com/c/5i0sM0wx/118-m-delta-uploads-3-of-4-medtronic-5-series)

lib/drivers/medtronic/medtronicDriver.js
```diff
@@ -1120,15 +1134,53 @@ module.exports = function (config) {
...
+              lastCBGPosition : result.currentCBGPosition,
+              lastGlucosePosition : result.currentGlucosePosition,
+              lastIsigPosition : result.currentIsigPosition
+            }}); // will form part of upload record
+
+            var currentPage = positionToPage(lastCBGPosition, result.currentCBGPosition, 'CBG history');
+            var glucosePage = positionToPage(lastGlucosePosition, result.currentGlucosePosition, 'glucose');
+            var isigPage = positionToPage(lastIsigPosition, result.currentIsigPosition, 'isig');
+
+            callback(null, { currentPage : currentPage, glucosePage: Math.min(glucosePage, MAX_CBG_PAGES) + 1, isigPage: Math.min(isigPage, MAX_CBG_PAGES) + 1 });
```
pazaan: The original code doesn't `+1` to the result of the `Math.min` function, unless this is an additional bug fix.

lib/drivers/tandem/tandemDriver.js
```diff
@@ -1482,11 +1272,33 @@ module.exports = function (config) {
...
+            if (lastEndPosition != null) {
+              debug('Last record read was', lastEndPosition, ', starting from there');
+              data.start_seq = lastEndPosition;
+            } else {
+              debug('Reading from record', result.payload['start_seq']);
+              data.start_seq = result.payload['start_seq'];
+            }
+
+            debug('Reading up until record', data.end_seq);
+            //newestEventRequester(end_seq, start_seq, progress, getNewestEvent);
```
pazaan: Remove commented code?

lib/drivers/medtronic/medtronicDriver.js
gniezen: Good catch, but yes, it's an additional bug fix. Basically it's so that we always retrieve the most recent page, even if we already retrieved it in the past, as there may be new records on that page.

gniezen:
Pinging @pazaan for final review approval.

lib/drivers/medtronic/medtronicDriver.js
```diff
@@ -25,6 +25,7 @@ var common = require('../../commonFunctions');
 var medtronicSimulator = require('./medtronicSimulator');
 var crcCalculator = require('../../crc.js');
 var debugMode = require('../../../app/utils/debugMode');
+var api = require('../../core/api.js');
```
pazaan: @gniezen do we need this? Isn't this the same as cfg.api?

lib/drivers/medtronic/medtronicDriver.js
gniezen: I wonder if this is what is causing the issue we're seeing. I'm removing the `+1` here so that it's the same as the original code again.

lib/drivers/medtronic/medtronicDriver.js
gniezen: Yep, you're right, I'll remove it.

PR #559, opened 2017-12-14
Support for Abbott FreeStyle Libre Pro

DorianScholz:
Changes with respect to original Libre (non-Pro) code:
- changed what data is requested from device, as the Pro does not have all the same data tables
- more relaxed treatment of "oldest valied record" number, since less records exist in the Pro due to the missing user interaction with the device
- updating of progress bar during device communication for better UX (also relevant for non-Pro)


lib/drivers/abbott/abbottFreeStyleLibre.js
```diff
@@ -93,32 +98,59 @@ export default function (config) {
...
+          (callback) => { protocol.getCfgData(CFG_TABLE_ID.METER_SETTINGS, callback); },
+          (callback) => { protocol.getCfgData(CFG_TABLE_ID.USER_PATIENT_CONFIGURATION, callback); },
+          (callback) => { protocol.getCfgData(CFG_TABLE_ID.USER_PATIENT_SETTINGS, callback); },
+          (callback) => { protocol.getDatabase(DB_TABLE_ID.GLUCOSE_RESULT, callback); },
+          (callback) => { protocol.getDatabase(DB_TABLE_ID.HISTORICAL_DATA, callback); },
+          (callback) => { protocol.getDatabase(DB_TABLE_ID.EVENT, callback); },
+        ];
+      } else {
+        getterFunctions = [
+          (callback) => { protocol.setCompression(COMPRESSION.ENABLED, callback); },
```
gniezen: Are we sure that there is no compression possible for Libre Pro?

package.json
```diff
@@ -198,7 +198,7 @@
     "eslint-import-resolver-webpack": "0.8.3",
     "eslint-plugin-import": "2.8.0",
     "eslint-plugin-jsx-a11y": "6.0.2",
-    "eslint-plugin-lodash": "2.4.3",
+    "eslint-plugin-lodash": "2.5.0",
```
gniezen: Was this dependency update required for anything? I'm happy for eslint plugins to be updated as part of this PR, just interested to know why it was done ;)

gniezen:
[Libre Pro](https://trello.com/c/2sVdsOOU/305-libre-pro)

package.json
DorianScholz: yarn complained about it being a different version from the one in app/package.json...

lib/drivers/abbott/abbottFreeStyleLibre.js
DorianScholz: that's where the upload got stuck with a timeout. 
the Pro does not answer to this command at all.
interestingly the Pro has an older firmware version (1.x) than the non-Pro (2.x)...
but the compression does not make a big difference for the pro anyhow, since there are mostly just historical glucose data sets on it, which don't include to many zeros and the compression is a very basic run-length encoding just for zeros...

gniezen: :+1:

DorianScholz:
@gniezen Could you have a look at why the tests are failing?
I'm not quite sure what the problems are and how to solve them.

gniezen:
Yep, I suspect it's just a temporary blip with GitHub, going to rebuild now and see what happens.

DorianScholz:
@gniezen could you review my latest commits and maybe also comment on the general approach of generating a unique sensor id based on it's first historical data point. It seems to do what we want, just want to make sure I did not miss anything...

lib/drivers/abbott/abbottFreeStyleLibre.js
```diff
@@ -169,7 +172,24 @@ export default function (config) {
       progress(0);
 
       if (data.deviceInfo.productId === FSLIBRE_PRO_PRODUCT_ID) {
-        data.deviceInfo.deviceId = `${data.deviceInfo.driverId}Pro-${data.deviceInfo.serialNumber}`;
+        let generatedSensorId = '';
+        const TABLE_ID_OFFSET = 0;
+        // find oldest historical data packet
```
pazaan: Can we add a comment here about why we use the oldest historical data packet to generate the sensor ID?
Something like "The oldest historical data packet contains a timestamp and the local timezone. Although the hashed value is not globally unique, it is unique for this device, and combined with the device ID is globally unique".

DorianScholz: Thanks for pointing this out. I added an explanation.

PR #560, opened 2017-12-15
Animas fix for "commandpacket.parser is not a function"

gniezen:
See linked Trello card below for details.


gniezen:
[[uploader] Uncaught TypeError: commandpacket.parser is not a function](https://trello.com/c/re4NHbSw/306-uploader-uncaught-typeerror-commandpacketparser-is-not-a-function)

PR #561, opened 2017-12-20
Tandem: Add percentage for temp basal crossing timed segment

gniezen:
One-liner to add percentage for temp basals that cross a timed segment.


gniezen:
[Tandem temp basal display bug](https://trello.com/c/zGZ65BOr/307-tandem-temp-basal-display-bug)

PR #564, opened 2017-12-22
MedT Direct: Fix event markers

gniezen:
For Medtronic Direct, 0x42 should be the insulin event marker, and 0x43 should be the "other" event marker.


gniezen:
[[uploader] Unknown Medtronic Direct record type](https://trello.com/c/heVgxuHy/312-uploader-unknown-medtronic-direct-record-type)

PR #566, opened 2018-01-03
Fix calibration device events for Tandem G4

gniezen:
See https://trello.com/c/VIhtCzG5 for details.


gniezen:
[Tandem calibration values show CGM value instead of calibration value](https://trello.com/c/VIhtCzG5/309-tandem-calibration-values-show-cgm-value-instead-of-calibration-value)

PR #568, opened 2018-01-11
Bayer Contour Next One support

gniezen:
See linked Trello card for details


gniezen:
[Add support for Bayer Contour Next One](https://trello.com/c/XGNCkEmk/317-add-support-for-bayer-contour-next-one)

lib/drivers/bayer/bayerContourNext.js
```diff
@@ -149,7 +150,11 @@ module.exports = function (config) {
   var parseHeader = function (header, callback){
     var pString = header.split('|');
     var pInfo = pString[4].split('^');
-    var sNum = pInfo[2].match(/^\d+\-\s*(\w+)/);
+
+    // serial number can be in two formats:
+    // 6301-1C2CF8C -> C2CF8C
```
pazaan: This would be 1C2CF8C with the original regex.

lib/drivers/bayer/bayerContourNext.js
```diff
@@ -149,7 +150,11 @@ module.exports = function (config) {
...
+    // 7830H5001733 -> 5001733
+    var sNum = pInfo[2].match(/^\d+\-?\w?\s*(\w+)/);
```
pazaan: This won't work for the old format (see comment above about original regex). Suggest `/^\d+[\w-]\s*(\w+)/`.
I'm also not quite sure why the `\s*` is in there, because I don't see any white space. But I'll assume that it must have occurred in an instance in the past for it to be included in the regex.

lib/drivers/bayer/bayerContourNext.js
```diff
@@ -397,12 +420,6 @@ module.exports = function (config) {
 
             var packetHead = struct.unpack(packet, 0, '3Z2b', ['HEADER', 'SIZE', 'BYTE1']);
 
-            if(packetHead['HEADER'] !== MAGIC_HEADER){
```
pazaan: Does the Link One have a different magic header? If so, should we set a different MAGIC_HEADER for that model?
Alternatively, if we're not checking the contents of the header at all any more, that we get rid of MAGIC_HEADER in favour of HEADER_SIZE = 3, and removing all references to MAGIC_HEADER.

lib/drivers/bayer/bayerContourNext.js
gniezen: Hmm, I decided to upload all 11 Bayer meters I have here and my "bug fix" won't work, as that one is the odd one out. I'll explain below in response to the regex comment.

lib/drivers/bayer/bayerContourNext.js
gniezen: I have not seen the white space either, but I also made the same assumption. So here is what happened when I uploaded the 11 meters (all the different models we support, with some mmol/L versions as well) I have here, using the regex you suggested: 10 meters have a correct serial number, except for the BCN Link with serial number C2CF8C, where 6301-1C2CF8C -> 1C2CF8C. 

According to the spec (which is publicly available, so I'll past the relevant sections here):

> Note that the serial number is actually the model number followed by the serial number field. Additionally, the model number can have a 5th character.

> Meter serial number (7350-SSSSSSS). The first 5 characters represent the meter model
number (the dash " – " can be replaced with a capital letter for derived product codes

So according to the spec it seems we should be storing that extra `1` as part of the serial number, even if that is not what is written on the physical device.

lib/drivers/bayer/bayerContourNext.js
gniezen: Funnily enough there never was a MAGIC_HEADER in the spec, as it was supposed to contain the report ID, host ID and device ID. All meters prior to the BCN One reported these as ABC, so it seems it has finally been changed to match the spec. These values are not of use to us, so I will follow your alternative suggestion.

gniezen:
Going to merge this into master and then release together with Verio & Verio Flex PR.

PR #572, opened 2018-02-16
appveyor: skip branch with PR

gniezen:
This is a one-line non user facing change to have appveyor not build for both branch and PR.


gniezen:
I think the reason why it still built both branch and pr is because I swapped the "Required" check from pr to branch previously.

gniezen:
Merged this as part of BCN One PR. Worked well.

PR #573, opened 2018-02-16
Electron update

gniezen:
This PR updates Electron to v1.8.2 and builds with Node v8.2.1.

@krystophv Anything else that we need to change? Do you know what we should be testing here?


krystophv:
This seems to hit all the things I think are necessary for the version update. Testing might come down to "smoke test a bit of everything", as nothing _should_ change, but everything _could_ have. I'm wondering if it'd be worth it to take the opportunity to do some other dependency updates at the same time since testing is likely to be similar? Maybe a couple quality of life improvements like limiting Rollbar to production builds?

gniezen:
@krystophv Yep, totally agree. I'm thinking we could tackle both https://trello.com/c/LLQU6QaX/29-new-card-uploader-dependency-updates and https://trello.com/c/144Btqt7/344-uploader-improve-rollbar-logging as part of this.

gniezen:
@krystophv OK, this one is ready for a second round of code review.

gniezen:
The second and third round of updates that @krystophv made LGTM :ship:

gniezen:
[NEW CARD: Uploader Dependency Updates](https://trello.com/c/LLQU6QaX/326-new-card-uploader-dependency-updates)

gniezen:
Note: @BecCooney found an issue where Verio and Verio Flex uploads would fail on El Capitan. Turns out the `description` field was missing, where we check if it contains `LifeScan Media`. This is a bug in the `drivelist` module that was fixed in the most recent version only 5 days ago (https://github.com/resin-io-modules/drivelist/blame/143ebbe84a6c1e534d689836d1cfb1d7cae328ff/lib/diskutil.js#L110) so I bumped the version number to the latest version.

PR #576, opened 2018-03-07
[WIP] Add right-click menu to allow selection of upload period, fix basal suspend bug

pazaan:
Fix bug that Eric found: Simulator didn't correctly handle the first basal event being a suspend event.


pazaan:
@gniezen - I know I'm missing a Unit Test for this case. I'm still working on it - I just wanted to get this build to Eric to unblock him. Maybe hold off on reviewing until the unit test is here.

lib/drivers/medtronic600/NGPUtil.js
```diff
@@ -52,6 +52,16 @@ class NGPTimestamp {
 }
 
 class NGPConstants {
+  static get PUMP_STATE_FLAGS() {
+    return {
+      SUSPENDED: 1,
+      BOLUSING: 2,
+      ACTIVE: 16, // INACTIVE means pump rewound, needs reservoir
```
gniezen: It looks like each of these flags is one bit that's either on or off. So if `ACTIVE` is 0 it's `INACTIVE`? Any idea what is stored in the two bits between `BOLUSING` and `ACTIVE`?

lib/drivers/medtronic600/medtronic600Simulator.js
```diff
@@ -482,16 +482,29 @@ class Medtronic600Simulator {
...
-            throw new Error('Should never get a duration of less than zero');
-          } else if (suppressedBasal.duration === 0) {
+          if (Date.parse(activeSuspendedBasal.time) >= Date.parse(suppressedBasal.time)) {
+            suppressedBasal.duration = Date.parse(activeSuspendedBasal.time) -
+              Date.parse(suppressedBasal.time);
+          } else {
+            // The suspend event is the first basal event in the history.
+            // Because we don't know what it's suspending, match it to the beginning of the
+            // first suspended event.
+            activeSuspendedBasal.time = suppressedBasal.time;
```
gniezen: it's usually dangerous to set `.time` without using `fillInUTCInfo`. Do we have to set `.time` here manually?

lib/drivers/medtronic600/medtronic600Simulator.js
```diff
@@ -482,16 +482,29 @@ class Medtronic600Simulator {
...
+              Date.parse(suppressedBasal.time).valueOf();
             this.removeEvent(suppressedBasal);
           }
         }
 
+        // It's possible for suppressed basal duration to get down to 0 here (which is fine).
+        if (suppressedBasal.duration < 0) {
+          throw new Error('Should never get a duration of less than zero');
+        } else if (suppressedBasal.duration === 0) {
+          this.removeEvent(suppressedBasal);
```
gniezen: Why do we need to remove this event? It's not uncommon to have basals with zero duration.

lib/drivers/medtronic600/NGPUtil.js
pazaan: Yes, it's a bitfield, so bit 4 being 0 would be INACTIVE. That's a term I came up with myself, because I had to reverse engineer these status bits. I haven't figure out what the missing bits are yet...

lib/drivers/medtronic600/medtronic600Simulator.js
pazaan: In this function (`applySuspendedBasals`), we're synthesizing (in an overlay fashion) suspended basals "over" the existing schedule (scheduled, temp and auto basals). So they're not basals that occured with 0 duration in the event history.

lib/drivers/medtronic600/medtronic600Simulator.js
pazaan: This is a unique case of us modifying the suspend event from its original. In this instance, we pretend that the event occurred after it really did (to the beginning of the first known basal event, which is already "UTC happy"). We do this because we have no idea of the basal event that the suspend is "suspending". 

Thinking through it further, though - this may be causing an issue for the deduplicator though, since it will look like a different event. I'm wondering if we should discard the entire suspended event?

darinkrauss: Which fields are changing here to cause this to look like a different event? Only some fields are used in the deduplication process and those fields depend somewhat on the type of event.

gniezen: I would suggest one of two options:

1. Keep it's original time, but set the duration to 0 and annotate with `basal/unknown-duration`
1. Drop the event, since we don't really have enough information to create a record (as in, both `duration` and `suppressed` is missing) and there's a small chance option 1  could overwrite the event from a previous upload, depending on how we perform deduplication (either now or in the future)

pazaan: @darinkrauss - the start date/time is changing.
@gniezen - in either case, we'd also need to ignore the underlying supressed basals as well. In case 1, we'd need to ignore the basals that _would_ have been suppressed by the actual duration, and in case 2, we'd need to drop the events that happen during the suspend period.

gniezen: > We do this because we have no idea of the basal event that the suspend is "suspending".

In that case, should we not keep the time of the suspended basal event as is, and just not fill in the suppressed field? (As from the Uploader's PoV, we don't have it)

pazaan: I think I have to do some testing to see how the deduplicator handles things...

pazaan:
[670G: Add raw microbolus data to `payload` field of `basal` data type](https://trello.com/c/WZbBxkXA/340-670g-add-raw-microbolus-data-to-payload-field-of-basal-data-type)

pazaan:
[<S> NEW CARD: 6-series Delta Upload](https://trello.com/c/K4dHbrSj/324-s-new-card-6-series-delta-upload)

PR #578, opened 2018-03-13
Skip zero-fill prime events

pazaan:
If the user skips a prime in the pump UI, the pump history will still
contain a fill event, but with an amount of zero. We skip these
zero-fill events.
- Fixes https://trello.com/c/GiWmPRw2

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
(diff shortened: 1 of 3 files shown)


pazaan:
[Cannula prime record created when attempting to skip that step during a set change.](https://trello.com/c/GiWmPRw2/336-cannula-prime-record-created-when-attempting-to-skip-that-step-during-a-set-change)

PR #581, opened 2018-03-13
Rollbar update

gniezen:
See attached Trello card.


gniezen:
[[uploader] Improve Rollbar logging](https://trello.com/c/144Btqt7/327-uploader-improve-rollbar-logging)

PR #582, opened 2018-03-14
Disambiguate 670G SMBGs

pazaan:
The 670G uses a history event called ClosedLoopBloodGlucoseReadingEvent
for SMBGs. This event is logged multiple times, with
a different context for each event.
This branch/commit consolidates all of the separate events into
a single SMBG event with multiple annotations and payload
entries.
- Fixes https://trello.com/c/jnkv0yVy

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -1,5 +1,4 @@
 /*
-
  * == BSD2 LICENSE ==
  * Copyright (c) 2017, Tidepool Project
  *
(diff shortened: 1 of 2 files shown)


pazaan:
[Multiple BG readings showing in the Basics view for each actual BG reading](https://trello.com/c/jnkv0yVy/342-multiple-bg-readings-showing-in-the-basics-view-for-each-actual-bg-reading)

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -942,6 +941,23 @@ class NGPHistoryParser {
     }
   }
 
+  findMatchingClosedLoopBGReadings(bgReadingEvent) {
+    // Find all of the ClosedLoopBloodGlucoseReadingEvents with the same BG value that
+    // occurred within 15 minutes of the BG_READING_RECEIVED.
```
gniezen: Yeah, 15 minutes seem reasonable. That's the same amount of time we use to deduplicate manual and linked values on Animas and Medtronic.

pazaan: Yeah. 15 minutes is also in the CLP code.

pazaan:
@gniezen At the moment, the SMBGs will show with a `?`. I guess it’s six if it shows annotation question marks, and half a dozen if the SMBG count in the daily view is wrong until this is merged.

gniezen:
@pazaan Fair enough. I just don't want us to forget about adding the annotations to tideline so that they render correctly. It would make sense to do it on the same card, or do you want to create a separate card for it?

pazaan:
Created https://trello.com/c/X5TCJYAZ for handling the annotations.

gniezen:
LGTM. :ship:

PR #583, opened 2018-03-14
Allow 'Auto Mode' as an active basal schedule

pazaan:
Auto Mode is a separate mode in the 600-series pumps that support it.
In order to show this in the Settings Page, if Auto Mode is active,
we need to synthesize an empty basal pattern called 'Auto Mode',
and set it to be the active basal pattern.
- Fixes missing feature of https://trello.com/c/JkZGdYaS


pazaan:
[6-Series device settings](https://trello.com/c/JkZGdYaS/338-6-series-device-settings)

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1067,6 +1087,16 @@ class NGPHistoryParser {
   }
 
   buildSettingsRecords(events) {
+    // Determine if the pump is in 'Auto Mode' by looking at the most recent history event
+    // and checking the Closed Loop state.
+    // If the pump is not in Closed Loop, we keep the active basal pattern from the status message.
+    const lastEvent = this.events[this.events.length - 1];
+    if (lastEvent.isClosedLoopActive) {
```
gniezen: Can we be confident that the last event in the pump history will always have the "closed loop active" flag? What if it's an alarm?

lib/drivers/medtronic600/NGPUtil.js
```diff
@@ -123,6 +123,23 @@ class NGPConstants {
     ];
   }
 
+  static get CL_TRANSITION_REASON() {
+    return {
+      INTO_ACTIVE_DUETO_GLUCOSE_SENSOR_CALIBRATION: 0,
+      OUTOF_ACTIVE_DUETO_USER_OVERRIDE: 1,
+      OUTOF_ACTIVE_DUETO_ALARM: 2,
+      OUTOF_ACTIVE_DUETO_TIMEOUT_FRM_SAFE_BASAL: 3,
+      OUTOF_ACTIVE_DUETO_PROLONGED_HIGH_SG: 4,
```
gniezen: Can we make these a bit easier to read by adding `_` between `INTO`, `OUTOF` and `DUETO`? For a moment there I thought "dueto" was Spanish :wink: 

Do we store these values anywhere? I'm guessing they're not supported by the data model yet, which is why Darin is currently working on updating the data model?

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Yes we can. _Every_ event contains this flag. They overloaded the source ID field, so it’s even backward compatible.

lib/drivers/medtronic600/NGPUtil.js
pazaan: Yeah, good point. They’re a copy/paste from CLP, but we can do better. TBH, we’re not even using these at the moment. I added them with the intention of using the ClosedLoopTransitionEvent, but then used the isClosedLoop flag instead, since it was more reliable.

gniezen:
LGTM. :shipit:

PR #587, opened 2018-03-22
Add support for the new automated deliveryType

pazaan:
- Add a function in the `objectBuilder` to support the Tidepool data
  model's new 'automated' deliveryType.
- Change the 600-series Auto Mode basals to use the new
  makeAutomatedBasal function.
deliveryType.
- Adds feature requested in https://trello.com/c/SmmJ2c2W

### DIFF
--- a/lib/drivers/medtronic600/cli/blob_loader.js
+++ b/lib/drivers/medtronic600/cli/blob_loader.js
@@ -23,7 +23,7 @@ function processMedtronic(driverMgr) {
       console.log(err.stack);
       process.exit();
     }
-    console.log(`${intro} All good! Uploaded [%s] events - check in blip :)`, result.postRecords.length);
+    console.log(`${intro} All good! Uploaded [%s] events - check in Tidepool for Web :)`, result.postRecords.length);
     process.exit();
   });
 }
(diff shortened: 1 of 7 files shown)


pazaan:
[Update Uploader to support Auto mode for 670G](https://trello.com/c/SmmJ2c2W/285-update-uploader-to-support-auto-mode-for-670g)

lib/drivers/medtronic600/cli/blob_loader.js
```diff
@@ -49,7 +49,7 @@ function loadFile(filePath, tz, userid) {
         filename: filePath,
         fileData: json,
         timezone: tz,
-        version: `${pkg.name} ${pkg.version}`,
```
gniezen: Is this what caused the issue?

lib/objectBuilder.js
```diff
@@ -131,6 +131,21 @@ module.exports = function () {
     };
   }
 
+  function makeAutomatedBasal() {
```
gniezen: We should add this to the docs.

lib/drivers/medtronic600/cli/blob_loader.js
pazaan: Yes, but the `platform` was returning back an error message that was invalid JSON. @darinkrauss has added it to his TODO list.

lib/objectBuilder.js
pazaan: Good point. Will do 👍

PR #589, opened 2018-03-22
Bug fix: Libre fails to decompress data

gniezen:
I added the capability to save binary blobs for Libre and fixed the CLI tools not uploading due to Rollbar changes. @DorianScholz fixed the bug for an edge case that causes Libre driver to fail while decompressing data.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -14,10 +14,7 @@
     "development": {
       "plugins": [
         "transform-class-properties",
-        "transform-es2015-classes",
-        ["transform-define", {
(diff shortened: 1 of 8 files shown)


gniezen:
@krystophv Please review Rollbar-related changes. 
@pazaan Please review driver-related changes.

gniezen:
[Libre uploading with gaps in data.](https://trello.com/c/5KPT5Iim/346-libre-uploading-with-gaps-in-data)

PR #590, opened 2018-03-27
Fix bugs in extended boluses

pazaan:
The code was previously returning extended bolus durations in minutes,
rather than seconds. Because we were using end events, rather than
begin events, this was causing errors in the recalculation of the event
start time.
Change the durations to be reported in seconds.
- Fixes https://trello.com/c/hZSvQjMZ

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -1521,13 +1521,13 @@ class NGPHistoryParser {
         event.timestamp.offset,
       );
       const bolus = this.cfg.builder.makeSquareBolus()
-        .with_duration(event.deliveredDuration * 1000)
+        .with_duration(event.deliveredDuration * sundial.SEC_TO_MSEC)
         .with_extended(event.deliveredAmount);
 
       if (event.programmedAmount !== event.deliveredAmount ||
         event.programmedDuration !== event.deliveredDuration) {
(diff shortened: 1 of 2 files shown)


pazaan:
[Daily view carb circle misaligned with related bolus bar.](https://trello.com/c/hZSvQjMZ/348-daily-view-carb-circle-misaligned-with-related-bolus-bar)

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1521,13 +1521,13 @@ class NGPHistoryParser {
...
       const bolus = this.cfg.builder.makeSquareBolus()
-        .with_duration(event.deliveredDuration * sundial.MIN_TO_MSEC)
+        .with_duration(event.deliveredDuration * 1000)
         .with_extended(event.deliveredAmount);
 
       if (event.programmedAmount !== event.deliveredAmount ||
         event.programmedDuration !== event.deliveredDuration) {
         bolus.with_expectedExtended(event.programmedAmount)
-          .with_expectedDuration(event.programmedDuration * sundial.MIN_TO_MSEC);
+          .with_expectedDuration(event.programmedDuration * 1000);
```
gniezen: You could use `sundial.SEC_TO_MSEC` instead? I was going to suggest adding a `sundial.MIN_TO_SEC` for the "times 60" use case, but I don't have strong feelings about it.

PR #591, opened 2018-03-29
fslibre fix missing records

DorianScholz:
this fixes two problems that were leading to old data not being uploaded:
- fix the record number wrap, which is not according to the specs (wrap record number not always multiple of 0x10000)
- handle user time change records from the historical DB
-> now we don't have to cap historical data based on the record DB entries


gniezen:
LGTM! :+1: I'll get a test release built and sent for QA.

PR #597, opened 2018-04-17
Bug fix for handling large number of BG targets and insulin sensitivities

gniezen:


gniezen:
[Upload fails when a large number of insulin sensitivities or carb ratios are specified](https://trello.com/c/MyWefnN2/365-upload-fails-when-a-large-number-of-insulin-sensitivities-or-carb-ratios-are-specified)

gniezen:
Merging this one so that I can release this together with #591

PR #598, opened 2018-04-19
Support for TRUE METRIX, TRUE METRIX AIR and TRUE METRIX GO

gniezen:


gniezen:
[[uploader] Trividia True Metrix meters](https://trello.com/c/OFkyZUe6/335-uploader-trividia-true-metrix-meters)

gniezen:
@pazaan One issue that I haven't been able to figure out is that until it gets to the `fetchData` step, the UI hangs after clicking Upload. I think this is related to using async/await, and I'm guessing you've already resolved this for the 600-series driver. Please have a look and tell me what I'm doing wrong :)

gniezen:
Great, just discovered a Windows-specific upload issue, where upload only works on the second try. @pazaan If you want, you can hold off on the review until it's fixed, or review the specific commit later.

gniezen:
OK, looks like I fixed the Windows-specific issue. After connecting, the driver has to power off the meter and then wake it up before it will work on Windows. Makes total sense, right? :stuck_out_tongue:

pazaan:
@gniezen Is that what was causing the UI hang before getting to `fetchData`?

gniezen:
@pazaan Nope, I'm still looking for your advice regarding that one.

gniezen:
@pazaan OK, I've tried to promisify the `receive` function, but for some reason it's missing HID packets that `readTimeout` doesn't. So I promisified `readTimeout` instead. It's not perfect, as it still blocks when it's doing its thing, but at least you can now see the progress bar while it's busy.

lib/hidDevice.js
```diff
@@ -81,15 +102,32 @@ module.exports = function(config) {
...
 
+  function sendPromisified(bytes) {
+    return new Promise((resolve, reject) => {
+      send(bytes, function(err) {
+        if (err) {
+          // exceptions inside promise aren't automatically thrown,
+          // so we have to reject them here
+          reject(err);
+        }
+        resolve();
```
pazaan: Should this be in an `else` clause? Is it otherwise simultaneously calling `reject` and `resolve`?

pazaan:
@gniezen, yeah, I reckon we should back out the promisified send...

gniezen:
OK, done.

PR #600, opened 2018-04-20
timezone copy updates and focus TZ selector on cancel

krystophv:
Some copy updates and simple state additions to handle keeping track of/setting the focus state of the timezone selector when an upload is cancelled from the time check dialog.

Trello card: https://trello.com/c/ERwBfwRa


krystophv:
[Update Device Time Check Copy](https://trello.com/c/ERwBfwRa/369-update-device-time-check-copy)

app/components/TimezoneDropdown.js
```diff
@@ -25,6 +25,16 @@ var cx = require('classnames');
 var styles = require('../../styles/components/TimezoneDropdown.module.less');
 
 class TimezoneDropdown extends React.Component {
+  constructor(props) {
+    super(props);
```
cbwebdevelopment: Don't actually need to pass `props` to `super` here, but not hurting anything either.

app/components/TimezoneDropdown.js
```diff
@@ -25,6 +25,16 @@ var cx = require('classnames');
...
+    this.timezoneSelect = null;
+
+    this.setTimezoneSelect = element => {
+      this.timezoneSelect = element;
+    };
+  }
+
   static propTypes = {
     onTimezoneChange: PropTypes.func.isRequired,
     selectorLabel: PropTypes.string.isRequired,
```
cbwebdevelopment: Should add `onBlur` propType declaration

cbwebdevelopment: also `isTimezoneFocused`

app/components/TimezoneDropdown.js
cbwebdevelopment: Actually, NVM. I see that they recommend it in the React docs.
`Class components should always call the base constructor with props`

Currently, it works without it, but should leave it as per their recommendations.

PR #602, opened 2018-04-24
Device list fixes

gniezen:
See attached Trello cards.

PR #603, opened 2018-04-25
Properly decode Medtronic bolus types; handle dual square wave bolus with no normal portion

gniezen:
See attached Trello card.


gniezen:
[2 consecutive Square calculator Boluses are merged into a single bolus when uploaded from Medtronic 530G](https://trello.com/c/izgIetxG/378-2-consecutive-square-calculator-boluses-are-merged-into-a-single-bolus-when-uploaded-from-medtronic-530g)

lib/drivers/medtronic/medtronicDriver.js
```diff
@@ -972,14 +972,17 @@ module.exports = function (config) {
               callback(null, 'eot');
             });
         },
+        /*
```
pazaan: Is this commented out code expected to be a part of this review?

lib/drivers/medtronic/processData.js
```diff
@@ -356,29 +373,38 @@ var buildBolusRecords = function (records) {
...
         }
 
       } else {
-        // normal bolus
-        bolus = cfg.builder.makeNormalBolus()
-          .with_normal(amount);
 
-        if(programmed !== amount) {
-          bolus = bolus.with_expectedNormal(programmed);
+        if (subType != BOLUS_TYPES.SQUARE) {
```
pazaan: `!==`?

lib/drivers/medtronic/processData.js
```diff
@@ -356,29 +373,38 @@ var buildBolusRecords = function (records) {
...
+            var actualDuration = Math.round((amount / (programmed * 1.0)) * duration);
+            bolus = bolus.with_expectedExtended(programmed)
+                      .with_expectedDuration(duration)
+                      .with_duration(actualDuration);
+          }
+      }
+
+    } else {
+      // normal bolus
+      if (subType != BOLUS_TYPES.NORMAL && subType != BOLUS_TYPES.NORMAL_VARIABLE) {
```
pazaan: `!==`?

lib/drivers/medtronic/medtronicDriver.js
gniezen: Yep. I'm not sure if it will resolve the issue in the card specified in the comment, as the issue is very intermittent. So my reasoning is to comment it out (sending the 'X' is not necessary, as it's used as a wakeup call, which is now performed by the EOT sent just before it) and if it works I can later delete it as part of the card in the comment.

gniezen:
Merging this into master first as v2.5.10, will then be merged into #602 and deployed to prod as v2.5.11.

PR #604, opened 2018-04-27
Fixes for 600-series calibrations

pazaan:
The code was previously using flags in SMBG readings to determine if the
event is a calibration. However, these flags only indicated an _intent_
to calibrate, and not whether the calibration was successful or not,
leading to discrepancies with CareLink reports.
Now using the CALIBRATION_COMPLETE event to register `deviceEvent`s of
subType `calibration`.

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -423,10 +423,7 @@ class ClosedLoopBloodGlucoseReadingEvent extends NGPHistoryEvent {
   }
 
   get isCalibration() {
-    return (
-      this.bgReadingContext === NGPUtil.NGPConstants.BG_CONTEXT.BG_SENT_FOR_CALIB ||
-      this.bgReadingContext === NGPUtil.NGPConstants.BG_CONTEXT.ENTERED_IN_SENSOR_CALIB
-    );
+    return this.bgReadingContext === NGPUtil.NGPConstants.BG_CONTEXT.BG_SENT_FOR_CALIB;
   }
 }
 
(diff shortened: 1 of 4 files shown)


lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -2049,7 +2050,7 @@ module.exports = (config) => {
 
       const sessionInfo = {
         delta: cfg.delta,
-        deviceTags: ['insulin-pump'],
+        deviceTags: ['insulin-pump', 'cgm'],
```
gniezen: For time-check to work correctly, you need to define the `deviceTags` at the beginning just after you've cloned `cfg`, as `cfg.deviceTags`, and then pass them to `sessionInfo` here as `deviceTags: cfg.deviceTags,`. This is because we need to know the type of the device when we check the time in order to display the correct message.

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -979,6 +988,18 @@ class NGPHistoryParser {
...
 
+  findSmbgForCalibration(calibrationEvent) {
+    const calibrationIndex = _.findIndex(this.events, event => event === calibrationEvent);
+
+    return _.findLast(
+      this.events, item => (item.eventType === NGPHistoryEvent.EVENT_TYPE.BG_READING ||
+        (item.eventType === NGPHistoryEvent.EVENT_TYPE.CLOSED_LOOP_BG_READING &&
+          item.isFirstReading())) &&
+        item.bgValue === calibrationEvent.bgValue
+      , calibrationIndex,
```
gniezen: I know this is super nit-picky,  but isn't that comma in a strange place?

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1287,34 +1308,43 @@ class NGPHistoryParser {
     return this;
   }
 
+  buildCalibrationRecords(events) {
+    for (const event of this.eventsOfType(NGPHistoryEvent.EVENT_TYPE.CALIBRATION_COMPLETE)) {
+      // In the Tidepool data model, calibration event timestamps match against the SMBG timestamp.
```
gniezen: I'm not sure if that is always true? For other devices I've just used the timestamp of the calibration event. The reason why they (almost) always match is because most devices don't generate a separate SMBG event, in which case it needs to be created. I guess in this case you want to know when the measurement happened, not when the calibration was accepted, so it makes sense.

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Yep, it is. Thanks `eslint-config-airbnb 16.1.0`!

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: I was basing it from [this data model documentation](http://developer.tidepool.org/data-model/device-data/types/deviceEvent/calibration.html#subtype).

gniezen: :+1:

PR #606, opened 2018-04-30
UTC bootstrapping fix

gniezen:
See attached Trello card for more details

### DIFF
--- a/lib/TimezoneOffsetUtil.js
+++ b/lib/TimezoneOffsetUtil.js
@@ -71,13 +71,16 @@ module.exports = function(timezone, mostRecent, changes) {
     }
 
     // if timezone offset is invalid, move to conversion offset until valid
-    while (timezoneOffset > 840) {
-      timezoneOffset -= 1440;
-      conversionOffset += 1440 * 60 * 1000;
+    var TWENTY_FOUR_HOURS = 1440;
+    var MAX_TIMEZONE_OFFSET = 840;
+    var MIN_TIMEZONE_OFFSET = -720;
+    while (timezoneOffset > MAX_TIMEZONE_OFFSET) {
+      timezoneOffset -= TWENTY_FOUR_HOURS;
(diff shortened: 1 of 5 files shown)


gniezen:
[Apply UTC Bootstrapping Fix for Invalid Timezone Offset to new data](https://trello.com/c/rh2oLpjR/385-apply-utc-bootstrapping-fix-for-invalid-timezone-offset-to-new-data)

yarn.lock
```diff
@@ -267,10 +267,6 @@ ansi-html@0.0.7:
   version "0.0.7"
   resolved "https://registry.yarnpkg.com/ansi-html/-/ansi-html-0.0.7.tgz#813584021962a9e9e6fd039f940d12f56ca7859e"
 
-ansi-regex@*, ansi-regex@^3.0.0, ansi-regex@~3.0.0:
```
darinkrauss: Did you mean to include these changes to this PR?

lib/TimezoneOffsetUtil.js
```diff
@@ -69,6 +69,16 @@ module.exports = function(timezone, mostRecent, changes) {
     else {
       conversionOffset += difference.rawDifference;
     }
+
+    // if timezone offset is invalid, move to conversion offset until valid
+    while (timezoneOffset > 840) {
```
darinkrauss: Might want to consider using well-named constants rather than the raw numbers, which just look like secret values unless you understand that min/max time zone offset is -12/14 hours and `timezoneOffset` is in minutes. This would apply to the `840`, `-720`, and `1440`. Up to you, though.

yarn.lock
gniezen: Yeah, that's fine. I think it's just some sub-dependencies that got updated.

gniezen:
Merging this one into master first, to be deployed to production as part of v2.6.2

PR #609, opened 2018-05-03
[WIP] Comms error handling bug fix

pazaan:
Sets retries back to zero after issuing a "ResendPacketsCommand"
so that we don't spam the output command queue on the USB stick
with requests.
Doing so causes the stick to get buffer issues, and also means that we
double up on getting packets that we already had requested.

Also throw an InvalidStateError if we get packets we already received,
to make it really obvious that something is still not quite right.

### DIFF
--- a/lib/drivers/medtronic600/medtronic600Driver.js
+++ b/lib/drivers/medtronic600/medtronic600Driver.js
@@ -761,6 +761,9 @@ class MultipacketSession {
 
   addSegment(segment) {
     debug(`*** Got a Multipacket Segment: ${segment.packetNumber + 1} of ${this.packetsToFetch}, count: ${this.segmentsFilled + 1}`);
+    if (this.packetsToFetch === this.segmentsFilled) {
+      throw new InvalidStateError('Segments already filled. Should not be getting duplicates.');
+    }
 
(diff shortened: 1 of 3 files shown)


pazaan:
[When the 670G sensor has blocked communication with the pump during upload, the upload fails.](https://trello.com/c/XLio0cA6/383-when-the-670g-sensor-has-blocked-communication-with-the-pump-during-upload-the-upload-fails)

pazaan:
@gniezen This is a definite improvement, but I'll leave this branch open in case I make further improvements.

lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -988,6 +991,14 @@ class TransmitPacketRequest extends NGPRequest {
           this.ngpRetries += 1;
         }
         debug(`Multipacket retry: ${this.ngpRetries}`);
```
gniezen: I guess this line can now be removed, as it looks like a duplicate of line 995?

pazaan: Removed the duplicate information from line 995.

PR #614, opened 2018-05-04
Correctly separate SMBGs of the same value with a 15 minute window

pazaan:
It's possible for multiple SMBG events to occur within 15 minutes, but
still be legitimately considered separate SMBG events.
Add a fix to correctly separate the SMBGs.

### DIFF
--- a/lib/drivers/medtronic600/medtronic600Driver.js
+++ b/lib/drivers/medtronic600/medtronic600Driver.js
@@ -2028,6 +2028,15 @@ module.exports = (config) => {
 
         events = _.sortBy(events, datum => datum.time);
 
+        // If the delta information hasn't been defined after reading the pump, use the last event
+        // date, if in debug mode. This can happen when using the blob_loader.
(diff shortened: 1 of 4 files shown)


lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -2028,6 +2028,15 @@ module.exports = (config) => {
 
         events = _.sortBy(events, datum => datum.time);
 
+        // If the delta information hasn't been defined after reading the pump, use the last event
+        // date, if in debug mode. This can happen when using the blob_loader.
+        if (process.env.DEBUG_ERROR === 'true' && _.isUndefined(cfg.delta)) {
```
gniezen: Hmm, I'm just trying to think through what unintended consequences this  will have, as debug mode can be enabled by a user in production mode. If it only affects blob_loader usage, why not just check if it's the blob_loader instead? I always use `var isBrowser = typeof window !== 'undefined';` to determine if it's the blob_loader or not.

pazaan: That's a great suggestion. Will do that - thanks!

PR #615, opened 2018-05-04
Calculate suspend duration from pump RTC rather than device time

pazaan:
The Real Time Clock (RTC) of the pump is always advancing, making it a
reliable mechanism to determine duration between events, as opposed to
the pump's display time.

### DIFF
--- a/lib/drivers/medtronic600/medtronic600Simulator.js
+++ b/lib/drivers/medtronic600/medtronic600Simulator.js
@@ -475,9 +475,6 @@ class Medtronic600Simulator {
               break;
             }
           }
-          if (suppressedBasal == null) {
-            throw new Error('Could not find previous suppressed basal');
-          }
         } else {
           suppressedBasal = null;
         }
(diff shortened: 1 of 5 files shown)


pazaan:
[Cannot currently upload Lennart's 640G](https://trello.com/c/JQGTaHzP/361-cannot-currently-upload-lennarts-640g)

lib/drivers/medtronic600/medtronic600Simulator.js
```diff
@@ -475,9 +475,6 @@ class Medtronic600Simulator {
               break;
             }
           }
-          if (suppressedBasal == null) {
```
gniezen: Just checking: So it's ok if the previous suppressed basal is not found?

pazaan: Yep, because if the suppressed basal is missing, it won't generate the suspend, which means that we have to rely on a previous upload, or just don't upload it.
Take this scenario:
A schedule basal of 3 hours happens, then there's a suspend 2 hours in for 30 minutes. That works fine if all the data is there.
However, if we started reading the data before the schedule event (say an hour into the scheduled basal), we'll get the suspend event, but NOT the underlying suppressed information. In that instance, we don't show the scheduled basal, OR the suspend. 😄

PR #616, opened 2018-05-08
Bump version number for QA build

pazaan:


pazaan:
[Daily view carb circle misaligned with related bolus bar.](https://trello.com/c/hZSvQjMZ/348-daily-view-carb-circle-misaligned-with-related-bolus-bar)

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1017,6 +1023,19 @@ class NGPHistoryParser {
     return null;
   }
 
+  // This is here rather than in the simulator because the boluses both share a bolusNumber
+  // as an identifier.
+  findMatchingBolusProgrammedEvent(bolus) {
+    const matchingBolusIndex = _.findIndex(this.events, event => event === bolus);
+
+    return _.findLast(
+      this.events, event => event.eventType === bolus.eventType - 199 &&
```
gniezen: What is the 199 - can it be a constant definition instead?

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1529,14 +1548,20 @@ class NGPHistoryParser {
         continue;
       }
 
+      // We need the timestamp of the time the Bolus event was programmed, not
+      // when it finished. Even for a normal basal, delivery for 100g of carbs can
+      // take a few minutes.
+      const programmedBolus = this.findMatchingBolusProgrammedEvent(event);
+      const beginTimestamp = _.isUndefined(programmedBolus) ? event.timestamp :
```
gniezen: Maybe we can factor all of these out into a `getBeginTimestamp` function? 

How rare would it be to have an bolus without a programmed event, where we only have the event timestamp? Should we maybe add an annotation, payload note or just a debug statement when it can't find the programmed timestamp?

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: I'll make a const (and document it)

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Yes, good idea about a single function.

It would be unusual, but possible. In fact, maybe we should completely discard these events in this case (similar to the suspends without matching suppressed basal info).

PR #619, opened 2018-05-10
 Fix suspend duration bug introduced in PR#615

pazaan:
#615 was made to fix a bug that could happen due to a time change.
However, it introduced a bug where the duration was set in seconds,
rather than in milliseconds.
Fix this introduced bug, as well as making a unit test to protect
against the failure again.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
-  "version": "2.5.9-600series-qa.25",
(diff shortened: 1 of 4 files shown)


pazaan:
[Cannot currently upload Lennart's 640G](https://trello.com/c/JQGTaHzP/361-cannot-currently-upload-lennarts-640g)

pazaan:
[6-series bug: basal suspend gaps missing from Daily view.](https://trello.com/c/T5x6YSN7/402-6-series-bug-basal-suspend-gaps-missing-from-daily-view)

PR #620, opened 2018-05-14
Bump version for QA build

pazaan:


pazaan:
[If Auto Mode starts with no microbolus delivery, data is incorrect](https://trello.com/c/sYxXIylr/403-if-auto-mode-starts-with-no-microbolus-delivery-data-is-incorrect)

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1427,6 +1443,23 @@ class NGPHistoryParser {
...
       events.push(basal);
     }
+
+    // Add gaps from beginning of Auto Mode start until first microbolus, or basal event
+    for (const event of this.eventsOfType(NGPHistoryEvent.EVENT_TYPE.CLOSED_LOOP_TRANSITION)) {
+      if (event.transitionValue ===
+        NGPUtil.NGPConstants.CL_TRANSITION_VALUE.CL_INTO_ACTIVE) {
+        const gap = this.getAutoModeStartGap(event);
+        const basal = this.cfg.builder.makeAutomatedBasal()
+          .with_rate(0.0)
```
gniezen: Does this need an annotation, as it's fabricated from the auto-mode start event? For example, we use `medtronic/basal/fabricated-from-schedule` and `tandem/basal/fabricated-from-new-day` for other basals that are fabricated from events, see https://github.com/tidepool-org/tideline/blob/master/js/plot/util/annotations/annotationdefinitions.js

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1046,6 +1046,22 @@ class NGPHistoryParser {
     return _.isUndefined(programmedBolus) ? undefined : programmedBolus.timestamp;
   }
 
+  getAutoModeStartGap(autoModeStartEvent) {
+    const autoModeEventIndex = _.findIndex(this.events, event => event === autoModeStartEvent);
+
+    // Auto Mode can only last for 2 hours without a microbolus
```
gniezen: Should this search then be limited to two hours?

pazaan: Ah, no. I need to delete that comment. Turns out that there were cases where that wasn't true.

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Sounds like a "yes" to me 😄

PR #621, opened 2018-05-15
Ultra 2 comms fix

gniezen:
A number of Rollbar reports indicate that the device header is sometimes missing when retrieving data from the meter. This PR decreases the listen interval to make sure that we're not missing packets being received. I'm not sure that this will fix it, but it's a start.

PR #622, opened 2018-05-18
Update serialport to v6

gniezen:
This PR updates serialport to v6.2.0, but also contains code from the following PRs that have already been reviewed, so that they can all be tested together: #613 , #621 

Basically, all that needs to be reviewed here is commit 5fe7c93


gniezen:
[Update node-serialport to version 6](https://trello.com/c/ucvNt5d0/406-update-node-serialport-to-version-6)

gniezen:
@pazaan I've made reasonably significant changes to this PR, so should be re-reviewed. Specifically `oneTouchUltra2.js` and `serialDevice.js`.

gniezen:
One thing that was happening with the Ultra 2 driver is that the interval timer would re-trigger itself on slower machines causing partial uploads, so now I clear it as soon as we being receiving data. I've tested it and it now works fine on slower machines on Windows 7, 8.1 and 10.

I also discovered that there is a known upstream bug in serialport that causes race condition on connect. This is something that affects quite a large number of our users (> 70), so I implemented a solution that I got from the upstream bug report: Basically, retry to connect if it doesn't work on the first try :smile:

PR #623, opened 2018-05-21
Tandem CLI script (UPLOAD-122)

rspier:
Based on lib/drivers/abbott/cli/fslibre.js.


rspier:
FYI @pazaan

gniezen:
Hi @rspier, thanks for this! I think this is a first step towards having a stand-alone driver (by decoupling API from driver), which will be useful if/when we want to share the drivers as their own npm packages, so that others can use it in their apps. 

It's also a start towards having binary blobs for the Tandem driver. In future we want to be able to store all binary blobs on the backend, so that we can reproduce issues on our side and/or re-process the raw data if necessary. This work is planned for the current quarter, see https://trello.com/c/DMgzvGOL for more details.

One concern is that to merge this PR we will need to re-test the delta upload functionality. To get an idea of what is required, see the test strategy on this card: https://trello.com/c/ieSwFQq5/5-m-delta-uploads-4-of-4-tandem. The workload of our QA team is currently very high, so I don't see this happening soon.

rspier:
You're welcome.  It scratched an itch.

I took the simple path to decoupling.  I had considered feature flags (i.e. checkDeviceTime, incrementalUpload) but they were going to be either significantly more invasive or more confusing.  Right now the Tandem driver doesn't have a centralized "default config" concept.

Re binary blobs, if you trust the record parsing, I think it would be pretty simple to add the binary (raw) record as a field on the decoded version.

Re testing, thanks for sharing that.  It made me take another look at the code, and I found a bug in my code (now fixed.)  There's no rush on merging this, although it's pretty straightforward to test and can probably be tested as part of another change.  I'm very excited that you have a QA department.  Would it be helpful for me to document the functionality working with log extracts?

I also found a pre-existing bug in the driver, it always re-fetches the last record from the device.

gniezen:
Closing this PR as the Tandem driver is now private :disappointed:

ferringb:
@gniezen can you please link to a commit or ticket detailing exactly what 'private' is going to mean for end users w/ tandem pumps?  Worst case if I'm losing support I'll know exactly what to point at when I yell at tandem... ;)

rspier:
Looks like https://github.com/tidepool-org/uploader/pull/1503/commits/34e707f8752fa7988dbe2f140c964d99f852db4c

gniezen:
This will not impact end users w/ tandem pumps.

Here is the ticket: https://tidepool.atlassian.net/browse/UPLOAD-835

Moving forward, Tandem will be working directly with third parties that distribute software that connects with Tandem devices.

PR #625, opened 2018-05-31
Converts pump serial to pump MAC

gniezen:
What it says on the tin. 

@pazaan I filled in my pump serial as I wasn't sure if you want yours in the public repo. You're welcome to change it.


gniezen:
Ah, never mind what I said about your pump serial number, I see it's already being sent in `setPumpMACFromSerial`

PR #626, opened 2018-06-01
Handle the last basal event being a temp basal start

pazaan:
Previously when the upload happened in the middle of a temp basal
(ie temp basal hadn't completed), the data was uploaded incorrectly
as a regular scheduled basal, meaning that for the next upload, the
basals would be overlayed, rather than replaced.
This commit fixes that bug.
* Fixes https://trello.com/c/fk1vqCBz

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -211,6 +211,8 @@ class NGPHistoryEvent {
         return new ClosedLoopTransitionEvent(this.eventData);
       case NGPHistoryEvent.EVENT_TYPE.BASAL_SEGMENT_START:
         return new BasalSegmentStartEvent(this.eventData);
+      case NGPHistoryEvent.EVENT_TYPE.TEMP_BASAL_PROGRAMMED:
+        return new TempBasalProgrammedEvent(this.eventData);
       case NGPHistoryEvent.EVENT_TYPE.TEMP_BASAL_COMPLETE:
         return new TempBasalCompleteEvent(this.eventData);
       case NGPHistoryEvent.EVENT_TYPE.REWIND:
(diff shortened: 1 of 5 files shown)


lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -490,7 +520,10 @@ class TempBasalCompleteEvent extends NGPHistoryEvent {
   }
 
   get duration() {
-    return this.eventData.readUInt16BE(0x12) * 60;
```
gniezen: WTH? The event size can change? Is there a comment we can put here to explain why this happens? Or is it just :man_shrugging: ?

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1520,6 +1558,43 @@ class NGPHistoryParser {
...
+        NGPHistoryEvent.EVENT_TYPE.TEMP_BASAL_COMPLETE,
+      ).next().value;
+    if (lastTempBasalEvent &&
+      lastTempBasalEvent.eventType === NGPHistoryEvent.EVENT_TYPE.TEMP_BASAL_PROGRAMMED) {
+      const suppressedBasal = this.findSuppressedBasal(lastTempBasalEvent);
+      const basal = this.cfg.builder.makeTempBasal()
+        .with_rate(lastTempBasalEvent.rate)
+        .with_expectedDuration(lastTempBasalEvent.expectedDuration)
+        .with_duration(lastTempBasalEvent.expectedDuration)
+        .set('suppressed', {
```
gniezen: What happens if there is no suppressed basal? Does it just send up `null`?

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: I don't know why this happens. I just noticed this in the CLP code today.

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: Yes, it would send up null. The only time I could see this happening is if the *only* basal event in the whole upload was a TEMP_BASAL_PROGRAMMED.

pazaan:
[[Uploader 600-series] When a temp basal starts before an upload, data is uploaded incorrectly](https://trello.com/c/fk1vqCBz/411-uploader-600-series-when-a-temp-basal-starts-before-an-upload-data-is-uploaded-incorrectly)

PR #633, opened 2018-06-14
Handle missing records for 5/7-series

gniezen:
See attached Trello cards for details.

### DIFF
--- a/lib/drivers/medtronic/processData.js
+++ b/lib/drivers/medtronic/processData.js
@@ -1085,6 +1085,8 @@ function buildTimeChangeRecords(records, mostRecent, type) {
 
   if(type === 'cgm') {
     timeChangeRecords = filterHistory([CBG_RECORD_TYPES.DATE_TIME_CHANGE], records);
+    //TODO: remove the following line
+    console.log('CGM time change records:', JSON.stringify(timeChangeRecords, null, 4));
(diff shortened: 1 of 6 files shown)


gniezen:
[Missing CGM time change records for 5/7-series](https://trello.com/c/ow0JdtX9/417-missing-cgm-time-change-records-for-5-7-series)

gniezen:
[After a 24-hour temp basal, resuming scheduled basal sometimes missing in 5/7-series pump history](https://trello.com/c/KToTSpZx/418-after-a-24-hour-temp-basal-resuming-scheduled-basal-sometimes-missing-in-5-7-series-pump-history)

lib/drivers/medtronic/processData.js
```diff
@@ -1083,6 +1089,11 @@ function buildTimeChangeRecords(records, mostRecent, type) {
 
   var timeChangeRecords = null;
 
+  function getCGMTimeChangeType(obj) {
+    var type = obj.date[1] >> 5 & 7;
```
pazaan: This data is used for other history types as well, and is called `subCode`. Not sure if you want to make it more generic now. Not a big deal, since you're only using it for the TimeChangeType at the moment.

gniezen: Good point. Will change it.

PR #636, opened 2018-06-15
Remove CCITT checksum from ComD payload

pazaan:
Having the CCITT checksum as part of the ComD payload caused some issues
for messages that were expecting to be able to read the whole payload,
and found an unwanted extra 2 bytes there (the checksum).
Also renamed the member variable to be consistent with its purpose.

* Fixes https://trello.com/c/S11Adtjz

This also includes a change to only leave the 802.15.4 network if we
have an ad hoc connection. Doing so all the time was causing linked
meters and pumps to become unlinked, which was bad.

### DIFF
--- a/lib/drivers/medtronic600/medtronic600Driver.js
+++ b/lib/drivers/medtronic600/medtronic600Driver.js
@@ -703,6 +703,8 @@ class TransmitPacketResponse extends NGPResponse {
     super(payload, pumpSession);
 
     // Decrypt response and write it to another member
+    // TODO: These aren't necessarily bad messages. It looks like some of them
+    // could just be pings when the pump is busy processing data.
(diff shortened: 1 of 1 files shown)


pazaan:
[[600-series bug] Cannot read pump or CGM history page if the response is not a multipacket transfer](https://trello.com/c/S11Adtjz/421-600-series-bug-cannot-read-pump-or-cgm-history-page-if-the-response-is-not-a-multipacket-transfer)

pazaan:
[<M> NEW CARD: 6-series ContourNextLink 2.4 ad-hoc pairing](https://trello.com/c/ZbLoMLY0/329-m-new-card-6-series-contournextlink-24-ad-hoc-pairing)

PR #642, opened 2018-06-22
Additional pump settings

gniezen:


gniezen:
[Additional pump settings](https://trello.com/c/7hyUrl24/415-additional-pump-settings)

lib/drivers/insulet/insuletDriver.js
```diff
@@ -461,6 +462,12 @@ module.exports = function (config) {
     eLogRecSizeErr: { value: 8, name: 'eLogRecSizeErr' }
   };
 
+  var TEMP_BASAL_TYPES = {
```
pazaan: Should we add an "enum" somewhere common to standardise `percent` and `Units/hour`?

gniezen: This one is specific to OmniPod though, as other pumps may have them in a different order.

gniezen:
@pazaan Can you please review the changes I made in commit 3719443 after Darin's latest changes to the data model?

lib/drivers/insulet/insuletDriver.js
```diff
@@ -1474,9 +1474,13 @@ module.exports = function (config) {
       .with_bgTarget(bgTargetSettings)
       .with_bolus({
         calculator: {
-          enabled: data.eeprom_settings.BOL_CALCS ? true : false
+          enabled: data.eeprom_settings.BOL_CALCS ? true : false,
+          insulin: {
+            duration: data.eeprom_settings.INSULIN_DURATION * 30,
+            units: 'minutes'
```
pazaan: Shouldn't this be `hours`? The validation on the platform side is `0.0 <= x <= 10.0`

lib/drivers/tandem/tandemDriver.js
```diff
@@ -1348,17 +1348,17 @@ module.exports = function (config) {
...
         amountMaximum : {
           value: profile.max_bolus,
           units: 'Units'
         },
         calculator: {
-          enabled: profile.carb_entry ? true : false
+          enabled: profile.carb_entry ? true : false,
+          insulin: {
+            duration : profile.insulin_duration,
+            units: 'minutes'
```
pazaan: Shouldn't this be `hours`? The validation on the platform side is `0.0 <= x <= 10.0`

lib/drivers/insulet/insuletDriver.js
pazaan: I meant more something like:
```
const TEMP_BASAL_TYPES = {
  OFF: 'off',
  PERCENT: 'percent',
  UNITS: 'Units/hour',
};
```
in a common place, so that in our driver code we can use, say, `TEMP_BASAL_TYPES.UNITS`, instead of `'Units/hour'`, so that we don't end up with typos. I guess that it's not that big of a win in JavaScript, because there's no type-safety and compile-time checks.

gniezen:
I'm going to merge this already, as I want to release it together with #716 as v2.8.2.

PR #653, opened 2018-06-25
Make sure we read all the data on an Ultra 2 meter

gniezen:


lib/drivers/onetouch/oneTouchUltra2.js
```diff
@@ -203,17 +204,42 @@ module.exports = function (config) {
...
-          var ms = l.match(dpattern);
-          index += 1;
-          if(ms){
+         return cb(new Error('Checksum error'));
+    }
+
+    var count = 0;
+
+    async.whilst(
+      function () { return (splites.length - 2) < nrecords + 1; },
```
pazaan: Should this be called `splits` instead of `splites`?

gniezen:
[Ultra 2 not uploading all available data on meter](https://trello.com/c/m1Q6G5UO/430-ultra-2-not-uploading-all-available-data-on-meter)

PR #659, opened 2018-06-29
Upload blobs

gniezen:


gniezen:
[Store upload blobs](https://trello.com/c/DMgzvGOL/435-store-upload-blobs)

app/containers/App.js
```diff
@@ -126,6 +127,7 @@ export class App extends Component {
   setServer = info => {
     console.log('will use', info.label, 'server');
     var serverinfo = serverdata[info.label];
+    serverinfo.environment = info.label;
```
krystophv: This will use `Development`, `Staging`, `Production`, etc. instead of `dev`, `stg`, `prd`. I'm fine with the more verbose names, just noting that the card seems to reference the shortened/lowercase versions of how we refer to the environments in most developer settings.

app/actions/utils.js
```diff
@@ -56,10 +60,35 @@ export function makeDisplayModal(dispatch) {
   };
 }
 
-export function makeUploadCb(dispatch, getState, errCode, utc) {
-  return (err, recs) => {
+export function makeUploadCb(dispatch, getState, errCode, utc, api) {
+  return async (err, recs) => {
```
krystophv: This will be the first `async` function outside of device code. I'm not opposed to using the newer syntax, but it does bring up a slightly larger conversation around consistency.

lib/core/api.js
```diff
@@ -636,6 +657,23 @@ api.getMostRecentUploadRecord = function(userId, deviceId, cb) {
   });
 };
 
+api.upload.blob = function(blob, contentType) {
+  api.log('POST /blobs');
+
+  var digest = btoa(md5(blob, null, true));
+  var blobObject = new Blob([blob], {type: contentType});
+
+  return new Promise((resolve, reject) => {
```
krystophv: Basically every other `api.` exposed function follows the function signature of accepting a callback as the last argument as opposed to returning a `Promise`. I won't argue that the the newer `Promise` symantics aren't better (oh, they're so much better...) but I don't think we should introduce the new syntax one function at a time as it'll lead to a lack of internal consistency across the api surface. (Does this function return a `Promise` vs. do I give it a callback.) I'd love to make a sweeping set of changes to update a lot of this sort of thing, but I'm going to recommend sticking with the callback for now.

app/actions/utils.js
krystophv: And I think this will be less relevant if the `api` signature is changed to accept a callback instead of return a `Promise`

gniezen: I thought we could use any ES6 functions, but I understand you concern regarding consistency and confusion about it being `Promise` or a callback that's being returned. Will change to callbacks.

krystophv: We can use any ES6 functions. We've just not made the conceptual jump to a newer async control semantics in any of our repos yet. While I very much enjoy the async/await syntax I think that moving to it makes more sense as a larger coordinated changeset as opposed to just migrating one function at a time.

gniezen:
Pinging @pazaan for reviewing new changes since last review.

PR #663, opened 2018-07-03
Add `wizard/target-automated` annotation when in Auto Mode

pazaan:


pazaan:
[When in Auto mode, Boluses that include BG data list “Target: Null” in the tooltip](https://trello.com/c/krrF6CmW/434-when-in-auto-mode-boluses-that-include-bg-data-list-target-null-in-the-tooltip)

PR #675, opened 2018-07-13
Catch HID send/receive errors and return them

gniezen:


gniezen:
[Uncaught Error: Cannot write to HID device](https://trello.com/c/i95g40KF/442-uncaught-error-cannot-write-to-hid-device)

PR #678, opened 2018-07-23
Remove fabricated-from-automode-start annotation

pazaan:
See https://trello.com/c/sYxXIylr/403-if-auto-mode-starts-with-no-microbolus-delivery-data-is-incorrect#comment-5b4527582ff6ce8edfda6847


pazaan:
[If Auto Mode starts with no microbolus delivery, data is incorrect](https://trello.com/c/sYxXIylr/403-if-auto-mode-starts-with-no-microbolus-delivery-data-is-incorrect)

PR #680, opened 2018-07-26
Throw an error when attempting to use ad hoc pairing with a linked meter

pazaan:
Throw an error when the user attempts to use ad hoc pairing to a pump
with a different serial number than the meter is already linked with

* Fixes https://trello.com/c/6ESCkbTM

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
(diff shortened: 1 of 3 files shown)


pazaan:
[[600-series bug] Throw an error when attempting to use ad hoc pairing with a linked meter](https://trello.com/c/6ESCkbTM/428-600-series-bug-throw-an-error-when-attempting-to-use-ad-hoc-pairing-with-a-linked-meter)

PR #683, opened 2018-08-03
Bug fixes for Soft Launch issues

pazaan:
If the first basal event from the pump history we read is a suspend
event that doesn't overlap a scheduled basal, throw it away
since we don't know what the supressed basal information is

* Fixes https://trello.com/c/FQ2o9TgH

### DIFF
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -1562,29 +1562,34 @@ class NGPHistoryParser {
     if (lastTempBasalEvent &&
       lastTempBasalEvent.eventType === NGPHistoryEvent.EVENT_TYPE.TEMP_BASAL_PROGRAMMED) {
       const suppressedBasal = this.findSuppressedBasal(lastTempBasalEvent);
-      const basal = this.cfg.builder.makeTempBasal()
-        .with_rate(lastTempBasalEvent.rate)
-        .with_expectedDuration(lastTempBasalEvent.expectedDuration)
-        .with_duration(lastTempBasalEvent.expectedDuration)
-        .set('suppressed', {
-          type: 'basal',
-          deliveryType: 'scheduled',
-          rate: suppressedBasal.rate,
-          scheduleName: suppressedBasal.patternName,
-        });
+      // Don't create the temp basal if we don't find the underlying suppressed info
+      // This could happen if the ONLY basal event we retrieve from the pump history
+      // is a TEMP_BASAL_PROGRAMMED.
+      if (!_.isUndefined(suppressedBasal)) {
+        const basal = this.cfg.builder.makeTempBasal()
+          .with_rate(lastTempBasalEvent.rate)
+          .with_expectedDuration(lastTempBasalEvent.expectedDuration)
+          .with_duration(lastTempBasalEvent.expectedDuration)
+          .set('suppressed', {
+            type: 'basal',
+            deliveryType: 'scheduled',
+            rate: suppressedBasal.rate,
+            scheduleName: suppressedBasal.patternName,
+          });
 
-      // We need to find the normal rate here to calculate event.rate, because
-      // if it's a PERCENT record, event.rate will be 0.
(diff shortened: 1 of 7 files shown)


pazaan:
[[600-series bug] Throw away any isolated initial suspends](https://trello.com/c/FQ2o9TgH/453-600-series-bug-throw-away-any-isolated-initial-suspends)

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1566,29 +1562,34 @@ class NGPHistoryParser {
...
-        .set('suppressed', {
-          type: 'basal',
-          deliveryType: 'scheduled',
-          rate: suppressedBasal.rate,
-          scheduleName: suppressedBasal.patternName,
-        });
+      // Don't create the temp basal if we don't find the underlying suppressed info
+      // This could happen if the ONLY basal event we retrieve from the pump history
+      // is a TEMP_BASAL_PROGRAMMED.
+      if (!_.isUndefined(suppressedBasal)) {
```
gniezen: How about still displaying the temp basal, but not adding the suppressed info if not available? Suppressed info is an optional field.

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1566,29 +1562,34 @@ class NGPHistoryParser {
...
-        basal
-          .with_percent(lastTempBasalEvent.percentageOfRate / 100.0)
-          .with_rate(suppressedBasal.rate * (lastTempBasalEvent.percentageOfRate / 100.0));
-      }
+        // We need to find the normal rate here to calculate event.rate, because
+        // if it's a PERCENT record, event.rate will be 0.
+        if (lastTempBasalEvent.type === NGPUtil.NGPConstants.TEMP_BASAL_TYPE.PERCENT) {
+          basal
+            .with_percent(lastTempBasalEvent.percentageOfRate / 100.0)
+            .with_rate(suppressedBasal.rate * (lastTempBasalEvent.percentageOfRate / 100.0));
```
gniezen: If you do decide to follow previous suggestion, you'll have to use the current pump settings to determine the rate in the case where suppressed info is not available.

lib/drivers/medtronic600/medtronic600Simulator.js
```diff
@@ -506,13 +506,21 @@ class Medtronic600Simulator {
...
           activeSuspend.deviceTime = basal.deviceTime;
           activeSuspend.duration = Date.parse(activeSuspend.basalEndTime) - Date.parse(basal.time);
           Medtronic600Simulator.setSuppressedBasal(activeSuspend, basal);
-          Medtronic600Simulator.checkDuration(activeSuspend);
+
+          if (activeSuspend.duration <= 0) {
+            // If the suspend duration is not a positive number, then the suspend
+            // is not overlapping a scheduled basal.
+            // Delete the suspend, since we don't know what it's suppressing.
+            this.removeEvent(activeSuspend);
```
gniezen: Maybe just don't add suppressed info if it's not available, instead of removing the suspend event?

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: I have confirmed that uploading a temp basal without a suppressed basal after uploading the same temp basal that includes the suppressed information will mark the basal with the suppressed info as inactive. This is not what we want. So either we need to modify the deduplicator to prefer basals with suppressed info, or keep this uploader code as-is.

lib/drivers/medtronic600/NGPHistoryParser.js
pazaan: See above comment

lib/drivers/medtronic600/medtronic600Simulator.js
pazaan: See above comment

lib/drivers/medtronic600/NGPHistoryParser.js
gniezen: OK, let's keep it as-is for the moment.

PR #691, opened 2018-08-21
Don't try to upload blob when using blob loader

gniezen:
What it says on the tin. No Trello card is fix is a one-liner and it's not user facing.


gniezen:
Merging as this does not need to be push to prod right away and doesn't need QA.

PR #698, opened 2018-09-03
Add new insulin settings for 6-series 

pazaan:


pazaan:
[Additional pump settings](https://trello.com/c/7hyUrl24/415-additional-pump-settings)

pazaan:
I don't have any tests for pump settings, since it's just a direct call to `builder.makePumpSettings`, and it wouldn't really be adding any value that I can see. Or am I missing something?

gniezen:
Never mind. I just had a look at why I had to change the unit tests for the 5/7-series, and it was mainly because I now call `.done()` in the simulator.

PR #701, opened 2018-09-06
Don't record an empty client.private entry

gniezen:
See attached Trello card


gniezen:
[Uploader shouldn't send `client.private` when there's no content](https://trello.com/c/0IAOAEXl/466-uploader-shouldnt-send-clientprivate-when-theres-no-content)

PR #703, opened 2018-09-06
Rollbar should always record username

gniezen:
See attached Trello card for details


gniezen:
[Rollbar should always record username](https://trello.com/c/NAUHEvHf/467-rollbar-should-always-record-username)

gniezen:
Merging #701 into this branch so that they can be released together.

PR #713, opened 2018-10-05
Allow for silent option in installer on Windows 8/8.1/10

gniezen:


gniezen:
[Windows installer for Uploader should install USB driver silently](https://trello.com/c/aFhdUDcq/474-windows-installer-for-uploader-should-install-usb-driver-silently)

PR #716, opened 2018-10-10
Fix for Verio & Verio Flex on macOS Mojave

gniezen:
Unfortunately the diff does not show the actual code changes, as these are inside the .zip. For code review, you'll have to extract the.zip and compare against master.


gniezen:
@pazaan OK, our scripts are now outside of the .zip file

resources/mac/helpers/org.tidepool.disk-permissions/org.tidepool.disk-permissions.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+for DEVICE in `ls /dev/rdisk*`; do
+    diskutil info $DEVICE | grep -q "LifeScan Media"
+    if [ "$?" == "0" ]; then
```
pazaan: Could be `if [ $? -eq 0 ]`

resources/mac/helpers/org.tidepool.disk-permissions/install.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+BASE_PATH="$(cd "$(dirname "${0}")"; pwd)"
+HELPER_SCRIPT="org.tidepool.disk-permissions.sh"
+HELPER_SCRIPT_PATH="/Library/PrivilegedHelperTools"
+PLIST_NAME="org.tidepool.disk-permissions.plist"
+PLIST_PATH="/Library/LaunchDaemons"
+
+if [ "${1}" = "--uninstall" ]; then
```
pazaan: The more standard way of parsing args is using `case`. Not a blocker.

resources/mac/helpers/org.tidepool.disk-permissions/install.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+BASE_PATH="$(cd "$(dirname "${0}")"; pwd)"
```
pazaan: This would set `BASE_PATH` to `pwd` even if the `cd` failed. To mitigate this, you can make it use `&&` instead of `;` (which would make `BASE_PATH` be `""`). You could then validate that the `cd` was valid by checking `-n $BASE_PATH` (or throw an error if `-z $BASE_PATH`).
It may also not hurt to add to redirect STDERR to /dev/null for the `cd` so that the error doesn't print out (since this script executes no matter what anyway).
So, in it's entirety:
```
BASE_PATH="$(cd "$(dirname "${0}")" 2>/dev/null && pwd)"
if [ -z $BASE_PATH ]; then
   echo "Well, that went poorly"
   exit 1
fi
```

gniezen: Should that be `test -z` ?

resources/mac/helpers/org.tidepool.disk-permissions/install.sh
gniezen: I agree that `case` would be better, but let's make that change if/when we add more than two args.

resources/mac/helpers/org.tidepool.disk-permissions/org.tidepool.disk-permissions.sh
gniezen: :+1:

PR #717, opened 2018-10-16
buildDualBolusRecords now consistent with other bolus build records

pazaan:
- Fixes: https://trello.com/c/bZMPsQYV
Also refactored, because no matching record is a valid condition, so
reflect that in the code (no longer an error that is caught)

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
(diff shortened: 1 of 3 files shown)


pazaan:
@gniezen I know you already approved, but can you please review my latest commit?

pazaan:
Re-opened as https://github.com/tidepool-org/chrome-uploader/pull/721, due to a premature merge.

PR #718, opened 2018-10-16
Rollbar server root

krystophv:
Configuration setting to allow for proper linkage of stacktrace filename/line numbers via sourcemaps when using built output. 

See: https://docs.rollbar.com/docs/source-maps#section-can-i-still-set-up-source-linking-when-using-source-maps-


gniezen:
@krystophv As this is not user-facing and almost a one-liner, and doesn't currently have a Trello card associated with it, should we maybe just merge it into master for the next release?

krystophv:
@gniezen Yeah, I feel like this sort of completely non-user facing config change can follow that process and just get merged in/go out with some other feature.

gniezen:
Going to merge this one as discussed above.

PR #720, opened 2018-10-17
Revert "buildDualBolusRecords now consistent with other bolus build records

pazaan:


gniezen:
Going to merge this, as I'll be releasing some other work to prod first while this is in testing.

PR #722, opened 2018-10-22
5/7-series CGM delta upload fix

gniezen:
See https://trello.com/c/yRKhiyqw for details


gniezen:
[5/7-series: Only most recent CGM data with delta upload if more than 32 pages of CGM data available](https://trello.com/c/yRKhiyqw/476-5-7-series-only-most-recent-cgm-data-with-delta-upload-if-more-than-32-pages-of-cgm-data-available)

PR #734, opened 2018-11-28
Merging master into develop

gniezen:
There are still a couple of PRs that were created against master instead of develop. This is to merge the changes that went to prod recently back into develop. They've already been reviewed, but due to branch protection we can't just merge it in.


gniezen:
Hmm, for some reason I can't merge because GitHub wants to merge develop back into master, which is not what we want. Maybe I should just disable branch protection temporarily and do this manually?

pazaan:
@gniezen Yeah, I'm okay with doing that to get it done.

PR #735, opened 2018-12-03
Verio & Verio Flex support for Windows

gniezen:
See https://trello.com/c/QvV3uNlH for details

### DIFF
--- a/lib/drivers/onetouch/oneTouchVerio.js
+++ b/lib/drivers/onetouch/oneTouchVerio.js
@@ -289,7 +289,7 @@ class BlockDevice {
       debug('Drives:', JSON.stringify(drives, null, 4));
       drives.forEach((drive) => {
         if (drive.description && drive.description.includes('LifeScan') && !drive.system) {
-          if ((process.platform === 'win32') && drive.mountpoints) {
+          if ((process.platform === 'win32') && drive.mountpoints.length > 0) {
             this.devicePath = '\\\\.\\'.concat(trimEnd(drive.mountpoints[0].path, '\\'));
           } else {
             this.devicePath = drive.raw;
--- a/lib/drivers/onetouch/oneTouchVerio.js
+++ b/lib/drivers/onetouch/oneTouchVerio.js
@@ -16,9 +16,11 @@
  */
 
 import fs from 'fs';
-import { assign, clone, invert } from 'lodash';
+import { assign, clone, invert, trimEnd } from 'lodash';
 // eslint-disable-next-line
 import drivelist from 'drivelist';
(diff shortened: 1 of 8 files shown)


gniezen:
[[uploader] Verio & Verio Flex on Windows](https://trello.com/c/QvV3uNlH/501-uploader-verio-verio-flex-on-windows)

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -318,16 +332,17 @@ class BlockDevice {
 
   closeDevice(callback) {
     if (this.fileHandle) {
-      fs.close(this.fileHandle);
-      this.fileHandle = null;
-    }
-    if (callback) {
+      fs.close(this.fileHandle, () => {
+        this.fileHandle = null;
+        return callback();
```
DorianScholz: I think this should get a check before calling callback, since it may not be defined when called from e.g. checkDevice. (Same as done in the else clause below.)

test/app/reducers/misc.test.js
```diff
@@ -70,9 +70,9 @@ describe('misc reducers', () => {
       });
       let expectedResult = _.pickBy(devices, filterDevicesFn('win'));
       expect(actualResult).to.deep.equal(expectedResult);
-      // at least one device is unavailable on Windows, so available devices should be less than
+      // at least one device may be unavailable on Windows, so available devices should be less than
```
DorianScholz: The comment should also be changed to reflect the condition change to "at most".

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -16,9 +16,11 @@
  */
 
 import fs from 'fs';
-import { assign, clone, invert } from 'lodash';
+import { assign, clone, invert, trimEnd } from 'lodash';
 // eslint-disable-next-line
```
pazaan: Why are the eslint-disables required for the following 2 modules?

lib/drivers/onetouch/oneTouchVerio.js
gniezen: Good catch! I'm also going to replace `fs.close` with `fs.closeSync` to simplify the code a little bit.

lib/drivers/onetouch/oneTouchVerio.js
gniezen: Otherwise we get a `Missing file extension for "drivelist" (import/extensions)` linting error. I'm guessing this is because they are native modules.

I'm going to add the `ignorePackages` option to `.eslintrc` so that we don't need the eslint-disables.

PR #738, opened 2018-12-20
Consistent error codes

jweismann:
I noticed an inconsistency in the error-codes shown on the PDM and the ones recorded in the data object of the IBF parser. It seems to be an off-by-one issue that is fixed by a tiny change in the field format.

Thanks a lot for sharing the source code for this extremely valuable IBF parser.

Cheers


gniezen:
@jweismann Would you mind adding a comment on this PR to state you agree to the Tidepool Contributor License Agreement? See the following for how to go about doing this: https://developer.tidepool.org/contributors/

jweismann:
@gniezen I agree to the Tidepool Contributor License Agreement. Thanks!

gniezen:
[New Omnipod Dash](https://trello.com/c/M1wDdLw5/452-new-omnipod-dash)

gniezen:
PR is now successfully passing CI checks after #772 was merged, where we now avoid forked builds but still allow units tests and linting to run.

This PR should be merged together with new OmniPod Dash work (see linked Trello card), so that it can be tested together.

PR #719, opened 2018-10-17
[WIP] Goodbye, Jellyfish 👋

gniezen:
What it says on the tin.

PR #751, opened 2019-01-15
stop throwing away good errors

krystophv:
Took a moment between tasks to try to address our tendency to create a new `Error` object and selectively copy over properties from a caught `Error` in order to have something friendly to display to the user. 

I saw two mechanisms that would probably work - (1) just attach the caught error to the new error and then have the error logging look for this property and log to Rollbar if it exists or (2) modify most of the functions to accept a new parameter which would potentially be an `originalError`. (1) seemed like a less intrusive approach.

Submitting for thoughts/review.

### DIFF
--- a/app/actions/async.js
+++ b/app/actions/async.js
@@ -195,7 +195,8 @@ export function doCareLinkUpload(deviceKey, creds, utc) {
           details: err.message,
           utc: actionUtils.getUtc(utc),
           code: 'E_FETCH_CARELINK',
-          version: version
+          version: version,
+          originalError: err
         };
         dispatch(syncActions.fetchCareLinkFailure(errorText.E_FETCH_CARELINK));
         return dispatch(syncActions.uploadFailure(fetchErr, fetchErrProps, targetDevice));
@@ -255,7 +256,8 @@ export function doDeviceUpload(driverId, opts = {}, utc) {
           details: err.message,
           utc: actionUtils.getUtc(utc),
           code: 'E_SERIAL_CONNECTION',
-          version: version
+          version: version,
+          originalError: err
         };
         return dispatch(syncActions.uploadFailure(displayErr, deviceDetectErrProps, targetDevice));
       }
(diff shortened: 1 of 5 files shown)


app/actions/async.js
```diff
@@ -195,7 +195,8 @@ export function doCareLinkUpload(deviceKey, creds, utc) {
           details: err.message,
           utc: actionUtils.getUtc(utc),
           code: 'E_FETCH_CARELINK',
-          version: version
+          version: version,
```
gniezen: Let's leave the Carelink code as is, given it's been deprecated. Part of #719 is to remove the Carelink driver completely, so let's not introduce new merge conflicts.

app/utils/errors.js
```diff
@@ -30,7 +30,8 @@ const errorProps = {
   sessionTrace: 'Session Trace',
   stringifiedStack: 'Stack Trace',
   utc: 'UTC Time',
-  version: 'Version'
+  version: 'Version',
+  originalError: 'Original Err'
```
gniezen:
```suggestion
  originalError: 'Original Error'
```

lib/core/api.js
```diff
@@ -728,6 +734,9 @@ api.errors.log = function(error, message, properties) {
     } else {
       rollbar.error(error);
     }
+    if(error.originalError){
+      rollbar.error(error.originalError);
```
gniezen: This means there will be multiple error reports for the same event, right? This is already happening to some extent, so wouldn't be too bad, but I'm wondering if we should rather just log the original error when available, adding the new error object as a custom field (which will then show up under `message.extra`). As an example, see line 733 where the blob ID is added as a custom field.

lib/core/api.js
```diff
@@ -723,11 +729,15 @@ api.errors.log = function(error, message, properties) {
   api.log('GET /errors');
 
   if (rollbar) {
-    if (error.data && error.data.blobId) {
-      rollbar.error(error, { blobId: error.data.blobId });
-    } else {
-      rollbar.error(error);
+    var extra = {};
```
krystophv: @gniezen - I reorganized most of this. Could use another glance.

gniezen: Looks good! :+1:

krystophv:
No trello card, just developer experience improvement (like the rollbar server root config change).

gniezen:
OK, that's fine, I'll just merge it in then, thanks!

PR #754, opened 2019-01-17
Record IOB when meal or correction IOB is zero

gniezen:


gniezen:
[OmniPod IOB (possibly) miscalculated](https://trello.com/c/tWxKUIcy/527-omnipod-iob-possibly-miscalculated)

PR #757, opened 2019-01-21
Rolling back electron to v3 due to node crypto issue

gniezen:
What it says on the tin.


gniezen:
[Update dependencies (including Electron)](https://trello.com/c/2DYR2Xlx/517-update-dependencies-including-electron)

PR #758, opened 2019-01-21
always reset resuming event after basal

gniezen:
See attached Trello card.


gniezen:
[Negative duration for automated suspend-resume](https://trello.com/c/GI0fUhFD/536-negative-duration-for-automated-suspend-resume)

PR #784, opened 2019-02-21
Calculator boluses not recorded that includes carbs but no insulin delivery

gniezen:


gniezen:
[[bug] Carbs not visualized for calculator bolus that includes carbs but no insulin delivery](https://trello.com/c/dZsluC3F/566-bug-carbs-not-visualized-for-calculator-bolus-that-includes-carbs-but-no-insulin-delivery)

gniezen:
We still use `finalBasal` in some of the other drivers too, so I think it's ok. I just changed it for 5/7-series because for that specific section it now does more than just fix up the final basal.

PR #792, opened 2019-03-01
Fixes 600-series settings history and adds additional pump settings to history

gniezen:
What it says on the tin.


gniezen:
[600-series driver does not handle changes in pump history correctly](https://trello.com/c/WoixEsvg/577-600-series-driver-does-not-handle-changes-in-pump-history-correctly)

PR #804, opened 2019-03-20
Collection of Abbott Libre fixes

gniezen:
See attached Trello cards for details.


gniezen:
[[uploader] TypeError when there are no records on a Libre](https://trello.com/c/JZASCL2P/596-uploader-typeerror-when-there-are-no-records-on-a-libre)

gniezen:
[[uploader] Upload Libre "scan" records as SMBG record](https://trello.com/c/O9i3tqCA/595-uploader-upload-libre-scan-records-as-smbg-record)

gniezen:
[[uploader] Libre date&time changes are stored as two separate time change events](https://trello.com/c/gei6SOfn/315-uploader-libre-datetime-changes-are-stored-as-two-separate-time-change-events)

lib/drivers/abbott/freeStyleLibreData.js
```diff
@@ -233,33 +233,59 @@ export default class FreeStyleLibreData {
...
+        record.timeChangeFields.oldReaderTime,
+        record.timeChangeFields.oldUserTimeOffset,
+      );
+
+      const nextChange = timeChanges[i + 1];
+
+      if (nextChange) {
+        const distance = nextChange.headerFields.recordNumber - record.headerFields.recordNumber;
+        // even if combined date/time changes are made directly after another,
+        // they should still be more than 50 records apart
```
pazaan: Where does this magic number come from? Is that defined in a spec? As a maintainer, if this came from a spec, and the spec was wrong, I would take a different action than trying a different magic number :)

lib/drivers/abbott/.eslintrc
```diff
@@ -11,7 +11,8 @@
       {
         "allowForLoopAfterthoughts": true
       }
-    ]
+    ],
+    "lodash/prefer-lodash-method": "off"
```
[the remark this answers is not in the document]
gniezen: When Dorian developed the Libre driver, he used ES6 methods instead of Lodash methods. Since all the ES6-ified files in that folder is for the Libre, I thought making it generally true for that folder is fine. To be honest, I wouldn't mind making this generally true for all our ES6-ified drivers - I don't quite see why we want to prefer Lodash methods, given that it's an Electron app and we know the ES6 methods will always be available?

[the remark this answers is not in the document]
pazaan: No problem. I was just checking. I’d been following a previous convention I’d found in the code to prefer lodash. 
I’m happy to make ES6 the new convention :)

lib/drivers/abbott/freeStyleLibreData.js
gniezen: Nope, not from spec, but from observation. Which is why I defined it as a constant in case we want to change it in future. I wish there was a better way of doing this, but since the spec does not distinguish between date changes and time changes (but still records them separately), I resolved to just looking if there are date/time change events near each other.

We follow a similar approach in the Tandem driver:

https://github.com/tidepool-org/chrome-uploader/blob/develop/lib/drivers/tandem/tandemDriver.js#L1492-L1495

pazaan: Understood. It’s a bit disconcerting how much trial and error is involved in driver development.

pazaan: Understood. It’s a bit disconcerting how much trial and error is involved in driver development.

PR #821, opened 2019-04-03
Toggle between full/delta uploads for Tandem and Medtronic 5/7-series

gniezen:


gniezen:
[For Tandem uploads, toggle between full & delta uploads](https://trello.com/c/pGocpkbs/609-for-tandem-uploads-toggle-between-full-delta-uploads)

PR #833, opened 2019-04-11
Merge back bug fixes in 2.13.0 release branch

gniezen:
I've fixed two things discovered during QA testing:

- we should not store `basal.temporary.type` in `pumpSettings`, as it's set per temp basal and not as part of the settings
- `bolus.calculator.enabled` wasn't being updated correctly


gniezen:
Just realised I can't merge update from `develop` until we've actually pushed this release branch to production first, but I want @pazaan to at least review the bug fixes before releasing this.

PR #834, opened 2019-04-11
Hotfix: ensure TrueMetrix checksum is not more than 2 bytes long 

gniezen:
What it says on the tin.


gniezen:
[TrueMetrix Error: Checksums not matching.](https://trello.com/c/vHqdHJVc/647-truemetrix-error-checksums-not-matching)

PR #867, opened 2019-06-04
Sign Windows drivers with new certifcate

gniezen:


gniezen:
[Update code signing certificates](https://trello.com/c/Uv8NwRg4/600-update-code-signing-certificates)

gniezen:
Note that we had to have a number of Tidepoolers (more than 10) install the Uploader before release, for the code signing certificate (HS, not EV) to build up enough trust with MS so that it doesn't throw a warning pop-up. This affects both the app and driver installers, but not the Windows 10 drivers themselves as they are signed with an EV cert.

PR #876, opened 2019-06-19
Don't record BG values auto-populated by CGM as manual SMBG

gniezen:


gniezen:
[Tandem X2 wizard bolus with BG auto-populated from CGM shows up as manual SMBG reading](https://trello.com/c/G8UMlLrs/718-tandem-x2-wizard-bolus-with-bg-auto-populated-from-cgm-shows-up-as-manual-smbg-reading)

PR #887, opened 2019-06-25
600-series: Allow extended boluses with 0 normal (UPLOAD-8)

pazaan:
The Medtronic 600-series driver simulator previously discarded
*any* bolus events with a normal value of 0.
However, this meant that extended boluses with a zero component
were also discarded. This fixes that bug.

* Fixes https://trello.com/c/s67Yc2qP (for 600-series driver)

### DIFF
--- a/lib/drivers/medtronic600/medtronic600Simulator.js
+++ b/lib/drivers/medtronic600/medtronic600Simulator.js
@@ -399,7 +399,7 @@ class Medtronic600Simulator {
     return _.filter(this.events, (event) => {
       // filter out zero boluses
       if (event.type === 'bolus') {
-        if (event.normal === 0 && !event.expectedNormal) {
+        if (event.subType === 'normal' && event.normal === 0 && !event.expectedNormal) {
           return false;
         }
       } else if (event.type === 'wizard') {


pazaan:
@gniezen - FYI - this issue will also exist for the 5/7-series driver. See https://github.com/tidepool-org/chrome-uploader/blob/develop/lib/drivers/medtronic/medtronicSimulator.js#L806

gniezen:
In the 5/7-series driver, if there's no normal portion, it's recorded as a square bolus instead (https://github.com/tidepool-org/chrome-uploader/blob/develop/lib/drivers/medtronic/processData.js#L373-L375). Conversely, if the extended portion is zero we record it as a normal bolus (https://github.com/tidepool-org/chrome-uploader/blob/develop/lib/drivers/medtronic/processData.js#L373-L375).

Do you want to duplicate this behaviour in the 600-series driver, or do you prefer to still record it as a dual bolus? Looks like there is precedent for recording a zero normal portion, if you look at the notes at http://developer.tidepool.org/data-model/device-data/types/bolus/dual-square.html#normal

ginnyyadav:
[Backend Regression (?) - Dual Wave Boluses do not store correctly on the backend.](https://trello.com/c/s67Yc2qP/719-backend-regression-dual-wave-boluses-do-not-store-correctly-on-the-backend)

PR #928, opened 2019-08-08
Use new metadata API to get user list (UPLOAD-57)

krystophv:
Updates `getUploadGroups` API call to use the updated `getAssociatedUsersDetails` call instead of iterating through `findProfile` fetches for all available upload users.

Addresses [UPLOAD-57]

[UPLOAD-57]: https://tidepool.atlassian.net/browse/UPLOAD-57


krystophv:
Adding @cbwebdevelopment as reviewer as it's just an API change

PR #951, opened 2019-09-06
Updates to run in local Tilt/K8s environment [WEB-210]

cbwebdevelopment:
Part of [WEB-210]

To be used alongside tidepool-org/development#24

Because the k8s internal routing always uses the same port, I've updated the "Local" config to always map to port 3000.

This would of course not work in the current Docker Compose stack we use, so, while I'd like to this PR reviewed ahead-of-time, I won't actually merge it into `develop` until we officially move over to the local K8s env and deprecate the Compose stack.

[WEB-210]: https://tidepool.atlassian.net/browse/WEB-210


cbwebdevelopment:
Closing this PR in favor of https://github.com/tidepool-org/uploader/pull/959

PR #954, opened 2019-09-12
added install script for linux

nathanac2013:
it doesnt cover all the bases but it covers the majority , it has to be ran on a debian based system so debian elementeryos ubuntu the likes like that cuz thats the package manager i used to install the missing dependencies and stuff


atombrella:
There's a PR #1244 that wraps the build in a docker container. Maybe this would solve also this request?

gniezen:
@atombrella That's true, I'm going to close this PR in favour of #1244. This script requires Uploader to have superuser access, while the Docker container seems like a safer setup, and supports all Linux distros.

PR #973, opened 2019-10-16
Adds support for Bluetooth LE meter (CareSens N Premier BLE) - UPLOAD-12

gniezen:
Addresses [UPLOAD-12].

[UPLOAD-12]: https://tidepool.atlassian.net/browse/UPLOAD-12


lib/drivers/bluetoothLE/bluetoothLEDriver.js
```diff
@@ -0,0 +1,196 @@
...
+   },
+
+   async getConfigInfo(progress, data, cb) {
+     debug('in getConfigInfo', data);
+     progress(0);
+
+     try {
+       _.assign(cfg.deviceInfo , await cfg.deviceComms.ble.getDeviceInfo());
+       cfg.deviceTags = ['bgm'];
+       cfg.deviceInfo.deviceId = `${[cfg.deviceInfo.manufacturers]}-${cfg.deviceInfo.model}-${remote.getGlobal('bluetoothDeviceId')}`;
```
pazaan: What does `${[cfg.deviceInfo.manufacturers]}` return?

gniezen: As you know, we usually hard-code the first part of the `deviceId` to be the name of the manufacturer. The Bluetooth driver is the first time we have a driver that can be used to upload devices from different manufacturers, so I pull in the manufacturer name from the `deviceInfo.manufacturers` array that gets added to the the upload record. 

In the case of the CareSens meter, `cfg.deviceInfo.manufacturers` will be `i-SENS`.

lib/drivers/bluetoothLE/bluetoothLEDriver.js
```diff
@@ -0,0 +1,196 @@
...
+       cfg.builder.setDefaults({ deviceId: cfg.deviceInfo.deviceId });
+
+       cfg.deviceComms.ble.once('numberOfRecords', (result) => {
+         debug('Number of records:', result);
+       });
```
nhamming: As you most likely know, at some point this becomes meaningless. Should one get the number of records if that doesn't improve the program flow?

lib/drivers/bluetoothLE/bluetoothLEDriver.js
```diff
@@ -0,0 +1,196 @@
...
+           code: 'bg/out-of-range',
+           value: 'high',
+         };
+       } else if (record.value < 20) {
+         record.value = 19;
+         annotation = {
+           code: 'bg/out-of-range',
+           value: 'low',
+         };
+       }
```
nhamming: Are these HI and LO values consistent for all potentially supported BG meters? If not, should they be part of the device configuration to be more specific?

lib/drivers/bluetoothLE/bluetoothLEDriver.js
```diff
@@ -0,0 +1,196 @@
...
+
+       if (record.type !== 10) { //check that it's not control solution
```
nhamming: Similar comment here about the value `10` and if this is specific to a certain meter or able to support all sorts of meters.

nhamming: Is there a benefit of uploading control solution records?

gniezen: This does indeed apply to all meters - according to the spec, the type field in the glucose measurement characteristic will be set to 10 if it's control solution.

We would like to upload control solution records in future, but at the moment it is not supported by the backend. We've also not had a single request from users to see control solution records, which probably explains why this hasn't been prioritized on the backend.

lib/drivers/bluetoothLE/bluetoothLEDriver.js
gniezen: This is something I wish was in the BLE glucose spec, so that meters would have to report their HI/LO thresholds, as it turns out none of them do.

Almost all the meters we currently support in the Uploader have a 20/600 HI/LO threshold, but if we do find one that differs, I'll have to figure out how to update this logic. I did check that the CareSens N Premier BLE meter, which is the first meter to use this driver, does indeed have 20/600 thresholds.

Ideally meters would report +/- infinity when it's out-of-range, right? Unfortunately they don't always do that, and you don't want to record a value that you know is out of spec for that meter (sometimes the only place you can find the actual thresholds is in the meter user manual).

lib/drivers/bluetoothLE/bluetoothLEDriver.js
gniezen: You're right, I should remove this as I'm not using that value anymore.

PR #1001, opened 2019-11-13
Webpack update [UPLOAD-3] [UPLOAD-87]

krystophv:
In an effort to solve UPLOAD-3, we figured out that UPLOAD-87 needed to also be taken care of.

Highlights of this work: 
- Update `webpack` to 4.x
- Update `babel` to 7.x
- Update many smaller dependencies to their latest versions
- Move to `node` 10.17.0 (required by latest eslint)
- Move to `jest` as a test runner (largely via `jest-codemods`)
- Simplify/clarify webpack config to differentiate between `dev` and `prod` builds
- Bring many aspects of the project more inline with modern `electron-react-boilerplate`


package.json
```diff
@@ -1,80 +1,85 @@
...
-    "build-main-quiet": "cross-env NODE_ENV=production node -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.electron.js",
+    "build-main-quiet": "cross-env NODE_ENV=production webpack --config webpack.config.main.prod.babel.js",
     "build-renderer": "yarn build-renderer-quiet --progress --profile --colors",
-    "build-renderer-quiet": "cross-env NODE_ENV=production node -r babel-register ./node_modules/webpack/bin/webpack --config webpack.config.production.js",
-    "build": "npm run build-main && npm run build-renderer",
-    "build-quiet": "npm run build-main-quiet && npm run build-renderer-quiet",
-    "start": "cross-env NODE_ENV=production electron ./app/",
-    "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development",
+    "build-renderer-quiet": "cross-env NODE_ENV=production webpack --config webpack.config.renderer.prod.babel.js",
+    "build": "concurrently \"yarn build-main\" \"yarn build-renderer\"",
```
gniezen: Nice, this should speed up things a bit!

package.json
```diff
@@ -1,80 +1,85 @@
...
+    "identity-obj-proxy": "3.0.0",
     "immutability-helper": "3.0.1",
     "is-electron": "2.2.0",
     "lodash": "4.17.15",
     "lzo-decompress": "0.1.2",
     "pl2303": "0.0.5",
     "plist": "3.0.1",
-    "prop-types": "15.7.0",
+    "prop-types": "15.7.2",
     "react": "16.4.0",
```
gniezen: Will updating React to 16.11.0 be too heavy of a lift for this PR?

package.json
```diff
@@ -173,81 +178,99 @@
...
+    "@babel/polyfill": "7.7.0",
+    "@babel/preset-env": "7.7.1",
+    "@babel/preset-react": "7.7.0",
+    "@babel/register": "7.7.0",
+    "@babel/runtime-corejs2": "7.7.2",
+    "@jest-runner/electron": "2.0.2",
     "@ronomon/direct-io": "2.4.3",
     "aws-sdk": "2.472.0",
-    "babel-core": "6.26.3",
+    "babel-core": "7.0.0-bridge.0",
```
gniezen: What do we need to do to get rid of this bridge in future?

package.json
```diff
@@ -173,81 +178,99 @@
...
     "concurrently": "5.0.0",
     "cross-env": "6.0.3",
-    "css-loader": "0.28.11",
+    "css-loader": "3.2.0",
     "devtron": "1.4.0",
     "difflet": "1.0.1",
     "drivelist": "6.4.3",
     "electron": "3.1.1",
-    "electron-builder": "21.2.0",
+    "electron-builder": "22.1.0",
```
gniezen: This is still marked as a pre-release, so maybe we should hold off on this if it's not required to make things work?

package.json
```diff
@@ -173,81 +178,99 @@
...
+    "less": "3.10.3",
+    "less-loader": "5.0.0",
+    "mini-css-extract-plugin": "0.8.0",
     "minimist": "1.2.0",
-    "mocha": "6.2.2",
     "node-hid": "0.7.9",
     "object-invariant-test-helper": "0.1.1",
-    "open": "0.0.5",
-    "react-hot-loader": "4.3.2",
+    "open": "7.0.0",
```
gniezen: Nice! :grinning:

package.json
```diff
@@ -1,80 +1,85 @@
...
     "classnames": "2.2.6",
     "commander": "4.0.1",
+    "core-js": "2.6.10",
     "cp2102": "0.0.3",
     "decompress": "4.2.0",
-    "electron-debug": "2.2.0",
+    "electron-debug": "3.0.1",
     "electron-is-dev": "1.1.0",
+    "electron-log": "3.0.9",
+    "electron-updater": "4.2.0",
```
gniezen: Why did this move from `devDependencies` to `dependencies`?

app/index.js
```diff
@@ -30,11 +31,15 @@ ipcRenderer.on('action', function(event, action) {
   store.dispatch(action);
 });
 
+const AppContainer = process.env.PLAIN_HMR ? Fragment : ReactHotAppContainer;
```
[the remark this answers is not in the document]
krystophv: This would be set by the user in the event that they wanted to not use the react hot module replacement mechanism, was directly from the boilerplate.

package.json
krystophv: As long as we have the cores available 🤞

package.json
krystophv: Maybe not? It was just outside of the original scope and I didn't want to unnecessarily bulk up the workload. If we think now would be the right time to approach it, then I'm amenable to it.

package.json
krystophv: Output from `yarn why babel-core` reveals `Hoisted from "@jest-runner#electron#@jest-runner#rpc#jscodeshift#babel-core"`, as the only dependency still pointing at `babel-core` instead of `@babel/core`, so once that dependency is updated to point to the new dependency path, we can drop this. (additionally `babili-webpack-plugin` was pointing at it, but is an unused dependency so I'll remove it)

package.json
krystophv: Yeah, I'm fine with that. I was just trying for as many small updates as I could get and presumed it wasn't a pre-release since it was a final version number (I didn't look too closely).

package.json
krystophv: Right? The reason all this started.

package.json
krystophv: Because it's one of the packages that is required _after_ the build process directly by `main.dev.js` (and this is where it lives in the boilerplate project). It may also have to do with the fact that the boilerplate uses the Webpack DLL splitting feature for all of the things under `dependencies` to help with build time (mostly repeat-build time I believe) and doesn't use the two `package.json` layout anymore.

package.json
```diff
@@ -1,80 +1,85 @@
 {
   "name": "tidepool-uploader",
-  "version": "2.22.2",
+  "version": "2.22.3-webpack-update",
```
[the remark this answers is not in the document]
krystophv: Ah yeah, I mostly just changed it so I could verify that CI build artifacts were actually working

package.json
gniezen: I'm fine with doing it as a separate step.

PR #1003, opened 2019-11-14
Ensure everything works under macOS Catalina (UPLOAD-112)

gniezen:
Addresses:

- [UPLOAD-112]

[UPLOAD-112]: https://tidepool.atlassian.net/browse/UPLOAD-112


gniezen:
@krystophv Please have a look at this PR when you get a chance. It removes the kernel extensions from the macOS app, as Apple are clearly on a path to phase them out completely, so we depend on our user-space drivers instead if these kernel extensions are not installed.

PR #1034, opened 2020-01-02
Search for most recent IBF file on Windows (UPLOAD-118)

gniezen:
Addresses [UPLOAD-118].


gniezen:
@pazaan Can you also review 3f1bc21 please? Since the version of Powershell on Windows 7 is too old for connecting to the PDMs, we'll have to fall back to the old way of doing things on Windows 7.

PR #1039, opened 2020-01-10
Support for Freestyle Precision/Optium Neo (UPLOAD-218)

gniezen:
Addresses:

- [UPLOAD-218]


lib/drivers/abbott/freeStyleLibreProtocol.js
```diff
@@ -610,13 +610,13 @@ export default class FreeStyleLibreProtocol {
...
 
-      dbRecords = dbRecords.split('\r\n');
+      dbRecords = dbRecords.split('\r\n').filter(Boolean);
 
       if (dbRecordCount !== dbRecords.length) {
         return errorCallback(new Error(`Invalid database record count: ${dbRecordCount} != ${dbRecords.length}`), data);
       }
 
-      return successCallback(null, dbRecords);
+      return successCallback(dbRecords);
```
pazaan: This signature is different. Does this fix a bug? Was data previously not being returned?

lib/drivers/abbott/abbottFreeStyleNeo.js
```diff
@@ -0,0 +1,415 @@
...
+      },
+      (err) => {
+        debug('requestTextResponse Error:', err);
+        reject(err);
+      });
+    });
+  }
+
+  async getSerialNumber() {
+    const serialNumber = await this.commandResponse('$serlnum?');
```
pazaan:
```suggestion
    return this.commandResponse('$serlnum?');
```

And delete the next line.

lib/drivers/abbott/abbottFreeStyleNeo.js
```diff
@@ -0,0 +1,415 @@
...
+        );
+      }).catch((error) => {
+        debug('Error in getConfigInfo: ', error);
+        return cb(error, null);
+      });
+    },
+
+    fetchData(progress, data, cb) {
+      debug('in fetchData', data);
+      (async () => {
```
pazaan: If you wanted to, you can remove the "then/catch" Promise handling with:
```
(async () => {
  try {
    data.records = await driver.getRecords();
    progress(100);
    return cb(null, data);
  } catch (error) {
    return cb(error, null);
  }
})()
```

lib/drivers/abbott/freeStyleLibreProtocol.js
gniezen: `getDbRecords` is not actually being used in the Libre driver, as the Libre uses the binary protocol and not the text protocol. This is documented in the comments on lines 623 and 630, but not in the function itself. To be sure, I'm going to add a smoke test of the Libre reader to the test strategy.

lib/drivers/abbott/abbottFreeStyleNeo.js
gniezen: Yeah, I like that idea.

lib/drivers/abbott/abbottFreeStyleNeo.js
gniezen: :+1:

PR #1047, opened 2020-01-24
Read exchange carb ratios for current pump settings on 630G/640G (UPLOAD-282)

gniezen:
https://tidepool.atlassian.net/browse/UPLOAD-282


lib/drivers/medtronic600/medtronic600Driver.js
```diff
@@ -1325,18 +1325,29 @@ class BasicNgpParametersCommand extends TransmitPacketRequest {
...
+      return ratios;
+    };
+
+    let ratios = readRatios(0x06, 10);
+    if (ratios.every(item => item.amount === 0)) {
+      // if carb ratios are all zero, they are in exchanges instead of
+      // grams and stored in a different UInt32BE
+      // Since pumpSettings doesn't support `exchanges` as a unit, convert all
+      // exchanges to grams.
+      ratios = readRatios(0x0A, 1000 * 15.0);
```
pazaan: Are exchanges always 15g? I think I've seen instances in the past where it was 10g? Isn't it essentially up to the user?

gniezen: I based it on the the pump settings history: https://github.com/tidepool-org/uploader/blob/develop/lib/drivers/medtronic600/NGPHistoryParser.js#L1384-L1390

I guess the conversion factor stays the same per pump model? I.e., you choose either exchanges or grams, you don't choose the actual conversion factor.

PR #1050, opened 2020-01-28
Update Electron to v7 (UPLOAD-281)

gniezen:
Make yourself a cup of coffee, this one's a doozy! :coffee: I used the opportunity to upgrade both `lib/core` and the Omnipod driver to ES6, so there's a lot to review.

PR #1065, opened 2020-02-13
Fix version number

gniezen:
This one slipped through with #1050


gniezen:
Going to merge this one already as it's just a version number correction.

PR #1133, opened 2020-04-08
Multi languages (UPLOAD-336)

jehernandezrodriguez:
Creation de Uplander multi languages with 118next.

- Translation in spanish is up, i dont know if is necessary.
- We will discuss global variable in each `/app/components` for call i18next.


gniezen:
This is great, thank you @jehernandezrodriguez ! Can you run `yarn` again to generate a new `yarn.lock` file and commit that, as that is currently preventing the CI checks from running?

When I'm running it locally on my machine, I still see a number of missing translations?

image

gniezen:
Excellent, CI tests are now passing. We'll do a code review and get back to you.

jehernandezrodriguez:
Hello! Thank you, I await your response. I uploaded a few changes for translations of device instructions.

app/main.dev.js
```diff
@@ -492,3 +510,28 @@ if(!app.isDefaultProtocolClient('tidepoolupload')){
...
+          createWindow();
+      });
+    } else {
+      createWindow();
+    }
+  }
+});
+
+function setLanguage() {
+  let lng = app.getLocale();
```
gniezen, line 531: Can we have this default to English for now, and only use the locale if an environment variable (let's say,  `I18N_ENABLED` is set to `true`, similar to how it's currently set up in Tidepool Web)?

locales/en/translation.json
```diff
@@ -0,0 +1,176 @@
+{
+  "Tidepool Uploader":"",
```
gniezen, line 2: Shouldn't we be using the same for key and value everywhere in this file, to prevent everything showing up in `locales/en/translation.missing.json` again?

app/main.dev.js
jehernandezrodriguez: Hello!. Of course. Maybe you could tell me where in Tidepool Web is similar. Pls.

What I understand is this:

```
var enabled = process.env.I18N_ENABLED || false
if (!enabled) return 
```

Only configure the language by region if the environment variable is `true`. In `utils / config.i18next` it is defaulted to "en".

gniezen: Yes, that sounds right. The Tidepool Web implementation is a bit different, as it's possible to also select the language from an account form and store that as part of the user preferences, but we can add that functionality to Uploader at a later date.

mrinnetmaki:
@jehernandezrodriguez are you still actively working on this? Need any assistance?
I'm really looking forward to seeing i18n implemented in the Uploader.
Leet me know if there's anything you need help with.

jehernandezrodriguez:
@mrinnetmaki Hello, I hope to upload the local changes soon. Changes are not long. Then we check the translations.

jehernandezrodriguez:
Hi. It seems that all components already have translations. But I have the problem that in production mode i18next does not detect or find the translation files.

Maybe @gniezen  or @mrinnetmaki  can help me add or tell me the best way to add the translation files when the installers are packaged ``` .exe or .dmg```.

gniezen:
@jehernandezrodriguez I think the easiest may be to put the translation files in `resources/`, to make sure they get copied into the final package. The path will be different between development and production, so  have a look here at how to distinguish between them: https://github.com/tidepool-org/uploader/blob/master/app/utils/drivers.darwin.js#L73-L84

jlaunonen:
Hi,
with some fiddling I was able to make the branch work in all "dev", "start" and "package" modes (tested on Linux).
As suggested by @gniezen, the easiest way to make this work is to move the locales-directory resources and adjust `app/utils/config.i18next.js` accordingly. Additionally, for the translation files to actually be accessible in app package, the locales-directory must be added to the build.extraResources in `package.json`.

No additional runtime configuration should be needed for this.

So the file layout should be something like

    $projectRoot
      resources
        locales
          en
            translation.json
            translation.missing.json
          es
            translation.json
            translation.missing.json
        ...

and the backend block in configuration like

     loadPath: 'resources/locales/{{lng}}/{{ns}}.json',
     addPath: 'resources/locales/{{lng}}/{{ns}}.missing.json'

and in package.json (one added line)

    "extraResources": [
       { ... },
       "sounds/",
       "locales/"
    ]

This results the translation files being copied as is to resulting package and thus are readable (and most probably modifiable) by user. Better solutions probably need better backend for i18next loading the resources (in place of i18next-fs-backend), and/or some kind of integration with webpack.

jehernandezrodriguez:
> Hi,
> with some fiddling I was able to make the branch work in all "dev", "start" and "package" modes (tested on Linux).
> As suggested by @gniezen, the easiest way to make this work is to move the locales-directory resources and adjust `app/utils/config.i18next.js` accordingly. Additionally, for the translation files to actually be accessible in app package, the locales-directory must be added to the build.extraResources in `package.json`.
> 
> No additional runtime configuration should be needed for this.
> 
> So the file layout should be something like
> 
> ```
> $projectRoot
>   resources
>     locales
>       en
>         translation.json
>         translation.missing.json
>       es
>         translation.json
>         translation.missing.json
>     ...
> ```
> 
> and the backend block in configuration like
> 
> ```
>  loadPath: 'resources/locales/{{lng}}/{{ns}}.json',
>  addPath: 'resources/locales/{{lng}}/{{ns}}.missing.json'
> ```
> 
> and in package.json (one added line)
> 
> ```
> "extraResources": [
>    { ... },
>    "sounds/",
>    "locales/"
> ]
> ```
> 
> This results the translation files being copied as is to resulting package and thus are readable (and most probably modifiable) by user. Better solutions probably need better backend for i18next loading the resources (in place of i18next-fs-backend), and/or some kind of integration with webpack.

Thank you. I've already made the changes, let's wait for gniezen's comments.

gniezen:
Thanks both, that's exactly what I meant! :smile: 

I see the tests are now failing because it can't find the global `i18n` variable, probably because `remote.getGlobal` is only available in Electron?

jehernandezrodriguez:
> Thanks both, that's exactly what I meant! 😄
> 
> I see the tests are now failing because it can't find the global `i18n` variable, probably because `remote.getGlobal` is only available in Electron?

Hi @gniezen 
I think in `'devices.json'` , `I18next` is not initialized. I uploaded a change for devices translation when is render in view.

gniezen:
OK, so the problem is that `i18n` isn't created during the test setup, because `main.dev.js` isn't called.

I got it to work by changing line 5 of `device.js` to:

```js
let i18n = remote.getGlobal( 'i18n' );

if ((process.env.NODE_ENV === 'test') && !i18n) {
  i18n = require('i18next');
}
```
I don't know if there's a better way to set up the test in Jest so that `remote.getGlobal` returns something?

jehernandezrodriguez:
> OK, so the problem is that `i18n` isn't created during the test setup, because `main.dev.js` isn't called.
> 
> I got it to work by changing line 5 of `device.js` to:
> 
> ```js
> let i18n = remote.getGlobal( 'i18n' );
> 
> if ((process.env.NODE_ENV === 'test') && !i18n) {
>   i18n = require('i18next');
> }
> ```
> 
> I don't know if there's a better way to set up the test in Jest so that `remote.getGlobal` returns something?

Yes it's a part of the problem

I've already tried this way, but in dev and prod `inext18` it doesn't initialize because reducers start first than the app in `main.dev.js`.

We could call and initialize `'i18next'` in `devices.js`, but we have an asynchronous problem because in `devices.js` the variable 'devices' is exported before 'i18next' finishes searching for the words. I tried to wait for the 'i18next' promise but it didn't work for me.

krystophv:
Since `i18next` will always be initialized before running any app code during regular use due to it being gated by the window open, I think we can probably solve this as a test-only scenario. Experimenting locally, it seems that adding a mock for `electron` as follows (in `__mocks__/electron.js`)
```
const electron = jest.genMockFromModule('electron');

function getGlobal(string) {
    if(string === 'i18n'){
        return { t: (string) => string };
    }
}

electron.remote = { getGlobal };

module.exports = electron;
```
And adding `jest.mock('electron');` just after the `import`s in `test/app/reducers/misc.test.js` allows the tests to pass as expected.

jehernandezrodriguez:
> Since `i18next` will always be initialized before running any app code during regular use due to it being gated by the window open, I think we can probably solve this as a test-only scenario. Experimenting locally, it seems that adding a mock for `electron` as follows (in `__mocks__/electron.js`)
> 
> ```
> const electron = jest.genMockFromModule('electron');
> 
> function getGlobal(string) {
>     if(string === 'i18n'){
>         return { t: (string) => string };
>     }
> }
> 
> electron.remote = { getGlobal };
> 
> module.exports = electron;
> ```
> 
> And adding `jest.mock('electron');` just after the `import`s in `test/app/reducers/misc.test.js` allows the tests to pass as expected.

Thanks @krystophv

jlaunonen:
Hi, are there any critical things still missing or could this be merged?

gniezen:
@jlaunonen I'm going to prioritise this for code review and QA during our next sprint.

gniezen (approved):
This looks good to me! I'm going to build a test version so that we can get this through QA.

gniezen (approved):
LGTM! This will be added to our next sprint (starting today) for QA.

gniezen:
Closing this PR in favour of #1358 , created to enable QA builds since it's not an external PR.

gniezen:
Just a quick update to say that this has now shipped in Uploader v2.35.0. Thanks all!

PR #1143, opened 2020-04-16
Omnipod fixes (UPLOAD-323)

gniezen:
Addresses [UPLOAD-323], and also fixes some issues with uploading .ibf files via CLI.

[UPLOAD-323]: https://tidepool.atlassian.net/browse/UPLOAD-323


gniezen:
I'm trying to work my way through the dependency updates, but there's a lot! One blocker is that `platform-client` needs to be updated. `platform-client` is used by both Uploader and Web, and uses `superagent` to send network requests. Unfortunately the version of superagent used by `platform-client`  is so old that it does not support timeouts and all the other goodies that could avoid issues from the client side.

I found the reason why it has not been updated in https://tidepool.atlassian.net/browse/SECURITY-19, in a comment from @krystophv on 2018-09-12:

> This one is going to be non-trivial as superagent has significant breaking changes between the current version and latest, specifically surrounding error handling. This will ripple down into a "test everything that makes a network request" QA task for both Uploader and Web.

I wonder if the time has come to bite the bullet and take on this tech debt? We also need a newer version of `superagent` to do retries and timeouts with the 502/504 errors we've been experiencing.

krystophv:
@gniezen - As time goes on, QA is developing a more and more comprehensive set of automated end-to-end tests for Web. Essentially the lift for testing all Web functionality is slowly becoming smaller. It might be worth coordinating a conversation just to see how far along it is and how hard it would be to get confidence in testing for things not yet automated.

PR #1149, opened 2020-04-22
Check device time before starting to read data from Ultra 2 (UPLOAD-304)

gniezen:
Second attempt to address [UPLOAD-304].

[UPLOAD-304]: https://tidepool.atlassian.net/browse/UPLOAD-304

### DIFF
--- a/lib/drivers/onetouch/oneTouchUltra2.js
+++ b/lib/drivers/onetouch/oneTouchUltra2.js
@@ -316,23 +316,7 @@ module.exports = function (config) {
          return cb(new Error('Checksum error'));
     }
 
-    common.checkDeviceTime(cfg, function(err, serverTime) {
-      if (err == 'updateTime') {
-        cfg.deviceInfo.annotations = 'wrong-device-time';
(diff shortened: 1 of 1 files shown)


lib/drivers/onetouch/oneTouchUltra2.js
```diff
@@ -401,7 +385,26 @@ module.exports = function (config) {
       };
 
       async.retry({ times: RETRIES, interval: TIME_TO_WAIT }, sendReceive, function(err, result) {
-        callback(err, result);
+
+        common.checkDeviceTime(cfg, function(err, serverTime) {
+          if (err == 'updateTime') {
+            cfg.deviceInfo.annotations = 'wrong-device-time';
+            debug('Time set to');
```
pazaan, line 392: Either remove debug statement or add time?

gniezen: That line confused me too, until I ran the code and realised that it's printing the time on the next line (in the `setDateTime` method).

PR #1163, opened 2020-05-01
Upload Verio & Verio Flex meters on macOS Catalina (UPLOAD-232)

gniezen:
Addresses [UPLOAD-232].

[UPLOAD-232]: https://tidepool.atlassian.net/browse/UPLOAD-232


gniezen:
OK, should be ready for review now.

lib/drivers/onetouch/catalina-sudo/sudo-askpass.osascript.js
```diff
@@ -0,0 +1,41 @@
...
+/* eslint-disable */
+
+ObjC.import('stdlib')
+
+const app = Application.currentApplication()
+app.includeStandardAdditions = true
+
+const result = app.displayDialog('Please enter your computer password to allow Tidepool Uploader to access your device.', {
+  defaultAnswer: '',
+  withIcon: 'note',
```
pazaan, line 29: This doesn't appear to be a valid icon. ie, it will just use the default app icon.

lib/drivers/onetouch/oneTouchVerio.js
```diff
@@ -300,6 +304,13 @@ class BlockDevice {
         if (this.devicePath) {
           debug('devicePath:', this.devicePath);
 
+          if (os.platform() === 'darwin' && semver.compare(os.release(), '19.0.0') >= 0) {
+            // >= macOS Catalina
+            const cmd = `chmod a+rw ${this.devicePath}`.replace(/'/g, "'\\''");
```
pazaan, line 309:
```suggestion
            const cmd = `chmod a+rw "${this.devicePath}"`;
```

lib/drivers/onetouch/catalina-sudo/sudo-askpass.osascript.js
gniezen: According to the [official documentation](https://developer.apple.com/library/archive/documentation/LanguagesUtilities/Conceptual/MacAutomationScriptingGuide/DisplayDialogsandAlerts.html) there _should_ be `note`, `stop` and `caution` system icons. I just tested it again with all three, and the `note` icon indeed does just display the default app icon. If I don't add this line, it doesn't display an icon at all, so I'm going to leave it as is.

gniezen:
@krystophv Would you mind reviewing the changes I made on this PR since Lennart's last review?

gniezen:
@pazaan Just one last double-check now that I've removed the test suffix, please :smile:

PR #1166, opened 2020-05-04
Make sure we delete duplicate IBF files in temp folder (UPLOAD-306)

gniezen:
Addresses [UPLOAD-306] and extends the retry timeout for large files.

[UPLOAD-306]: https://tidepool.atlassian.net/browse/UPLOAD-306


gniezen:
@pazaan Another case of removing the test suffix. I guess this will happen with every PR from now on.

PR #1193, opened 2020-05-22
Add support for uploading CareSens meters via USB cable (UPLOAD-315)

gniezen:
Addresses: 

- [UPLOAD-315]

[UPLOAD-315]: https://tidepool.atlassian.net/browse/UPLOAD-315


lib/drivers/i-sens/careSens.js
```diff
@@ -0,0 +1,604 @@
...
+
+    getConfigInfo(progress, data, cb) {
+      progress(0);
+
+      (async () => {
+        cfg.deviceInfo.serialNumber = await driver.getSerialNumber();
+        cfg.deviceInfo.deviceId = `${cfg.deviceInfo.driverId}-${cfg.deviceInfo.serialNumber}`;
+        cfg.deviceInfo.deviceTime = sundial.formatDeviceTime(await driver.getDateTime());
+        debug('Config:', cfg);
+      })().then(() => {
```
pazaan, line 469: Do you need a `.then` here? You're already in an `async` block, so the code in this block could just be in the `async` block, no?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tidepool-uploader",
-  "version": "2.30.1",
+  "version": "2.30.0-i-sens-hid.1",
```
pazaan, line 3: Same as version query above.

app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
-  "version": "2.30.1",
+  "version": "2.30.0-i-sens-hid.1",
```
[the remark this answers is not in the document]
gniezen: Now that I've switched to GitHub flow, I'l be removing this version tag after Eric completes QA, and then merge it into `master` still as `2.30.1`. I guess I should've made it `.1` instead of .0` here, but it's the suffix that matters for QA. I will only rev the version number to `2.31.0` when I actually push a new version to prod.

lib/drivers/i-sens/careSens.js
```diff
@@ -0,0 +1,604 @@
...
+    return sundial.buildTimestamp(fields);
+  }
+
+  async getNumberOfRecords() {
+    const result = await this.commandResponse(COMMAND.CURRENT_INDEX);
+    return struct.extractBEShort(result.data, 0);
+  }
+
+  async getRecords(nrOfRecords) {
+    // TODO: read more than 27 records
```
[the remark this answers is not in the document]
gniezen: Oops, yes, I should take out - it's been completed.

package.json
gniezen: You know what - since I'm already making changes, I'll change this to `v2.30.1-i-send-hid.1` to make it clearer that I'm only adding a new suffix.

lib/drivers/i-sens/careSens.js
gniezen: Good point! I'll remove it.

gniezen:
Ticket has been approved, so test suffixes have been removed. Ready for final check.

lib/hidDevice.js
```diff
@@ -92,8 +93,10 @@ module.exports = function(config) {
       debug('just tried to send nothing!');
     } else {
       var arr = Array.from(buf);
-      arr.unshift(0); // The first byte of arr must contain the Report ID.
-                      // As we only work with a single report, this is set to 0x00.
+      if (config.deviceInfo.driverId !== 'CareSens') {
+        arr.unshift(0); // The first byte of arr must contain the Report ID.
```
pazaan, line 97: Why is this different for `CareSens`? Might be nice to have a comment explaining why...

gniezen: Ah, very good point! I'll add a comment.

PR #1214, opened 2020-06-11
[UPLOAD-278] react upgrade

krystophv:
To address [UPLOAD-278]

Upgrades `react`/`react-dom` and `react-redux`, and as dependent upgrades to those, `redux-form`, `react-router-dom` and replaces `react-router-redux` with `connected-react-router` since the former is now an abandoned project. 

[UPLOAD-278]: https://tidepool.atlassian.net/browse/UPLOAD-278


gniezen:
Merged in changes from master and removed test suffix after ticket was approved, so just a final check before I can merge this PR.

PR #1237, opened 2020-06-26
Renamed docs/StartHere.md to README.md for usability.

atombrella:
I find it slightly more intuitive that the overview is rendered by default under the files in Github.

PR #1244, opened 2020-07-03
Docker for linux development

TheDukeDK:
So starting out with TidePool and using Fedora 32 as my OS and it has been a bit of a pain. It seems there are some dependencies that I could not meet to get the uploader working properly on Fedora. My colleague however could get it up and running on his Ubuntu. 

So I had the idea to dockerize this based on Ubuntu 18.04 image which I could run on my Fedora laptop. 

This might be useful for developers who are using Linux but are not on an Ubuntu distribution and make it easier to get the uploader running locally.

### Strategy

**Dockerfile**

- Base is Ubuntu 18.04
- Installs all the needed dependencies like libusb-dev, etc.
- Installs a bunch of other needed libs for the GUI.
- Installs node v12.18.1
- Installs Yarn.

**docker-compose.yaml**
- Builds the image
- Mounts the code from the host inside the container, e.g "." 
- Mounts the .X11 socket inside the container so the GUI can use the hosts X11 server.
- Mounts /dev/bus/usb inside the container
- Uses hosts network

**.env**
- Used to set environment variables for docker-compose.

**entrypoint.sh**
- On container start runs "yarn install" followed by "yarn dev"

### Usage

1. Build the image
`docker-compose build`

2. Run it
`docker-compose up -d`

3. Other Yarn commands

    Even if you kill the TidePool Uploader GUI the container will continue to run. You can work with the yarn commands like you would locally by using docker exec. 

    **Examples**

    Interactively select the yarn target
    `docker exec -it uploader bash -c "yarn run"`

    Rebuild
    `docker exec -it uploader bash -c "yarn build"`

    Start the Dev GUI
    `docker exec -it uploader bash -c "yarn dev"`


TheDukeDK:
I have tested this on hosts running Fedora 32 and Ubuntu 18.04. I have not yet tested whether the mounting of the usb works. But hope to in the near future.

gniezen:
This looks great, thanks @TheDukeDK ! Just this past week we've had another request for a Docker image: https://github.com/tidepool-org/uploader/issues/1238 I've been wondering what exactly the use case is, as I've been using Ubuntu and Elementary OS (also a Ubuntu-based distro) to build Uploader, so I wasn't even aware that it's an issue on Fedora.

I'll make some time to do a proper code review, but two things came up at first glance:

- That's a massive list of dependencies that may have to be checked and updated in future. Maybe we could shorten it to the ones suggested in the issue linked above: `apt-get install -y git curl gcc g++ make libavutil-dev libsecret-1-dev libudev-dev build-essential libc6 libusb-1.0.0-dev libusb-1.0` (and I think that can be reduced even further, as `build-essential` should install `gcc`, `g++` etc.)
- Could you add a section to the README with the explanations on how to use it?

atombrella:
@gniezen The container request came from both Tim (@TheDukeDK) and me, as we work together on this. My local host is Ubuntu 18.04 (that's the other test case referred in the description of this PR. I did also manage to get it running easily on my host as well, but having it isolated was a motivation for proposing it as a container).

TheDukeDK:
@gniezen - Yes, I know the dependency list is massive. Most of them are so the UI works for the X11 forwarding. I am pretty sure just installing the dependencies needed for build will not work. To be honest I am not sure **all** of them are needed. I just started adding dependencies like crazy :_P

I will take some time and start trimming the list to the bare minimum. Maybe I can find time this weekend.

TheDukeDK:
On a side note. If someone with electron experience could get the dynamic libs packaged as **part** of the electron-build. Then it should be possible to use this container to generate AppImage's I think. 

For example, libavutil.so, does not seem to be part of the end result of "yarn package-linux".

gniezen:
> Most of them are so the UI works for the X11 forwarding

Ah ok, that makes sense. I just got the container up and running, and the X11 forwarding is pretty cool! I was able to successfully upload a Tandem pump, but there's one caveat: The device needs to be plugged in _before_ the container is started. There may be some solutions here: https://stackoverflow.com/questions/24225647/docker-a-way-to-give-access-to-a-host-usb-or-serial-device One other thing that we may need is the udev rules (for the host OS) for all the supported devices. I have a list that I can share.

>  libavutil.so, does not seem to be part of the end result of "yarn package-linux".

We haven't done any work to package the Uploader on Linux yet. This is mainly due to the increased QA burden: When we add a new device, we're currently testing it on the latest two releases of MacOS, as well as Windows 7, 8/8.1 and 10 (both 32-bit and 64-bit versions), for a total of 8 environments. Now that Windows 7 is sunset, we may have some extra bandwidth to add Linux as a supported OS. Now the question becomes: Which and how many distros do we have to test?

gniezen:
It appears the easiest way to have it working while unplugging/re-plugging devices is to mount `/dev:/dev` instead, but that's not ideal. The alternative is to [use cgroups](https://docs.docker.com/engine/reference/commandline/create/#dealing-with-dynamically-created-devices---device-cgroup-rule), but that seems complicated.

TheDukeDK:
Yes, I think the cgroups would add some unneeded complexity. Do you think /dev:/dev with udev rules be sufficient?

TheDukeDK:
Maybe some guidance which is useful here. https://www.google.com/amp/s/www.losant.com/blog/how-to-access-serial-devices-in-docker%3fhs_amp=true

TheDukeDK:
I have trimmed the dependency list down quite a bit. I have also added a section to the readme. Next step I guess is the udev rules.

TheDukeDK:
Finally we have been trying to get a successful upload of a Medtronic 600 device to work with local Tidepool cluster. But have had zero success :-( 

Not with the container NOR with my colleagues Ubuntu laptop. 

Is there anything we should know when using the uploader on a Linux environment against a local Tidepool cluster?

atombrella:
@gniezen You mentioned that you can upload with the docker image. To which environment are you uploading? We're trying with a local Tidepool cluster installed, running using the Tilt tool (it's the first time either of us has come across it, and it seems quite neat).

gniezen:
I just successfully uploaded a 670G using the Docker image to a local environment (that I set up using https://github.com/tidepool-org/development/tree/k8s-stable). What error are you getting? Could it be related to udev permissions?

I just posted the udev rules for all supported devices as a gist here: https://gist.github.com/gniezen/2a40b7c191a27de94962a49c46a602df

They should go into `/etc/udev/rules.d/`, and then reboot (or follow https://unix.stackexchange.com/questions/39370/how-to-reload-udev-rules-without-reboot and replug device)

TheDukeDK:
@gniezen - Works with udev rules added to host machine. Is there anything else I should do as far as this PR is concerned? 

And THANK YOU for the help!

gniezen:
Excellent, glad to hear it worked! I'll  see if I can do a code review today.

.env
```diff
@@ -0,0 +1,9 @@
+#ROLLBAR_POST_TOKEN=
+
+API_URL=http://localhost:3000
+UPLOAD_URL=http://localhost:3000
+DATA_URL=http://localhost:3000
+BLIP_URL=http://localhost:3000
+DEBUG_ERROR=false
```
gniezen, line 7: For developers, it may be useful to have additional logging and debug options enabled by default?

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM ubuntu:18.04
+
+ENV DEBIAN_FRONTEND noninteractive
+ENV NODE_VERSION "v12.18.1"
```
gniezen, line 4: The Node version inside Electron is v12.13.0. It helps if the versions match, otherwise the native dependencies need to be recompiled.

README.md
```diff
@@ -39,6 +39,31 @@ $ yarn dev
 
 **NB:** React components and CSS will hot load after changes (this can be confirmed by watching the JavaScript console), but changes to device drivers and other code outside of the React components will not - use 'Reload' to reload after such changes. If the compilation/hot reload of a component fails for any reason (e.g. from a syntax error) you may need to reinitialize the hot loader by reloading the application.
 
+### Docker for Linux
+
+If you are running Linux you probably need to be using an Ubuntu distribution or derivative. To get around this for other distrubutions you can try to build a local docker image which is based on Ubuntu 18.04 and use the yarn/npm commands interactively.
+
+**NOTE:** You need to add udev rules to your host for uploads to actually work. You can find the udev rules in this [gist](https://gist.github.com/gniezen/2a40b7c191a27de94962a49c46a602df). The file should be placed in `/etc/udev/rules.d/` and the host should be rebooted.
```
gniezen, line 46: I'm wondering if we shouldn't maybe put the udev rules in the uploader repo itself instead of a gist that may disappear in future. Maybe `resources/linux`?

README.md
```diff
@@ -39,6 +39,31 @@ $ yarn dev
...
+
+1. Build the image
+    `docker-compose build` 
+    
+2. Run it
+    `docker-compose up -d` 
+
+3. Work with it interactively. 
+
+    Even if you kill the TidePool Uploader GUI the container will continue to run. You can work with the yarn commands like you would locally by using docker exec.
```
gniezen, line 57:
```suggestion
    Even if you kill the Tidepool Uploader GUI the container will continue to run. You can work with the yarn commands like you would locally by using docker exec.
```

.env
TheDukeDK: Agreed.

TheDukeDK: Done.

README.md
TheDukeDK: Agreed and done.

Dockerfile
TheDukeDK: Agreed and done.

gniezen:
Oh, one more thing: @TheDukeDK, would you mind agreeing to the Tidepool Contributor License Agreement like @atombrella did in this PR: https://github.com/tidepool-org/uploader/pull/1237 ?

TheDukeDK:
“I agree to the terms of Tidepool Project’s Volunteer/Contributor License Agreement
v1.1 as it exists at http://tidepool-org.github.io/files/TidepoolVCLA-1.1.pdf on 08-07-2020.”

TheDukeDK:
@gniezen - 

> Oh, one more thing: @TheDukeDK, would you mind agreeing to the Tidepool Contributor License Agreement like @atombrella did in this PR: #1237 ?

**Done**

gniezen:
Awesome! Great work, thanks guys!

PR #1250, opened 2020-07-09
Also publish new builds to S3 (UPLOAD-376)

gniezen:
Addresses [UPLOAD-376].

[UPLOAD-376]: https://tidepool.atlassian.net/browse/UPLOAD-376


gniezen:
Ticket approved, removed test suffix for final review before merging.

PR #1277, opened 2020-08-11
Add device weitai equil

pauloved:
add new device
name: Equil Insulin Patch Pump/Equil Insulin Micro Pump
deviceId:MTM-I
intro:It is small and lightweight also the tubeless micro insulin pump has 4 years warranty which could provide 72h insulin supply.
device-pic:
image


pauloved:
hi @gniezen  , Is this the correct PR ?

PR #1306, opened 2020-08-27
Dexcom user events, e.g. food, reported state, physical activity (UPLOAD-11)

gniezen:
Addresses [UPLOAD-11], and also ES6-ifies the Dexcom driver.

[UPLOAD-11]: https://tidepool.atlassian.net/browse/UPLOAD-11


lib/drivers/dexcom/dexcomDriver.js
```diff
@@ -142,70 +168,73 @@ module.exports = function (config) {
...
     'Vibrate',
     'Soft',
     'Normal',
     'Attentive',
-    'Hyposafe'
+    'Hyposafe',
   ];
 
-  var hasFlag = function (flag, v) {
+  const hasFlag = (flag, v) => {
```
pazaan, line 182: General style query: why are these named arrow functions, as opposed to regular functions?

lib/drivers/dexcom/dexcomDriver.js
```diff
@@ -509,266 +509,288 @@ module.exports = function (config) {
...
 
-    var parser = null;
-    if (rectype == RECORD_TYPES.EGV_DATA || rectype == RECORD_TYPES.BACKFILLED_EGV_DATA) {
-      parser = makeHeaderParser(parse_egv_records);
-    } else if (rectype == RECORD_TYPES.METER_DATA) {
-      parser = makeHeaderParser(parse_meter_records);
-    } else if (rectype == RECORD_TYPES.USER_SETTING_DATA) {
-      parser = makeHeaderParser(parse_setting_records);
+    let parser = null;
+    if (rectype === RECORD_TYPES.EGV_DATA || rectype === RECORD_TYPES.BACKFILLED_EGV_DATA) {
```
pazaan, line 552: Would a `switch` be more readable here?

lib/drivers/dexcom/dexcomDriver.js
```diff
@@ -1272,33 +1384,33 @@ module.exports = function (config) {
...
+            debug(err);
+            progress(100);
+            return cb(err, data);
+          }
           progress(100);
           return cb(null, data);
-        }
-      });
+        }, 'dataservices',
+      ); // TODO: REMOVE THIS! (switching to platform now would cause dupes)
```
pazaan, line 1396: Should this TODO be addressed?

pazaan (commented):
Main query is about whether the `TODO` needs to be actioned. 2 smaller general queries.

lib/drivers/dexcom/dexcomDriver.js
gniezen: eslint flagged these as they're unnamed functions: https://eslint.org/docs/rules/func-names. I've been changing everything to arrow functions because I think it looks cleaner and it matches the rest of the driver styles.

lib/drivers/dexcom/dexcomDriver.js
gniezen: Oh yes, good catch! After changing this back, I just discovered that Jellyfish has a validation check that leads to the following error:

`food.carbs should be a number, got[undefined]`

This should be changed, as the updated `food` type on platform uses a different data structure. We never used the old `food` structure that Jellyfish is trying to validate. I just created https://tidepool.atlassian.net/browse/BACK-1692 to track this.

Related question: If the Dexcom API uses platform instead of Jellyfish to upload `food` data, what happens with deduplication when someone uploads their receiver through Jellyfish?

lib/drivers/dexcom/dexcomDriver.js
gniezen: Yes! Will update.

lib/drivers/dexcom/dexcomDriver.js
pazaan: It won't be deduplicated. Same issue we have with multiple sources anywhere in Tidepool.

pazaan (dismissed):
LGTM 🚀

cbwebdevelopment (approved):
LGTM 🚢

PR #1360, opened 2021-01-26
Integrate spectron

ginnyyadav:
Simple spectron integration with a very simple test example so we can build off of it.


.circleci/config.yml
```diff
@@ -36,6 +36,7 @@ jobs:
     # Test
     - run: yarn lint
     - run: yarn test
+    - run: yarn run spectron-test
```
gniezen, line 39:
```suggestion
    - run: yarn spectron-test
```

appveyor.yml
```diff
@@ -35,6 +35,7 @@ test_script:
   - node --version
   - yarn lint
   - yarn test
+  - yarn run spectron-test
```
gniezen, line 38:
```suggestion
  - yarn spectron-test
```

package.json
```diff
@@ -31,7 +31,8 @@
     "package-win": "npm run build && electron-builder --win --x64",
     "package-mac": "npm run build && electron-builder --mac",
     "package-linux": "npm run build && electron-builder --linux",
-    "package-all": "npm run build && electron-builder -mwl"
+    "package-all": "npm run build && electron-builder -mwl",
+    "spectron-test": "yarn build && mocha -r esm"
```
gniezen, line 35: We've migrated from mocha to jest a while ago. I'm just wondering whether the spectron tests will work with jest, or is it only possible with mocha?

test/login.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  before(async () => {
+    app = await startApp();
+  });
+
+
+  after(async () => {
+    await stopApp(app);
+  });
+
+  it('should open', async () => {
```
gniezen, line 17: Is this actually `async`? I don't see an `await`.

test/login.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    };
+});
+
+  it('should have all UI elements', async () => {
+    await LoginScreen.signUpLink.should.eventually.exist;
+    await LoginScreen.tidepoolLogo.should.eventually.exist;
+    await LoginScreen.uploaderLogo.getText()
+      .should.eventually.equal('Uploader');
+    await LoginScreen.forgotPasswordLink.should.eventually.exist;
+    await LoginScreen.supportLink.getText()
```
gniezen, line 34: I remember we had an issue for a while where the support link wasn't working. Can we also add tests here to check that the URLs don't 404?

test/utilities.js
```diff
@@ -0,0 +1,29 @@
...
+import { should, use } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import electron from 'electron';
+import path from 'path';
+import LoginScreen from './src/LoginScreen';
+import Base from './src/Base';
+
+global.before(() => {
+  should();
+  use(chaiAsPromised);
```
gniezen, line 11: Would you mind adding a comment here to explain why `chaiAsPromised` is necessary  here?

gniezen (changes requested):
This is awesome! I've added a couple of suggestions and comments, but nothing major. Code quality is :100:

package.json
ginnyyadav: whoops, didn't realize that, totally will work with jest, will make it so!

test/utilities.js
ginnyyadav: this won't be here if we use jest i think.

ginnyyadav:
Went ahead and switched to jest...a couple things here..

- i made stricter checking on the URLs, but i can't go any further than that since they open in a chrome browser and not in the electron app. the positive side of this is that with the web nightwatch tests we can make sure these specific urls are not returning 404s.

- I had to put in an explicit wait since the login button on uploader on the windows machine is loading enough for the automation to click on it and render it as visible, but not loading completely enough to actually complete the function of logging in

- I've put in a remote session for debugging. the password is in the 1 password vault. I did this for debugging but not sure if you want me to leave it in there in case anyone else finds this helpful. If you'd like me to remove it I can

README.md
```diff
@@ -98,6 +98,15 @@ or
 ```bash
 $ yarn test
 ```
+To run the spectron UI tests in this repository use:
+
+```bash
+$ yarn run spectron-test
```
gniezen, line 104: `run` is optional for yarn, so `yarn spectron-test` is sufficient. These could also be changed in the appveyor and circleci config files.

package.json
```diff
@@ -32,7 +32,7 @@
     "package-mac": "npm run build && electron-builder --mac",
     "package-linux": "npm run build && electron-builder --linux",
     "package-all": "npm run build && electron-builder -mwl",
-    "spectron-test": "yarn build && mocha -r esm"
+    "spectron-test": "yarn run build && jest --config test/spectron/jest.config.js test/spectron/test"
```
gniezen, line 35: You can remove `run` here as well.

README.md
```diff
@@ -153,6 +162,8 @@ We use the following environment variables on the CI servers:
 | PUBLISH_FOR_PULL_REQUEST | Both     | `true`, build artefact for PR |
 | ROLLBAR_POST_TOKEN       | Both     | Rollbar logging |
 | FTP_AV_PASSWORD_TIDEPOOL | Appveyor | AV submission |
+| SPECTRON_USER_EMAIL      | Both     | Valid Tidepool username for Spectron tests |
+| SPECTRON_USER_PASSWORD   | Both     | Valid Tidepool password for SPectron tests |
```
gniezen, line 166:
```suggestion
| SPECTRON_USER_PASSWORD   | Both     | Valid Tidepool password for Spectron tests |
```

gniezen (changes requested):
Looks good, great job @ginnyyadav ! I just have two comments:

- `run` is not necessary when using `yarn`
- Please remove the `app/package-lock.json` file, as that as created by `npm`.

gniezen (approved):
LGTM! :shipit:

Issue #1364, opened 2021-02-10
App error in Docker - lack of library libXss.so.1


MariannaKrzewinska:
Hey,
When I tried to run the app in Docker using every app building command, I got an error:
```bash
[~/d/tidepool-uploader] : docker-compose up                                             (master) 22:43:51
WARNING: The ROLLBAR_POST_TOKEN variable is not set. Defaulting to a blank string.
Recreating uploader ... done
Attaching to uploader
uploader    | yarn install v1.22.5
uploader    | [1/4] Resolving packages...
uploader    | warning Resolution field "charm@1.0.2" is incompatible with requested version "charm@0.1.x"
uploader    | success Already up-to-date.
uploader    | $ concurrently "electron-builder install-app-deps" "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json"
uploader    | [0]   • electron-builder  version=22.9.1
uploader    | [1] node node_modules/fbjs-scripts/node/check-dev-engines.js package.json exited with code 0
uploader    | [0]   • installing production dependencies  platform=linux arch=x64 appDir=/home/node/uploader/app
uploader    | [0] electron-builder install-app-deps exited with code 0
uploader    | Done in 1.69s.
uploader    | yarn run v1.22.5
uploader    | $ cross-env START_HOT=1 yarn start-renderer-dev
uploader    | $ cross-env NODE_ENV=development webpack-dev-server --config webpack.config.renderer.dev.babel.js
uploader    | ~ ~ ~ ~ ~ ~ ~ ~ ~ ~
uploader    | ### DEBUG MODE ###
uploader    | ~ ~ ~ ~ ~ ~ ~ ~ ~ ~
uploader    | 
uploader    | ***** NOT using the default environment *****
uploader    | The default right-click server menu may be incorrect.
uploader    | API_URL = http://localhost:3000
uploader    | UPLOAD_URL = http://localhost:3000
uploader    | DATA_URL = http://localhost:3000
uploader    | BLIP_URL = http://localhost:3000
uploader    | Starting Main Process...
uploader    | 
uploader    | > tidepool-uploader@2.34.0 start-main-dev /home/node/uploader
uploader    | > cross-env HOT=1 NODE_ENV=development electron -r @babel/register ./app/main.dev.js
uploader    | 
uploader    | /home/node/uploader/node_modules/electron/dist/electron: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory
uploader    | npm ERR! file sh
[...]

PR #1365, opened 2021-02-10
#1364 Docker install libxss1 package

MariannaKrzewinska:
PR to the issue #1364


gniezen (approved):
I was able to reproduce the issue and the new dependency is clearly required by the Electron Linux build instructions. LGTM! :+1:

I'm going to merge this change without any QA testing as it doesn't have any user-facing changes.

PR #1366, opened 2021-02-11
Revert "Upload DASH PDMs using WebMTP (UPLOAD-349)" (UPLOAD-527)

gniezen:
This reverts tidepool-org/uploader#1284, which is blocking new Uploader releases, as the changes are currently on hold. See [UPLOAD-349] for details.

[UPLOAD-349]: https://tidepool.atlassian.net/browse/UPLOAD-349


gniezen:
@tjotala I'm reverting the DASH MTP changes from the master branch for now.

lib/drivers/insulet/insuletDriver.js
```diff
@@ -1757,22 +1788,66 @@ module.exports = (config) => {
...
+          ps.addCommand('Write-Host $item.Name');
+
+          progress(50);
+          debug('Invoking PowerShell..');
+          const output = await ps.invoke();
+          debug('Copied', output);
+          progress(80);
+          ps.dispose();
+
+          if (output === '') {
```
tjotala, line 1837: Could this check result in false positive if the PS script fails in any way? (e.g. prints out error message)

tjotala (approved):
LGTM, aside from one non-blocking curiosity question

lib/drivers/insulet/insuletDriver.js
gniezen: Good question. For errors, `ps.invoke()` throws an error that is caught by the `try/catch` block, and I'm not aware of any error messages being printed to `output` when the script fails. This code is what is currently in production, so unless I have reason to believe that this is causing an issue, I'd rather not change it. I mean, if all goes well we can get rid of this code again in a month or two, right?

tjotala: 👍🏻

gniezen:
@tjotala Just need re-approval after removing the test suffix from the version number.

cbwebdevelopment (approved):
LGTM 👍

PR #1367, opened 2021-02-14
#1364 docker-compose.override.yaml

MariannaKrzewinska:
Hello @gniezen, sorry for the delay - here's PR with instructions in README how to run the app by passing command in `docker-compose.override.yaml` (added also to `.gitignore`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ node_modules
 app/node_modules
 bower_components
 coverage
+docker-compose.override.yaml
 
 .DS_Store
 .idea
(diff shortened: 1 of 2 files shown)


gniezen:
Hi @MariannaKrzewinska ! I just want to clarify what the intent of this override file is, i.e., why not just use the existing `docker-compose up` command? Is it because you don't want to run `yarn install` every time? If so, you may need to change `command` to `entrypoint`, as to override the existing entrypoint defined in `entrypoint.sh`.

MariannaKrzewinska:
Thanks for pointing this out, I was following the commands from README for running the app in Docker (running docker-compose in detached mode etc) and didn't see that just running `docker-compose up` will do the trick. 
Other use case for docker-compose.override is to not commit by mistake changes in execution commands, but it doesn't seem that relevant, so I'll close the PR

PR #1369, opened 2021-03-08
Add support for ReliOn Premier meters (UPLOAD-373)

gniezen:
Addresses [UPLOAD-373] by adding support for:

- ReliOn Premier BLU
- ReliOn Premier Classic
- ReliOn Premier Voice

[UPLOAD-373]: https://tidepool.atlassian.net/browse/UPLOAD-373

### DIFF
--- a/lib/drivers/i-sens/careSens.js
+++ b/lib/drivers/i-sens/careSens.js
@@ -43,6 +43,11 @@ const ASCII_CONTROL = {
   ETX: 0x03,
 };
 
+const UART = {
+  ECHO: 0x80,
+  READ_FLASH: 0x8B,
+};
+
 const REPORT_ID = {
   GET_SET_UART_ENABLE: 0x41,
   GET_VERSION_INFO: 0x46,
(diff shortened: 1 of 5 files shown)


lib/drivers/i-sens/careSens.js
```diff
@@ -172,13 +327,142 @@ class CareSens {
...
+    const packet = {
+      bytes,
+      valid: false,
+      packet_len: 0,
+      payload: null,
+    };
+
+    const packetLength = bytes.length;
+    packet.packet_len = packetLength;
+    packet.valid = true;
```
krystophv, line 348: Not strictly necessary to change, but couldn't these assignments be done in the initial definition at 339?

lib/drivers/i-sens/careSens.js
```diff
@@ -172,13 +327,142 @@ class CareSens {
...
+
+    const packet = CareSens.extractPacket(buffer.bytes());
+    if (packet.packet_len !== 0) {
+      // cleanup the buffer data
+      buffer.discard(packet.packet_len);
+    }
+
+    if (packet.valid) {
+      return packet;
+    }
```
krystophv, line 366: It seems that `CareSens.extractPacket()` will _always_ return a packet with the `.valid` set to `true`. If that's the case, then we expect this logic branch to _always_ be run, correct?

krystophv (dismissed):
Couple comments/suggestions, but nothing that precludes this from approval. LGTM 👍

lib/drivers/i-sens/careSens.js
gniezen: Yes! I guess that's what happens if you copy & paste a function from another driver and then end up changing it so much that it's almost not necessary anymore.

lib/drivers/i-sens/careSens.js
gniezen: :+1:

gniezen:
@krystophv Can you just check if I addressed your comments correctly?

krystophv (approved):
LGTM 👍

gniezen:
@krystophv Sorry, just one last approval needed so that I can merge to master.

krystophv (approved):
LGTM 👍🏼

PR #1384, opened 2021-05-18
Hotfix v2.36.1 (UPLOAD-577)

gniezen:
Addresses [UPLOAD-577].

[UPLOAD-577]: https://tidepool.atlassian.net/browse/UPLOAD-577

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
-  "version": "2.36.0",
+  "version": "2.36.1",
   "description": "Tidepool Project Universal Uploader",
   "main": "./main.prod.js",
   "author": {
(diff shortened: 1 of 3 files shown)


app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
-  "version": "2.35.0",
+  "version": "2.36.1",
```
cbwebdevelopment, line 4: Shouldn't a hotfix only bump the patch version, and not the minor?

cbwebdevelopment (approved):
LGTM, so long as the minor version upgrade was intentional.  If not, I suggest only bumping the patch version.

app/package.json
gniezen: The minor version (https://github.com/tidepool-org/uploader/pull/1380) hasn't been merged yet, so this PR can supersede that one, unless you think we should merge that one first?

cbwebdevelopment: Nope, that's fine, and makes sense.

PR #1397, opened 2021-07-21
Update .bookignore file

gniezen:
We haven't published the updated docs to https://developer.tidepool.org/uploader in a long time, so when I just did this I had to update the .bookignore file for Gitbook.


gniezen:
Going to merge this, as it's not user-facing.

PR #1402, opened 2021-08-03
Upload binaries to Kaspersky using secure FTP

gniezen:
We received a notification from Kaspersky that they will only allow uploads via secure FTP going forward. As this is not a user-facing change and doesn't affect the Uploader code, there's no associated Jira ticket.


krystophv (approved):
LGTM 🚀

PR #1408, opened 2021-08-23
Remove unused packaging lines

gniezen:
Now that we package Uploader with a separate config file, we either need to (1) update the other packaging commands or (2) remove them. I opted for option 2.

Since this is not a user-facing change, there's no associated Jira ticket.


gniezen:
@krystophv Would you mind having a look at this PR - as it's not user-facing it doesn't have an associated Jira ticket?

krystophv (approved):
LGTM 👍🏼

PR #1419, opened 2021-09-14
Roll back serialport to 9.0.7

gniezen:
We updated to serialport 9.2.0 in https://github.com/tidepool-org/uploader/pull/1411, but any @serialport/bindings newer than 9.0.7 is broken on Linux, see https://github.com/serialport/node-serialport/issues/2266 for details. Until the right prebuilds are available, we'll have to use `resolutions` to keep @serialport/bindings to a working version.


gniezen:
Going to merge this without review, as it's a roll back.

PR #1431, opened 2021-10-17
Add a driver for uploading CSV downloaded from Abbott LibreView (UPLOAD-586)

kpouget:
This PR adds a driver for uploading CSV files downloaded from Abbott LibreView.

The data from the Freestyle Libre 2 can be uploaded to LibreView (using Abbott proprietary drivers), then LibreView interface allows exporting the data to a CSV file. This PR allows the uploading of this file to Tidepool.

Here is an example of data:
```
Glucose Data,Generated on,18-09-2021 16:39 UTC,Generated by,Kevin Pouget
Device,Serial Number,Device Timestamp,Record Type,Historic Glucose mg/dL,Scan Glucose mg/dL,Non-numeric Rapid-Acting Insulin,Rapid-Acting Insulin (units),Non-numeric Food,Carbohydrates (grams),Carbohydrates (servings),Non-numeric Long-Acting Insulin,Long-Acting Insulin Value (units),Notes,Strip Glucose mg/dL,Ketone mmol/L,Meal Insulin (units),Correction Insulin (units),User Change Insulin (units)
FreeStyle Libre 2,<SN>,04-09-2021 12:40,0,76,,,,,,,,,,,,,,
FreeStyle Libre 2,<SN>,15-07-2021 10:13,1,,142,,,,,,,,,,,,,
FreeStyle Libre 2,<SN>,16-07-2021 00:11,1,,150,,,,,,,,,,,,,
FreeStyle Libre 2,<SN>,14-07-2021 13:33,6,,,,,,,,,,,,,,,
FreeStyle Libre 2,<SN>,14-07-2021 20:14,6,,,,,,,,,,,,,,,
FreeStyle Libre 2,<SN>,03-09-2021 19:28,2,,,,,,,,,,,97,,,,
FreeStyle Libre 2,<SN>,14-09-2021 08:31,3,,,,,,,,,,,,0.1,,,
```
with the column `Record Type` giving the type of the line:
```
    var typesToRead = {
        0: "Historic Glucose mg/dL",
        1: "Scan Glucose mg/dL",
        2: "Strip Glucose mg/dL",
        3: "Ketone mmol/L",

        4: false, // Rapid-Acting Insulin (units) OR Non-numeric Long-Acting Insulin
        5: false, // Non-numeric Food
        6: false, // ?
    }
```

See also discussion in https://github.com/tidepool-org/uploader/issues/1388.


kpouget:
@gniezen I cannot get the GUI to work:
image
and I'm not sure how to investigate that, as the console don't show any error ...

kpouget:
Rebased and removed WIP files (`run ...`) and flags

gniezen:
@kpouget Does it happen with the mmol/L CSV file I linked above as well? If not, would you mind sending me the CSV file that it happens with to gerrit@tidepool.org?

Does the CLI mode still work for you?

According to the stack trace in the screenshot's error details it's happening in the `fetchData` section, and I think it's happening where we're converting the file data to a string: https://github.com/tidepool-org/uploader/pull/1431/files#diff-e3cb597eb71a42eb266e2cd6ca82468c86dc9d745c9fabfd59a7b48085b26c95R115

gniezen:
OK, I've switched from using `String.fromCharCode.apply()` (which was probably falling over due to the size of the array) to `TextDecoder()`, which is properly designed for the purpose. Please try again! :smile:

kpouget:
> Rebased and removed WIP files (run ...) and flags

I think you reintroduced the WIP that I removed with the previous push :) hopefully this version is up to date with master, and didn't lost anything during the rebase

(you'll need to `git reset --hard  39724f69` to sync the branch)

> OK, I've switched from using String.fromCharCode.apply() (which was probably falling over due to the size of the array) to TextDecoder(), which is properly designed for the purpose. Please try again!

yes, working better now :D

my CSV file has `19807` lines, but `maximum call stack size exceeded` really means recursion overflow, so was `String.fromCharCode.apply()` using a recursive loop to parse the string? .... I don't want to know 🙃

gniezen:
> (you'll need to `git reset --hard 39724f69` to sync the branch)

Will do. Please don't use force push going forward - we use a merge-before-push strategy with Uploader. 

> yes, working better now :D
> 
> my CSV file has `19807` lines, but `maximum call stack size exceeded` really means recursion overflow, so was `String.fromCharCode.apply()` using a recursive loop to parse the string? .... I don't want to know upside_down_face

Excellent! I think the recursion overflow happens because the browser can't handle such a large number of function arguments (`.apply()` converts the array to arguments). `TextDecoder` is the proper way to convert between arrays and strings.

gniezen:
@kpouget Is there anything else that you wanted to do on this PR, or can I start a full code review?

kpouget:
no, I think it's good enough for a first version!

lib/drivers/abbott/libreViewDriver.js
```diff
@@ -0,0 +1,355 @@
...
+      let entryCount = 0;
+      Object.keys(payload.devices).forEach((key) => {
+        const device = payload.devices[key];
+        debug(`Device ${key}: ${device.data.length} entries`);
+        entryCount += device.data.length;
+      });
+
+      if (entryCount === 0) {
+        debug('Error reading file, no data parsed.');
+        return null;
```
gniezen, line 216:
```suggestion
        return cb(new Error('Error reading file, no data parsed.'));
```

gniezen (changes requested):
I've only got one suggestion regarding returning an error message, but I think we could pass this off to QA soon!

lib/drivers/abbott/libreViewDriver.js
```diff
@@ -0,0 +1,356 @@
...
+
+      async.eachSeries(Object.keys(payload.devices), (key, done) => {
+        const device = payload.devices[key];
+        const deviceRecords = device.postRecords;
+
+        const sessionInfo = {
+          deviceTags: ['bgm', 'cgm'],
+          deviceManufacturers: ['Abbott'],
+
+          deviceModel: key,
```
gniezen, line 307: This should just contain the device model, not the serial number.

lib/drivers/abbott/libreViewDriver.js
kpouget: sure

lib/drivers/abbott/libreViewDriver.js
```diff
@@ -0,0 +1,356 @@
...
+          deviceSerialNumber: device.info.serialNumber,
+          deviceId: device.info.deviceId,
```
gniezen, line 309: Since we use the device ID for deduplication, it's probably not a good idea to have spaces in there (see https://developer.tidepool.org/data-model/device-data/common.html#deviceid for details).

Maybe just strip out the spaces, add the manufacturer and add a `-` between the name and serial number? Maybe a good idea to match the Libre driver, for example:

```js
deviceModel: "FreeStyle Libre"
deviceSerialNumber: "JCMW071-T1802"
deviceId: "AbbottFreeStyleLibre-JCMW071-T1802"
```

kpouget:
do you have all the CSV files you need for QA?

gniezen (changes requested):
I was just looking through the uploaded data and found a couple of more issues.

gniezen:
> do you have all the CSV files you need for QA?

Yeah, our QA team will be generating their own CSV files for testing, as they'll need to make sure they have all the required data (e.g. HI/LO, control solution tests, ketones etc.)

lib/drivers/abbott/libreViewDriver.js
kpouget: from the CSV I have:
```
FreeStyle Libre 2,MBGAxxx-Gxxx,18-09-2021 16:33,6,,,,,,,,,,,,,,,
FreeStyle LibreLink,xxx-xxx-xxx-xxx-xxx,15-07-2021 02:14,0,139,,,,,,,,,,,,,,
FreeStyle Libre,JCMW071-T1802,23-07-2019 11:05,6,,,,,,,,,,,,,,,
```
so something like that:
```
deviceModel = FIELD_1
deviceSerialNumber = FIELD_2
deviceId = ("abott" + deviceModel + deviceSerialNumber).replaceAll(" ", "")
```

gniezen: Yeah, that looks good to me. I'd just add a `-` between device model and serial number so that it matches the existing Libre driver. Also, use `'` instead of `"` to keep the linter happy.

kpouget: updated following your suggestion

kpouget: > I'd just add a - between device model and serial number

> AbbottFreeStyleLibre-JCMW071-T1802

I took too long to notice that the first `-` was to be 'hardcoded' while the second one was part of the S/N 🙃

gniezen: No worries, thank you! I've fixed the last couple of linting issues. I'm going to send this off to our QA team!

gniezen (approved):
LGTM! :shipit:

kpouget:
\o/

I was actually looking at the linter errors, I didn't notice them before :) (maybe it wasn't allowed to run)
image

gniezen:
@kpouget This is now live in v2.41.0. Thanks for all your help!

kpouget:
great @gniezen, thanks a lot for your help, I'm very proud about this first modest contribution :), and looking forward to find new topics to work on! I'll poke on slack to see if there's something around Kubernetes, to link it with my work topic

PR #1432, opened 2021-10-19
Add support for Roche Accu-Chek Smart Pix device

mrinnetmaki:
This is an initial pull request, to start discussion on supporting the Smart Pix reader device.

This version supports only model 1 / gray SmartPix readers, reading only insulin pumps (not glucometers yet).
The reader must be attached and mounted (it shows as mass storage device) before reading is started.

The XML part expects the document to be read with xml2js.

The pump data parser understands only name of Spirit/Combo pump, but assuming that it is the reader that generates the report format, it should handle other pumps too.

The parser understands data generated by Smart Pix 1 firmwares 3.01.03
and 3.05.01, but others should work too.

Time zone and time changes are not yet handled properly.

Props mainly to @jlaunonen for the work this far.

### DIFF
--- a/lib/drivers/roche/accuChekSmartPix.js
+++ b/lib/drivers/roche/accuChekSmartPix.js
@@ -9,8 +9,8 @@ import annotate from '../../eventAnnotations';
 import common from '../../commonFunctions';
 import TZOUtil from '../../TimezoneOffsetUtil';
 import { parseDtTm } from './smartpix/accuChekSmartPixCommonXML';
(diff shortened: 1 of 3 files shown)


gniezen:
@mrinnetmaki Will you be able to agree to the VCLA on https://developer.tidepool.org/contributors/ ?

mrinnetmaki:
> @mrinnetmaki Will you be able to agree to the VCLA on https://developer.tidepool.org/contributors/ ?

Hi! Only saw this now, apologies. VCLA sent to legal.

PR #1441, opened 2021-11-08
[UPLOAD-693] classic clinic interface

krystophv:
For [UPLOAD-693], re-tools when/how we display the "private workspace" for the new clinicians, bringing back the clinic interface for the private workspace instead of directly pushing the user to upload for themselves.

Also takes care of a few instances of places that needed to be updated to reflect Uploader's update to match Blip's `allUsers` state (where Uploader previously just had the user's profile, now we have the whole user object to mirror Blip). 

Again, this targets the `workspaces` branch since we don't want to merge into `master` until the clinics updates get tested as a while.

[UPLOAD-693]: https://tidepool.atlassian.net/browse/UPLOAD-693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/actions/async.js
+++ b/app/actions/async.js
@@ -539,7 +539,7 @@ export function setTargetTimezone(userId, timezoneName) {
         if (_.get(err,'status') !== 401) {
           dispatch(sync.updateProfileFailure(err));
         } else {
-          let newProfile = actionUtils.mergeProfileUpdates(allUsers[userId], updates);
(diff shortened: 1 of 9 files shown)


clintonium-119 (approved):
Great work :shipit:

PR #1466, opened 2022-01-13
[UPLOAD-734] existing email error

krystophv:
To handle displaying an error to the user for [UPLOAD-743]

[UPLOAD-743]: https://tidepool.atlassian.net/browse/UPLOAD-743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/ClinicUserEdit.js
```diff
@@ -200,6 +206,27 @@ class ClinicUserEdit extends React.Component {
...
+  renderNotification = () => {
+    var {creatingClinicCustodialAccount} = this.props.working;
+    if (_.isNull(creatingClinicCustodialAccount.notification)) {
+      return null;
+    }
+
+    return (
+      <div className={styles.error}>
+        <span>
+          {i18n.t(creatingClinicCustodialAccount.notification.message)}
```
clintonium-119, line 218: I'm not sure this will work as-is.  Have you checked if the error message is translatable with a variable dropped in as the key? This may need to follow the [variable interpolation](https://www.i18next.com/translation-function/interpolation#working-with-data-models) pattern:

```suggestion
          {i18n.t('{{creatingClinicCustodialAccount.notification.message}}', { creatingClinicCustodialAccount })}
```

Otherwise, you could wrap the error message strings in `errorMessages.js`  in the `t()` functions like we do in `blip`.  If I'm mistaken, and this works as-is, feel free to disregard.

clintonium-119 (changes requested):
Looks good.  Just have a question/suggestion about the translation approach for the error message.

If you test it and it works, just re-request the review and I'll approve.

app/components/ClinicUserEdit.js
krystophv: Yeah, this is the commonly used pattern in Uploader. The value for the string to be translated _is_ the key that gets used for the localization file. It's probably not the best pattern to use, but I'm also not sure that revamping it is in the scope of this ticket. I'm sure it'll need to be revisited as we scope out moving Uploader into Web.

clintonium-119 (approved):
LGTM 💥

clintonium-119 (approved):
LGTM 🚢

clintonium-119 (dismissed):
Looks good 👍

gniezen (approved):
Let's ship it! :shipit:

PR #1553, opened 2023-04-19
[UPLOAD-1045] token refresh

krystophv:
to handle [UPLOAD-1045], preemptively refreshing the auth token 10 seconds prior to expiry as opposed to precisely at time of expiry

Uploader analog of https://github.com/tidepool-org/blip/pull/1220


clintonium-119 (approved):
LGTM :+1:

PR #1555, opened 2023-05-02
Add support for deep linking in Linux

gniezen:
This doesn't have any user-facing changes, so doesn't have a Jira ticket to link to. Still, it needs to be reviewed before being merged.


gniezen:
@krystophv Do you think we can get this merged soon, as it does not change any production code, and would make my life a lot easier :smile:

krystophv (approved):
LGTM 👍🏼

PR #1572, opened 2023-08-08
[UPLOAD-1118] keycloak logging

krystophv:
For [UPLOAD-1118]

- updates to latest `keycloak.js`
- adds lots of defensive checks for validity of tokens coming in from keycloak and logs a number of error cases
- sends users to the logged out screen in unrecoverable error cases

[UPLOAD-1118]: https://tidepool.atlassian.net/browse/UPLOAD-1118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/keycloak.js
```diff
@@ -3,22 +3,25 @@ import React, { useState, useMemo, useEffect, useCallback } from 'react';
...
 let refreshTimeout = null;
 
-export const setTokenRefresh = (keycloak) => {
+export const setTokenRefresh = (tokenParsed) => {
   if (refreshTimeout) {
     clearTimeout(refreshTimeout);
     refreshTimeout = null;
   }
-  var expiresIn = (keycloak.tokenParsed['exp'] - new Date().getTime() / 1000 + keycloak.timeSkew) * 1000;
+  let timeskew = keycloak?.timeSkew ?? 0;
```
gniezen, line 23: Did we not have an issue with using the nullish coalescing operator (??) in Uploader with Webpack/Babel? Or am I thinking of another operator?

app/keycloak.js
```diff
@@ -90,13 +93,57 @@ const onKeycloakEvent = (store) => (event, error) => {
...
+      }
+
+      if (!store.getState().loggedInUser) {
+        store.dispatch(async.doLogin());
+      }
+      setTokenRefresh(tokenParsed);
+    } else {
+      // if we don't have a sub and exp, we can't save the session
+      if (rollbar) {
+        rollbar.error('keycloak token missing sub or exp', {
```
gniezen, line 132: You don't necessarily have to wrap this in `{ }`, right?

app/keycloak.js
```diff
@@ -90,13 +93,57 @@ const onKeycloakEvent = (store) => (event, error) => {
...
   if (tokens?.token) {
     store.dispatch(sync.keycloakTokensReceived(tokens));
-    api.user.saveSession(keycloak?.tokenParsed?.sub, keycloak?.token, {
-      noRefresh: true,
-    });
-    if (!store.getState().loggedInUser) {
-      store.dispatch(async.doLogin());
+    let tokenParsed;
+    try {
+      tokenParsed = jose.decodeJwt(tokens?.token);
```
gniezen, line 98: Can there ever be more than one token here?

gniezen (approved):
Some questions, but nothing preventing this from being approved. :+1:

app/keycloak.js
krystophv: There is a refresh token and ID token that are also on the `tokens` object, but we don't use them for explicitly for the session.

app/keycloak.js
krystophv: It's only an issue if the code that's using it doesn't get run through Babel. I think the issue we were having was that a dependency was using it and we generally exclude `/node_modules` from Babel loading.

gniezen: Ah yes, it was a dependency that was causing the issue. That makes sense.

PR #1588, opened 2023-09-28
[UPLOAD-1139] CircleCI resource class update

krystophv:
to address [UPLOAD-1139], we need to update the CircleCI runner class since the Intel based `medium` is being sunset at the end of Sept 2023

[UPLOAD-1139]: https://tidepool.atlassian.net/browse/UPLOAD-1139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gniezen (approved):
Great job on finding a workaround for this! :clap: LGTM!

We should keep an eye on electron-builder to see if they manage to start using Apple's new certs and if that resolves the issue, in which case we can remove this again :crossed_fingers:

PR #1600, opened 2023-11-30
app wont build with currently mentioned v12.X

jh-bate:
set as same version used in automated builds

worlds smallest nitpic


gniezen (approved):
Yep, we really need to do a thorough review of the README as there are more outdated stuff in there. Maybe good to do as part of the move to Uploader-in-Web.

I'll merge this in as it doesn't affect anything user-facing so doesn't need to go through QA.

PR #1601, opened 2023-12-14
add path suffix for web uploader login redirect

krystophv:


app/components/Login.js
```diff
@@ -74,7 +74,7 @@ export const Login = () => {
       </button>
     );
   };
-  let redirectUri = win.location.origin;
+  let redirectUri = win.location.origin + env.electron ? '' : '/uploader/';
```
darinkrauss, line 77: Do we need the trailing slash?

darinkrauss (approved):
One question. Assuming we do need the trailing slash, then approved.

app/components/Login.js
krystophv: Turns out, no. I'll drop it.

PR #1611, opened 2024-03-14
[UPLOAD-1187][UPLOAD-1186][UPLOAD-1179] enforce patient limit

krystophv:
for [UPLOAD-1187], [UPLOAD-1186] - add UI that shows patient limit modal when attempting to add a patient that would put the clinic over their limit. This "borrows" a lot of the `blip` code for these things verbatim, which technically includes some dead code paths, but I think the benefit from keeping the relevant files in sync outweighs the drawback of having some unused code. (I'm not dead set on this and could probably be convinced otherwise if there's a good argument for it)

adds the two line fix for [UPLOAD-1179] as it came up as an issue for me when doing verification

[UPLOAD-1187]: https://tidepool.atlassian.net/browse/UPLOAD-1187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UPLOAD-1186]: https://tidepool.atlassian.net/browse/UPLOAD-1186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UPLOAD-1179]: https://tidepool.atlassian.net/browse/UPLOAD-1179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (dismissed):
LGTM :100: 
I agree that it makes sense to just drop the code straight in from blip where possible, even if some of it is unused.

Less effort now, and less effort down the road.  win-win.

app/actions/async.js
```diff
@@ -1025,6 +1029,12 @@ export function checkUploadTargetUserAndMaybeRedirect() {
 
 export function clickAddNewUser(){
   return (dispatch, getState) =>{
+    // check state for clinic patient limit exceeded and if so, display the limit modal
+    const { selectedClinicId, clinics } = getState();
+    const clinic = _.get(clinics, selectedClinicId);
+    if (clinic?.ui?.warnings.limitReached) {
```
vojtech-cerveny, line 1035: **thought (non-blocking):** It would be nice to fetch data from server (`/settings/patient_count`) after clicking on `Add new` button - during testing we found that if you reach limit, your next step is deleting (some unused) patient from the clinic. If you do that, state on server is changed (249/250 patients), but Uploader has same state - clinic is still on the limit. You need to reopen Uploader to get actual state. 

I don't think that it is huge problem (you need to restart) and it doesn't happen often. But if this would be easy to fix, it would be nice to reduce the number of restarts needed to get the application to the actual state.

krystophv: We try not to tie user interactions like an onClick navigate directly to network requests like this, as we'd need to wait for the return from the server before we knew if we should display the warning modal or send the user to the add new patient form. Without turning it into a UI-blocking request, we'd probably show the new patient form and then decide if we need to show the modal when the request returns, leading to a sort of layout flash/rug-pull for the UX (and then make another navigation when the modal is closed to get the user _back_ to the patient selection. A part of the complication comes from how we've constructed our network-based async Redux state, which I plan on revisiting during the new Uploader UI/UX refresh, so this would be a good note to add to the designs there so those can be built with this in mind.

vojtech-cerveny: Cool, I am ok with that, that it can be future enhancement :) Thanks for explaining!

PR #1648, opened 2024-08-05
Add Bluetooth support for ReliOn Platinum meter (UPLOAD-1121)

gniezen:
Addresses [UPLOAD-1121].

[UPLOAD-1121]: https://tidepool.atlassian.net/browse/UPLOAD-1121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
LGTM 🚢

yarn.lock
```diff
@@ -4521,7 +4521,12 @@ caniuse-api@^3.0.0:
     lodash.memoize "^4.1.2"
     lodash.uniq "^4.5.0"
 
-caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001565:
+caniuse-lite@^1.0.0:
```
vojtech-cerveny, line 4524: **question:** Do we want to update this package in this PR? 🤔 I would say, that this change would need better testing than just "new-device" tests. WDYT?

gniezen: This is not a user-facing change. It just adds an additional version of the package to resolve a warning in the build process. I don't think testing is required for this.

vojtech-cerveny (changes requested):
✅ LGTM

vojtech-cerveny (approved):

PR #1639, opened 2024-06-25
Only read new records for delta-capable Bluetooth meters (UPLOAD-219)

gniezen:
Addresses [UPLOAD-219]. Based against `in-browser-merged`, so that should be merged first.

[UPLOAD-219]: https://tidepool.atlassian.net/browse/UPLOAD-219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1650, opened 2024-08-06
Improve messaging when there are no (new) records (UPLOAD-1295)

gniezen:
Addresses [UPLOAD-1295].

### DIFF
--- a/lib/drivers/abbott/abbottFreeStyleLibre.js
+++ b/lib/drivers/abbott/abbottFreeStyleLibre.js
@@ -263,7 +263,9 @@ export default function (config) {
       debug('processData: num post records:', data.post_records.length);
 
       if (data.post_records.length === 0) {
-        return cb(new Error('Device has no records to upload'), null);
+        const err = new Error('No records to upload');
+        err.code = 'E_NO_RECORDS';
+        return cb(err, null);
       }
 
       progress(100);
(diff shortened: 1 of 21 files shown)


gniezen:
This work should not be merged before #1639, as it has been merged into this PR>

lib/drivers/bayer/bayerContourNext.js
```diff
@@ -491,8 +491,9 @@ module.exports = function (config) {
         dataToPost.push(smbg);
       }
     } else {
-      debug('Device has no records to upload');
```
krystophv, line 494: Most of the rest of these seem to keep the `debug()` call. Should we keep it here too?

lib/drivers/abbott/abbottFreeStyleLibre.js
```diff
@@ -263,7 +263,9 @@ export default function (config) {
       debug('processData: num post records:', data.post_records.length);
 
       if (data.post_records.length === 0) {
-        return cb(new Error('Device has no records to upload'), null);
+        const err = new Error('No records to upload');
```
krystophv, line 266: Should we add a `debug()` call above this line to be consistent with other devices?

krystophv (dismissed):
LGTM 🚀 

Couple small questions about consistency with the `debug()` calls.

lib/drivers/abbott/abbottFreeStyleLibre.js
gniezen: Done.

lib/drivers/bayer/bayerContourNext.js
gniezen: Done.

krystophv (dismissed):
LGTM 🚀

PR #1664, opened 2024-10-31
v2.60.0

gniezen:


gniezen:
@vojtech-cerveny Ready for a final smoke test before we release on Monday?

vojtech-cerveny:
Yeah, definitely :) Thanks for notice!

PR #1665, opened 2024-11-19
UPLOAD-1409 - Enhanced error reporting

vojtech-cerveny:
This pull request includes changes to enhance error reporting to Rollbar and update  
Enhancements to error reporting:

* [`app/actions/utils.js`](diffhunk://#diff-1964116493daea0e90a7e76fe37248f370f4a6fe866a003be6f65333f7fa0f9bL208-L225): Modified the `sendToRollbar` function to handle the case when `rollbar` is not defined, and improved the error reporting by merging additional properties into the reported error.

By merging `...props` to extra, application send also data included in `uploadErrProps` 

```js
      let uploadErrProps = {
        details: err.message,
        utc: getUtc(utc),
        name: err.name || 'Uncaught or API POST error',
        step: err.step || null,
        datasetId: err.datasetId || null,
        requestTrace: err.requestTrace || null,
        sessionTrace: err.sessionTrace || null,
        code: err.code || errCode,
        version: version,
        data: recs,
        loggedInUser: loggedInUser,
        userEmail: userEmail,
        userName: name,
        os: os,
        device: driverId,
      };
```

During testing it, I tried to upload data without connecting device and I got this [error on rollbar](https://app.rollbar.com/a/tidepool/fix/item/Uploader/7782). 

And as you can see, ticket has new params - custom params added by this PR. 
new params


app/actions/utils.js
```diff
@@ -205,24 +205,24 @@ export function mergeProfileUpdates(profile, updates){
...
 
-      rollbar.error(err, extra, (err, data) => {
-        if (err) {
-          console.log('Error while reporting error to Rollbar:', err);
+    const extra = {};
+    if (_.get(props, 'data.blobId', false)) {
+      _.assign(extra, { blobId: props.data.blobId });
+    }
+    
+    rollbar.error(err, {...extra, ... props}, (reportingErr, data) => {
```
gniezen, line 217: Is there a gap between `...` and `props`? We should probably remove it.

app/actions/utils.js
```diff
@@ -205,24 +205,24 @@ export function mergeProfileUpdates(profile, updates){
...
+      if (reportingErr) {
+        console.log('Error while reporting error to Rollbar:', reportingErr);
         } else {
```
gniezen, line 220: The code block formatting is a bit off here.

app/actions/utils.js
gniezen: Hmm, I see now we have `extra` and `props`. Why not combine them earlier, by doing a `const extra = { ...props }`, instead of having to use two spread operators here.

gniezen (changes requested):
Just a couple of small changes requested.

gniezen:
@vojtech-cerveny I think we may need to be a bit more specific with which `props` we add. If the upload fails while uploading to the backend, I think there's a chance that the props will include all the data being sent (under `props.data`).

vojtech-cerveny:
Well, I would say that I am OK with all data except `props.data`.

Because we send different objects into this function
For example:
```js
        let disconnectedErrProps = {
          utc: actionUtils.getUtc(utc),
          code: 'E_HID_CONNECTION',
          version: version,
          loggedInUser: loggedInUser,
          userEmail: userEmail,
          userName: name,
          os: os,
          device: driverId,
        };
```

and 

```js
      let uploadErrProps = {
        details: err.message,
        utc: getUtc(utc),
        name: err.name || 'Uncaught or API POST error',
        step: err.step || null,
        datasetId: err.datasetId || null,
        requestTrace: err.requestTrace || null,
        sessionTrace: err.sessionTrace || null,
        code: err.code || errCode,
        version: version,
        data: recs,
        loggedInUser: loggedInUser,
        userEmail: userEmail,
        userName: name,
        os: os,
        device: driverId,
      };

```

so my solution would be to just delete props.data, WDYT?

```js
    const extra = { ...props };
    delete extra.data;
```

gniezen:
Yeah, I think that should be fine. I just did a test with Tandem where I made it fail during uploading to the backend and `props.data` included the raw packets, compressed binary blob, parsed records and raw records, so definitely exclude that one :wink:

gniezen (approved):
LGTM! :ship:

gniezen (approved):

PR #1675, opened 2025-01-31
Allow VERSION_SHA to be specified directly (UPLOAD-1441)

darinkrauss:
This allows using the Dockerfile without requiring the `.git` directory (if, for example, the `.git` directory was specified in the `.dockerignore` file).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ WORKDIR /app
 RUN mkdir -p dist node_modules .yarn-cache && chown -R node:node .
 
 FROM base as build
+ARG VERSION_SHA
 ARG API_URL
 ARG UPLOAD_URL
 ARG DATA_URL
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tidepool-uploader",
   "productName": "tidepool-uploader",
-  "version": "2.61.0",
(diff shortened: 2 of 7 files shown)


gniezen:
@darinkrauss Is there a Jira ticket that we can link this PR to?

darinkrauss:
@gniezen There is now. 😄 https://tidepool.atlassian.net/browse/UPLOAD-1441

gniezen:
LGTM, but I'll let @krystophv approve the PR as I haven't been working with the Docker stuff lately. Do we need testing for this, or can we close out the ticket as "it will be implicitly be tested as part of the next release candidate" as there aren't any user-facing changes?

vojtech-cerveny:
Hey, it is not just Dockerfile, right? Can we create build for this to check if this works properly? I don’t expect any issues, but better safe than sorry.

vojtech-cerveny:
I checked code and I found only one usage of VERSION_SHA, and it is in [`rollbar.js`](https://github.com/tidepool-org/uploader/blob/4bf5164211c6d5e8abc514fe1c20ce6aa4a084c6/app/utils/rollbar.js#L16)

And rollbar shows correct commit - it redirects me to the latest commit in this branch. I checked the build on macos.

krystophv (approved):
LGTM 🚀

PR #1700, opened 2025-09-08
Update error message for Libre 2 attempt to highlight cloud-to-cloud workflow (UPLOAD-1497)

gniezen:
Addresses [UPLOAD-1497].


lib/core/api.js
```diff
@@ -18,19 +18,19 @@ import _ from 'lodash';
 import async from 'async';
 import { format } from 'util';
 import sundial from 'sundial';
-import { v4 as uuidv4 } from 'uuid';
+import { v4 as uuidv4 } from 'uuid'; // eslint-disable-line import/no-unresolved
```
krystophv, line 21: AFAICT Copilot is correct? I get no lint issues if I remove the disable.

app/utils/uploadDataPeriod.js
```diff
@@ -14,8 +14,9 @@
  * not, you can obtain one from Tidepool Project at tidepool.org.
  * == BSD2 LICENSE ==
  */
-import env from '../../app/utils/env';
-import { ipcRenderer, ipcMain } from '../../app/utils/ipc';
+import env from '../../app/utils/env.js';
+import pkg from '../../app/utils/ipc.cjs';
+const { ipcRenderer, ipcMain } = pkg;
```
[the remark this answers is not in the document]
krystophv: This looks correctly evaluated, but I don't really have an opinion on it one way or another

app/main.dev.js
```diff
@@ -3,17 +3,23 @@ import _ from 'lodash';
...
+import { setLanguage, i18n } from './utils/config.i18next.cjs';
 import path from 'path';
 import fs from 'fs';
 import child_process from 'child_process';
 
+const { autoUpdater } = pkg;
+const require = createRequire(import.meta.url);
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const { sync: syncActions } = require('./actions/index.js');
```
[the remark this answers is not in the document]
krystophv: Yeah, can't just make that suggestion edit. Might be worth leaving a comment about why, though.

krystophv (approved):
LGTM 🚀 

I left a handful of comments, mostly about what Copilot called out. Nothing that I think _needs_ addressing but you can take action on them if you feel like it.

app/utils/uploadDataPeriod.js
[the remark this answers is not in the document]
gniezen: Ah yes, changing the extension to `.cjs` made it unnecessary.  I'll change it back.

lib/core/api.js
gniezen: Ah, seems to be a eslint config issue. That was also the reason for splitting `import { ipcRenderer, ipcMain } from '../../app/utils/ipc.cjs';` into two lines.

gniezen: Seems to be an eslint config issue - I've found a fix.

krystophv (dismissed):
LGTM 🚄

app/main.dev.js
```diff
@@ -677,6 +764,50 @@ ipcMain.on('bluetooth-pairing-response', (event, response) => {
...
+        // Send error back to renderer
+        if (mainWindow && !mainWindow.isDestroyed()) {
+          const error = {
+            msgType: 'error',
+            details: 'Helper process is not available'
+          };
+          mainWindow.webContents.send('native-reply', error);
+        }
+      }
+    }, 100);
```
krystophv: I'm slightly concerned that 100ms might not be sufficient for thin-client class machines that often end up being used in clinical settings. Would some sort of polling with a TTL be better than a one-time hard coded wait?

krystophv (dismissed):
LGTM :shipit:

PR #1744, opened 2026-04-23
Fix: validate required environment variables in config.server.js

madelynporter23:
This change adds validation for environment variables in config.server.js.

- Added a helper function to check if required environment variables are defined
- Updated configuration to use validated environment variables
- Ensures the application throws an error when required variables are missing

This prevents runtime issues caused by missing configuration values.


gniezen:
Closing this PR as the code changes don't reflect the description. The submitted changes appear to only update the lockfile, with no implementation of the environment variable validation described.

If this was submitted in error, feel free to open a new PR with only the intended changes. Thanks!

PR #1747, opened 2026-05-12
roll back sinon-chai to a veresion compatible with our chai

gniezen:


gniezen:
Going to merge this one without review as it's not touching production code (just a test dependency) and I want to see how it affects dependabot

PR #1751, opened 2026-05-28
Fix linting errors from install_tools check

gniezen:
Moving the fixes made by @darinkrauss in TidepoolApi to this repo, which is the single source of truth used by Redocly.


gniezen (commented):
I copied these changes over from https://github.com/tidepool-org/TidepoolApi/pull/191, so even though I'm shown as the author here, I'm really the reviewer. 

Changes LGTM! 👍

PR #1763, opened 2026-07-23
Dependency updates for July 2026 (UPLOAD-1684)

gniezen:
As part of the dependency updates I have:

- removed the jest electron runner as it is not maintained anymore
- removed the rewire dependency as it is not maintained anymore
- switched from eslint to exlint
- changed oxlint warnings to errors and fixed them
- upgraded to babel 8


test/app/actions/async.test.js
```diff
@@ -58,9 +75,10 @@ jest.mock('@electron/remote', () => ({
 
 describe('Asynchronous Actions', () => {
   afterEach(() => {
-    // very important to do this in an afterEach than in each test when __Rewire__ is used
-    // if you try to reset within each test you'll make it impossible for tests to fail!
-    __ResetDependency__('services');
+    // restore in an afterEach rather than at the end of each test, so that a
+    // failing assertion can't skip the restore
+    appState.services = defaultServices;
+    appState.versionInfo = defaultVersionInfo;
```
krystophv: I think the `versionInfo` reassignment here is going to just assign an _already modified_ `defaultVersionInfo` since it's assigned by reference on L54 and `doAppInit` will mutate it when it runs. 

We probably need to create new objects at creation and use them in this afterEach:
```
// line 53-54
const defaultServices = { ...appState.services };
const defaultVersionInfo = { ...appState.versionInfo };

afterEach(() => {
  appState.services = { ...defaultServices };
  appState.versionInfo = { ...defaultVersionInfo };
});
```

krystophv (dismissed):
LGTM 🚀

PR #1767, opened 2026-08-10
Display one-button boluses; fix negative duration bug (UPLOAD-1474, UPLOAD-300)

gniezen:
Addresses [UPLOAD-1474] and [UPLOAD-300]..

[UPLOAD-1474]: https://tidepool.atlassian.net/browse/UPLOAD-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[UPLOAD-300]: https://tidepool.atlassian.net/browse/UPLOAD-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/utils/errors.js
+++ b/app/utils/errors.js
@@ -38,9 +38,15 @@ export function addInfoToError(err, props) {
     if (!_.isEmpty(v) && v !== err.message &&
         k !== 'utc' &&
         k !== 'code' &&
-        k !== 'version' &&
-        k !== 'data'
+        k !== 'version'
       ) {
+      if (k === 'data') {
+        // the device data is needed for the debug download links on failed
+        // uploads, but has to stay non-enumerable so it is never serialized
+        // into the metrics query string or error reports
+        Object.defineProperty(err, k, { value: v, writable: true, configurable: true });
+        return;
+      }
       err[k] = v;
       if (
         k !== 'uuid' &&
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -1845,6 +1845,10 @@ class NGPHistoryParser {
         const bolus = this.cfg.builder.makeNormalBolus()
           .with_normal(event.deliveredAmount);
 
+        if (event.bolusSource === NGPUtil.NGPConstants.BOLUS_SOURCE.EASY_BOLUS) {
+          bolus.with_deliveryContext('oneButton');
+        }
+
         if (event.programmedAmount !== event.deliveredAmount) {
           bolus.with_expectedNormal(event.programmedAmount);
         }
--- a/lib/drivers/medtronic600/NGPHistoryParser.js
+++ b/lib/drivers/medtronic600/NGPHistoryParser.js
@@ -1770,7 +1770,20 @@ class NGPHistoryParser {
 
   buildSuspendResumeRecords(events) {
     for (const event of this.eventsOfType(NGPHistoryEvent.EVENT_TYPE.INSULIN_DELIVERY_STOPPED)) {
-      const resumeEvent = this.findResumeForSuspend(event);
+      let resumeEvent = this.findResumeForSuspend(event);
+
+      let duration = resumeEvent == null ? 0 :
+        (resumeEvent.timestamp.rtc - event.timestamp.rtc) * 1000;
+      if (duration < 0) {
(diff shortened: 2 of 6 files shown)


lib/drivers/tandem
```diff
@@ -1 +1 @@
-Subproject commit 670e05c218027077e7fe047b36a23314a0c2af66
+Subproject commit 04a160513257b0f216d51914f4f01801d67252bf
```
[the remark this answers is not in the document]
krystophv: this commit does exist on that repository. that repo is private.

lib/drivers/medtronic600/NGPHistoryParser.js
```diff
@@ -1770,7 +1770,20 @@ class NGPHistoryParser {
 
   buildSuspendResumeRecords(events) {
     for (const event of this.eventsOfType(NGPHistoryEvent.EVENT_TYPE.INSULIN_DELIVERY_STOPPED)) {
-      const resumeEvent = this.findResumeForSuspend(event);
+      let resumeEvent = this.findResumeForSuspend(event);
+
+      let duration = resumeEvent == null ? 0 :
+        (resumeEvent.timestamp.rtc - event.timestamp.rtc) * 1000;
+      if (duration < 0) {
+        duration = resumeEvent.timestamp.toDate().valueOf() - event.timestamp.toDate().valueOf();
```
krystophv, line 1778: Could we annotate the duration when we fall back to wall-clock time? I ran a synthetic case with a negative RTC delta and a clock 4.5 hours ahead: a 30-minute suspend becomes 5 hours. The simulator then represents the extra 4.5 hours as suspended, retaining the scheduled rates as `suppressed`. An offset-only change doesn't reach this branch; the RTC must go backwards too. Because `toDate()` includes each event's stored offset, an incorrectly re-entered clock after a reset can reach this case, and `processPages` drops the corresponding time-change record.

app/utils/errors.js
```diff
@@ -38,9 +38,15 @@ export function addInfoToError(err, props) {
...
         k !== 'code' &&
-        k !== 'version' &&
-        k !== 'data'
+        k !== 'version'
       ) {
+      if (k === 'data') {
+        // the device data is needed for the debug download links on failed
+        // uploads, but has to stay non-enumerable so it is never serialized
+        // into the metrics query string or error reports
+        Object.defineProperty(err, k, { value: v, writable: true, configurable: true });
```
krystophv: Could we set `enumerable: false` explicitly and cover an error that already has `data`? I ran that case: if `err.data` was assigned before this call, redefining it without `enumerable` preserves `true`, and `JSON.stringify(err)` still includes the device data. The current test covers only a new property.

test/lib/medtronic600/testNGPHistoryParser.js
```diff
@@ -133,6 +169,74 @@ describe('NGPHistoryParser.js', () => {
       historyParser.buildSuspendResumeRecords(events);
       expect(events[0]).to.deep.equal(expected);
     });
+
+    test('should fall back to clock-corrected timestamps when the RTC resets during a suspend', () => {
+      // suspend at 2019-03-01T12:00:00, then the pump loses its clock
+      // (TIME_RESET, so the RTC counter restarts lower), then resume at
```
krystophv, line 175: Could we add a fixture containing `TIME_RESET` and the following `USER_TIME_DATE_CHANGE`? This fixture contains only stop and restart records, so it checks the negative-RTC arithmetic but doesn't exercise how `processPages` handles the reset and drops the following time change.

krystophv (commented):
A couple of questions came up during review.

lib/drivers/medtronic600/NGPHistoryParser.js
gniezen: Yes, great idea! The only duration-related annotation we currently have is https://github.com/tidepool-org/tideline/blob/develop/js/plot/util/annotations/annotationdefinitions.js#L129-L131

```
'basal/unknown-duration': function(source, defs) {
      return t('Because of how {{source}} reports the data, we could not determine the duration of this basal rate segment.', {source: source});
    },
```

Not sure if we want to create a new one, and if so, how should we phrase it?

PR #1770, opened 2026-08-18
Release 2.68.0 to main branch

gniezen:
v2.68.0 has been released to production, so we can merge this to main


gniezen:
I'll try to address the Coderabbit suggestions in the next dependency update PR.

.config.js
```diff
@@ -118,21 +118,14 @@ function stringToBoolean(str, defaultValue) {
...
-
 const selectedServerEnv = env.browser ? serverEnvFromLocation() : 'prd';
 
 module.exports = {
   // this is to always have the Bows logger turned on!
   // NB: it is distinct from our own "debug mode"
   DEBUG: stringToBoolean(process.env.DEBUG, true),
   // the defaults for these need to be pointing to prod
-  API_URL: process.env.API_URL || serverEnvironments[selectedServerEnv].API_URL,
+  API_HOST: process.env.API_HOST || serverEnvironments[selectedServerEnv].API_HOST,
```
krystophv: With `API_HOST` unset, I sourced `config/qa2.sh` and evaluated the config: it selects the production API while the other endpoints select QA2. Seven tracked config scripts still export `API_URL`. Could we rename those exports to `API_HOST` in a follow-up?

lib/drivers/abbott/freeStyleLibreData.js
```diff
@@ -387,7 +387,7 @@ export default class FreeStyleLibreData {
         );
         compressedOffset += blockLength;
       } else if (blockType === COMPRESSION_TYPE.ZERO_COMPRESSED) {
-        decompressedBuffer = decompressedBuffer.concat(new Array(blockLength));
+        decompressedBuffer = decompressedBuffer.concat(Array.from({length: blockLength}, () => 0));
```
krystophv: Could a follow-up test exercise a zero-compressed packet? This also fixes zero-block decompression. I ran the CRC path with sparse holes and explicit zeros, and the results differ because sparse arrays skip bytes during CRC16 calculation.

krystophv (approved):
LGTM 👍🏼

